PR #8, opened 2023-09-11
m_Do_main OK

TakaRikka:


src/m_Do/m_Do_main.cpp
```diff
@@ -1,72 +1,490 @@
...
+
+    if (mMaxTotalUsedSize < totalUsedSize)
+        mMaxTotalUsedSize = totalUsedSize;
+
+    if (mMaxTotalFreeSize > freeSize)
+        mMaxTotalFreeSize = freeSize;
 }
 
+// unused data?
+static u32 lit_2100[] = {0x3F800000, 0x3F800000, 0x3F800000, 0x3F800000, 0x3F800000, 0x3F800000};
```
magcius, line 54: mind making these floats?

configure.py
```diff
@@ -313,6 +313,7 @@ def ActorRel(rel_name, status):
             Matching   ("f_pc/f_pc_stdcreate_req.cpp"),
 
             # machine
+            NonMatching("m_Do/m_Do_main.cpp"),
```
magcius, line 316: still nonmatching?

magcius (approved):

src/m_Do/m_Do_main.cpp
TakaRikka: didn't seem to work correctly when using floats

configure.py
TakaRikka: yeah there's some weird linking stuff going on I'm unsure about. everything looks matching as far as I can tell though

PR #12, opened 2023-09-11
d_a_andsw2 OK

LagoLunatic:
image

The extra symbols don't seem to break the match since I didn't use sqrt.


encounter:
Could you match the visibility info? On the left, you'll see that the bottom half of functions are `[l]`, meaning local. Marking those functions as `static` should fix it.

LagoLunatic:
> Could you match the visibility info? On the left, you'll see that the bottom half of functions are `[l]`, meaning local. Marking those functions as `static` should fix it.

Done.

PR #14, opened 2023-09-12
Match d_a_boss_item.cpp

Sage-of-Mirrors:
Had to make some changes to function headers in d_com_inf_game and f_op_actor_mgr.

### DIFF
--- a/include/d/d_com_inf_game.h
+++ b/include/d/d_com_inf_game.h
@@ -378,7 +378,7 @@ inline void dComIfGs_onSaveTbox(int i_stageNo, int i_no) {
     g_dComIfG_gameInfo.save.getSavedata().getSave(i_stageNo).getBit().onTbox(i_no);
 }
 
-inline BOOL dComIfGs_isStageBossEnemy() {
+BOOL dComIfGs_isStageBossEnemy(int i_stageNo) {
     return g_dComIfG_gameInfo.save.getMemory().getBit().isStageBossEnemy();
 }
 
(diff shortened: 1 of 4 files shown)


include/d/d_com_inf_game.h
```diff
@@ -378,15 +378,15 @@ inline void dComIfGs_onSaveTbox(int i_stageNo, int i_no) {
     g_dComIfG_gameInfo.save.getSavedata().getSave(i_stageNo).getBit().onTbox(i_no);
 }
 
-inline BOOL dComIfGs_isStageBossEnemy() {
+BOOL dComIfGs_isStageBossEnemy(int i_stageNo) {
```
TakaRikka, line 381: are you sure these are right? the parameter isnt getting used at all

Sage-of-Mirrors: Ghidra's interpretation of this function is much more complex than the one that is in the repo. I assume that they simplified the function for TP?

Sage-of-Mirrors: I didn't update the function's C++ because I was focusing on d_a_bossitem.

TakaRikka: there's two versions of the function according to tww debug map, one with an int param and one with void. maybe it wants the int one from the cpp file instead of this?

Sage-of-Mirrors: Ahhh, I see! That's why it has the `extern`s. One second.

PR #20, opened 2023-09-12
More d_a_item funcs

LagoLunatic:


src/d/actor/d_a_item.cpp
```diff
@@ -322,7 +517,7 @@ s32 daItem_Delete(daItem_c* i_this) {
 
 /* 800F89D0-800F89F0       .text daItem_Create__FP10fopAc_ac_c */
 s32 daItem_Create(fopAc_ac_c* i_this) {
-    /* Nonmatching */
+    ((daItem_c*)i_this)->_daItem_create();
```
TakaRikka, line 520: this should return the create ret value

src/d/actor/d_a_item.cpp
```diff
@@ -28,6 +58,24 @@ float daItem_c::getYOffset() {
...
+    case RED_JELLY:
+    case GREEN_JELLY:
+    case BLUE_JELLY:
+        return 0.0;
+    case SWORD:
+        return 20.0;
+    case SHIELD:
+        return 23.0;
+    case DROPPED_SWORD:
+        return 10.0;
```
TakaRikka, line 78: are these actually double values or are they missing a `f`?

PR #21, opened 2023-09-13
d_a_swc00

SuperDude88:
Mostly matching, some regalloc issues in execute (possibly an actor parameter inline)

### DIFF
--- a/src/d/actor/d_a_swc00.cpp
+++ b/src/d/actor/d_a_swc00.cpp
@@ -3,26 +3,90 @@
 // Translation Unit: d_a_swc00.cpp
 //
 
-#include "d_a_swc00.h"
-#include "dolphin/types.h"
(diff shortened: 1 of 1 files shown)


src/d/actor/d_a_swc00.cpp
```diff
@@ -3,26 +3,90 @@
...
+struct swc00_class : public fopAc_ac_c {};
+
+static inline s32 getSwBit(u32 param) {
+    return param & 0xFF;
 }
 
-/* 00000180-00000188       .text daSwc00_IsDelete__FP11swc00_class */
-void daSwc00_IsDelete(swc00_class*) {
-    /* Nonmatching */
+static s32 daSwc00_Execute(swc00_class* i_this) {
```
TakaRikka, line 17: if this is still nonmatching, i think leaving the nonmatching comment would be helpful

PR #25, opened 2023-09-13
d_a_tag_kb_item OK

Trueffeloot:


Trueffeloot:
Doesn't match for jpn tww? Do i have to fix something here?

src/d/actor/d_a_tag_kb_item.cpp
```diff
@@ -1,58 +1,131 @@
...
-void daTagKbItem_Draw(void*) {
-    /* Nonmatching */
+static int daTagKbItem_Draw(void* i_this) {
+    return static_cast<daTagKbItem_c*>(i_this)->_draw();
 }
 
 /* 00000234-00000258       .text daTagKbItem_Execute__FPv */
-void daTagKbItem_Execute(void*) {
-    /* Nonmatching */
+static int daTagKbItem_Execute(void* i_this) {
```
TakaRikka, line 101: ideally these return types should match the return type of the member function

Trueffeloot: not in this case though. Objdiff shows clrlwi's that need to be generated in the static execute and static draw function. It doesn't match else

magcius: hm, ill approve for now but i want to take a closer look at that later

PR #28, opened 2023-09-14
daShopItem_c Almost OK

SuperDude88:


include/d/d_cloth_packet.h
```diff
@@ -0,0 +1,59 @@
...
+    f32 mDrag;
+    f32 mFlyFlex;
+    f32 mHoistFlex;
+    s16 mWave;
+    s16 mWaveSpeed;
+    s16 field_0xF0;
+    s16 field_0xF2;
+    s16 mRipple;
+    s16 mRotateY;
+    u8 mCurArr;
```
TakaRikka, line 38: would be great if offset comments could be put on the left of members

PR #32, opened 2023-09-14
Started work on pushable boxes

LagoLunatic:


include/f_op/f_op_actor_mng.h
```diff
@@ -337,16 +337,16 @@ inline void dComIfGs_offSwitch(int i_no, int i_roomNo);
 inline BOOL dComIfGs_isSwitch(s32 i_no, s32 i_roomNo);
 inline void dComIfGs_offActor(int i_no, int i_roomNo);
 
-inline void i_fopAcM_onSwitch(const fopAc_ac_c* pActor, int sw) {
-    return dComIfGs_onSwitch(sw, fopAcM_GetHomeRoomNo(pActor));
+inline void fopAcM_onSwitch(const fopAc_ac_c* pActor, int sw) {
+    return dComIfGs_onSwitch(sw, pActor->orig.roomNo);
```
TakaRikka, line 341: are these changes needed to match? tp debug at least confirms `fopAcM_GetHomeRoomNo` was used

src/d/actor/d_a_obj_movebox.cpp
```diff
@@ -3,671 +3,874 @@
...
+        inline s32 prm_get_itemSave() const { return daObj::PrmAbstract(this, PRM_ITEMSAVE_W, PRM_ITEMSAVE_S); }
+        inline s32 prm_get_buoy() const { return daObj::PrmAbstract(this, PRM_BUOY_W, PRM_BUOY_S); }
+        inline s32 prm_get_dmy() const; // Unused?
+        
+        inline BOOL is_switch1() const { return fopAcM_isSwitch(this, prm_get_swSave1()); };
+        inline BOOL is_switch2() const { return fopAcM_isSwitch(this, prmZ_get_swSave2()); };
+        inline void on_switch1() const { fopAcM_onSwitch(this, prm_get_swSave1()); };
+        inline void on_switch2() const { fopAcM_onSwitch(this, prmZ_get_swSave2()); };
+        inline void off_switch1() const { fopAcM_offSwitch(this, prm_get_swSave1()); };
+        inline void off_switch2() const { fopAcM_offSwitch(this, prmZ_get_swSave2()); };
```
TakaRikka, line 135: inlines that are defined in the class don't need the inline keyword

include/f_op/f_op_actor_mng.h
LagoLunatic: It's needed for the regalloc to match, not only in obj_movebox but also in item and andsw0.

Although `fopAcM_GetHomeRoomNo` does still appear in the debug symbol maps for obj_movebox and andsw0, which I guess indicates it is used? Maybe there's a better fix.

TakaRikka: maybe just add a comment about it for now so it can be revisited if the issue comes back the other way around where it only matches with the inline

PR #35, opened 2023-09-15
d_a_kaji OK

LagoLunatic:


src/d/actor/d_a_kaji.cpp
```diff
@@ -3,41 +3,171 @@
...
+}
+
+bool daKaji_c::_draw() {
+    if (!l_p_ship->unk2CC) {
+        return 1;
+    }
+    g_env_light.settingTevStruct(TEV_TYPE_ACTOR, &current.pos, &mTevStr);
+    g_env_light.setLightTevColorType(mpMorf->getModel(), &mTevStr);
+    mpMorf->update();
+    return 1;
```
encounter, line 122: Let's use true/false for bool

src/d/actor/d_a_kaji.cpp
```diff
@@ -3,41 +3,171 @@
...
-    /* Nonmatching */
+static s32 daKajiDraw(void* i_this) {
+    return ((daKaji_c*)i_this)->_draw();
 }
 
 /* 00000568-00000570       .text daKajiIsDelete__FPv */
-void daKajiIsDelete(void*) {
-    /* Nonmatching */
+static s32 daKajiIsDelete(void* actor) {
+    return 1;
```
encounter, line 147: It's likely these all of these s32 functions return BOOL + TRUE/FALSE

encounter (approved):

PR #38, opened 2023-09-15
Match & link d_a_alldie; more d_com_inf_game

encounter:


src/d/d_com_inf_game.cpp
```diff
@@ -183,11 +183,11 @@ int dComIfG_play_c::getLayerNo(int i_roomNo) {
     int layer = dComIfGp_getStartStageLayer();
 
     if (layer < 0) {
-        int hour = dKy_getdaytime_hour();
+        layer = dKy_getdaytime_hour();
         if (dKy_checkEventNightStop()) {
             layer = 1;
         } else {
-            layer = (hour >= 6 && hour < 18) == 0;
+            layer = (layer >= 6 && layer < 18) ? 0 : 1;
```
TakaRikka, line 190: the `(layer >= 6 && layer < 18)` doesn't really make sense in what it's actually checking, since it's checking for day / night time, but if it improves matching then i guess it's fine for now

TakaRikka (approved):

src/d/d_com_inf_game.cpp
fig02: at this point would prob just name it `temp` or something generic since it has two meanings. this is an unfortunate somewhat common occurrence in the 64 zeldas at least lol

fig02: maybe not cause the whole function would be much harder to read. since its all about the layer number. but yea that sucks. maybe a comment on the temp reuse, idk

encounter: I'm not entirely sure the temp reuse is real, but it gets the function closer. It's almost like the entire contents of this if statement were inlined, but debug map doesn't show anything, nor does an inline of that size actually inline. It's a tricky one.

PR #41, opened 2023-09-16
d_seafightgame, m_Do_machine_exception,  m_Do_reset, m_Do_DVDError OK

TakaRikka:


src/m_Do/m_Do_Reset.cpp
```diff
@@ -1,32 +1,118 @@
...
+        do {
+        } while (true);
+    }
+
+    JUTXfb::getManager()->clearIndex();
+    mDoDvdErr_ThdCleanup();
+    JUTGamePad::clearForReset();
+    mDoAud_zelAudio_c::offInitFlag();
+
+#if VERSION != VERSION_JP
```
encounter, line 47: I wonder if we should start doing `VERSION > VERSION_JP` to make it clear things were added in future versions. Or `VERSION >= VERSION_USA` might be most clear?

encounter (approved):

src/m_Do/m_Do_Reset.cpp
TakaRikka: the 2nd one might be more clear, but yeah it should probably be discussed more in the discord along with how best to handle these kinds of version differences, since ifdef hell could get bad

PR #53, opened 2023-09-18
m_Do_printf

Jcw87:


include/m_Do/m_Do_printf.h
```diff
@@ -0,0 +1,22 @@
...
+    void OSReportDisable();
+    void OSReportEnable();
+    void OSReportForceEnableOn();
+    void OSReportForceEnableOff();
+    void OSVReport(const char* format, va_list list);
+    void OSReport(const char* fmt, ...);
+    void OSReport_FatalError(const char* fmt, ...);
+    void OSReport_Error(const char* fmt, ...);
+    void OSReport_Warning(const char* fmt, ...);
+    void OSPanic(const char* file, s32 line, const char* fmt, ...);
```
TakaRikka, line 17: shouldn't these be in OS.h? i know they're declared or weak or something so they get emitted in m_Do_printf, but feels like the prototypes should still be in the dolphin related headers

Jcw87: I checked the [Metroid Prime decomp](https://github.com/PrimeDecomp/prime/blob/304c6cfca9f011ba9677f12d115c28cdb8e73af8/src/Dolphin/os/OSError.c#L14-L40), and you can find `OSReport`, `OSVReport`, and `OSPanic` in `OSError.o` The implementation of `OSVReport` only calls `vprintf`, and doesn't use `__OSReport_disable` or related variables.

These symbols being weak wouldn't automatically place them in `m_Do_printf.o`, they would need to have overrides defined there. This tells me that wind waker and twilight princess override `OSVReport` and add these extra variables and related functions. So, it makes sense for `OSReport`, `OSVReport`, and `OSPanic` to be in the SDK headers, but not so much for the others. I included duplicates of these in `m_Do_printf.h` for the sake of completeness.

PR #58, opened 2023-09-18
[ACTOR] d_a_rectangle OK

MrPolymorph:
Possibly the easiest actor I've worked on.

image


LagoLunatic:
Ideally you should add the profile and method array, it won't match without those

MrPolymorph:
@LagoLunatic should all be sorted now

PR #69, opened 2023-09-21
Add .inc file association to recommended settings

LagoLunatic:


src/d/actor/d_a_player_HIO.inc
```diff
@@ -6,8 +6,8 @@
  * This file is not a standalone translation unit and is instead directly 
  * included into d_a_player_main.cpp.
  * 
- * The original name of this file is known because of an assert contained within
- * a function from this file.
+ * The original name of this file is not known, but a best guess was taken 
+ * based on the original names of the classes it contains.
  */
```
fig02, line 11: nice catch.
iirc taka said this one is known to exist from TP? we can make a version of this comment that says its not known from TWW but we assume it likely is the same from twilight princess. @TakaRikka

LagoLunatic: fwiw I don't see this string in the files for the release version of US TP, haven't checked other versions though.

PR #71, opened 2023-09-22
d_item_data & d_a_itembase OK

LagoLunatic:


src/d/d_item_data.cpp
```diff
@@ -3,26 +3,10152 @@
...
+    "tingle_figure.bti",
+    "bottle_09.bti",
+    "cmap_hint2.bti",
+    "delivery_19.bti",
+};
+
+/* 803842B0-803866B0       .data item_resource__10dItem_data */
+dItem_data_item_resource dItem_data::item_resource[0x100] = {
+    {
+        // 0x00
```
TakaRikka, line 254: iirc each entry in here is related to the corresponding item in the item table, so it might be nice to comment which the item name here, but its not a super important thing

TakaRikka (approved):

LagoLunatic:
itembase also matches now.

shop_item appears to be done but doesn't match because of the same issue as tag_ret (the compiler puts weak functions in the wrong order in RELs).

PR #73, opened 2023-09-23
Link J3DSys

encounter:


src/JSystem/J3DGraphBase/J3DSys.cpp
```diff
@@ -123,7 +123,7 @@ void J3DFifoLoadNrmMtxImm3x3(Mtx33 mtx, unsigned long idx) {
     GFX_FIFO(u32) = mtx_u32[8];
 }
 
-u8 NullTexData[16] = {
+u8 NullTexData[16] __attribute__((aligned(32))) = {
```
encounter, line 126: Not sure what type this is supposed to be, but the map & symbols.txt have it listed as `align:32`, which is odd. Required to get the section alignment correct.

src/m_Do/m_Do_main.cpp
```diff
@@ -51,7 +51,8 @@ void HeapCheck::CheckHeap1() {
 }
 
 // unused data?
-static u32 lit_2100[] = {0x3F800000, 0x3F800000, 0x3F800000, 0x3F800000, 0x3F800000, 0x3F800000};
+static u32 lit_2100[] = {0x3F800000, 0x3F800000, 0x3F800000};
+static u32 lit_2080[] = {0x3F800000, 0x3F800000, 0x3F800000};
```
encounter, line 55: Unrelated change, this allows `ninja diff` to work again. Before, it was complaining that the symbol sizes were different here and bailing.

PR #499, opened 2023-10-01
arrow, tag_ret, obj_Ygush00, kytag06, nh: all OK

LagoLunatic:
I switched arrow, tag_ret, and ygush to use the `-sym off` compiler flag. This might not be exactly what the original developers did but I think it's worth doing temporarily for the sake of making them match until someone figures out what MWCC is doing with implicit destructors.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -429,7 +429,7 @@ config.libs = [
             Object(NonMatching, "d/actor/d_a_ib.cpp"),
             Object(NonMatching, "d/actor/d_a_item.cpp"),
             Object(Matching,    "d/actor/d_a_itembase.cpp"),
-            Object(NonMatching, "d/actor/d_a_nh.cpp"),
+            Object(Matching,    "d/actor/d_a_nh.cpp"),
             Object(NonMatching, "d/actor/d_a_npc_fa1.cpp"),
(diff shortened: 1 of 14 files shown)


configure.py
```diff
@@ -414,7 +414,7 @@ def DolphinLib(lib_name, objects):
             Object(NonMatching, "d/d_snap.cpp"),
             Object(Matching,    "d/d_point_wind.cpp"),
             Object(NonMatching, "d/actor/d_a_agb.cpp"),
-            Object(NonMatching, "d/actor/d_a_arrow.cpp"),
+            Object(Matching,    "d/actor/d_a_arrow.cpp", cflags=[*cflags_framework, "-sym off"]),
```
magcius, line 417: This should just be `-sym off`, right? cflags_rel already includes cflags_framework

LagoLunatic: arrow isn't it a rel, it's an actor in main.dol.
If I do `cflags=["-sym off"]` it won't compile at all, it seems to replace all the flags with just "-sym off", instead of adding them.

magcius: ah. we can fix that I suppose, but for now

Issue #422, opened 2023-09-28
d_a_obj_mknjd


magcius:
MknjD

Sage-of-Mirrors:
In progress at https://github.com/Sage-of-Mirrors/tww/blob/obj_mknjd/src/d/actor/d_a_obj_mknjd.cpp

Sage-of-Mirrors:
Major progress merged with PR #504. As of this PR:
* `daObjMknjD::Act_c::CreateHeap()` is at 95% with regswaps; all other functions match 100%.
* Structure of the TU needs to be adjusted to get the TU to match 100%.

PR #504, opened 2023-10-04
Majority of daObjMknjD Done

Sage-of-Mirrors:
image
All functions except `daObjMknjD::Act_c::CreateHeap()` are matched; `CreateHeap()` has some regswaps. Unit also requires structural changes to match the full REL 100%.

Advances #422.


include/JSystem/J3DGraphAnimator/J3DModel.h
```diff
@@ -128,16 +129,16 @@ class J3DModel {
     Mtx33** getBumpMtxPtrPtr(u32 idx) const { return mpBumpMtxArr[1][idx]; }
     Mtx33* getBumpMtxPtr(u32 idx) const { return mpBumpMtxArr[1][idx][mCurrentViewNo]; }
     void setBaseScale(const Vec& scale) { mBaseScale = scale; }
-    void setUserArea(u32 area) { mUserArea = area; }
-    u32 getUserArea() const { return mUserArea; }
+    void setUserArea(void* area) { mUserArea = area; }
+    void* getUserArea() const { return mUserArea; }
```
TakaRikka, line 133: even though void* makes more sense, the map shows u32 was originally used, so it's probably best to keep it as close as possible

include/f_op/f_op_actor.h
```diff
@@ -135,20 +135,20 @@ class dEvt_info_c {
...
-    void suspendProc(void* actor) {
-        if (field_0x10 != NULL) {
-            field_0x14(actor);
-        }
-    }
+    //void suspendProc(void* actor) {
+    //    if (field_0x10 != NULL) {
+    //        field_0x14(actor);
+    //    }
+    //}
```
TakaRikka, line 142: if this stuff is gone its fine to remove it

src/d/actor/d_a_obj_mknjd.cpp
```diff
@@ -3,216 +3,1144 @@
...
+        enum Prm_e {
+            PRM_SWITCH_W = 0x08,
+            PRM_SWITCH_S = 0x00,
+
+            PRM_TYPE_W = 0x01,
+            PRM_TYPE_S = 0x10,
+        };
+
+        inline int prm_get_swSave() { return daObj::PrmAbstract<Prm_e>(this, PRM_SWITCH_W, PRM_SWITCH_S); }
+        inline u8 prm_get_Type() { return daObj::PrmAbstract<Prm_e>(this, PRM_TYPE_W, PRM_TYPE_S); }
```
TakaRikka, line 166: inline keyword is unnecessary for functions defined inside the class

src/d/actor/d_a_obj_mknjd.cpp
```diff
@@ -3,216 +3,1144 @@
...
 }
 
 /* 00000D28-00000E04       .text setGoal__Q210daObjMknjD5Act_cFi */
-void daObjMknjD::Act_c::setGoal(int) {
+void daObjMknjD::Act_c::setGoal(int i_staffIdx) {
     /* Nonmatching */
+    cXyz pos = *dComIfGp_evmng_getMyXyzP(i_staffIdx, "Posion");
+
+    mDoMtx_stack_c::transS(getPosition());
+    mDoMtx_YrotM(mDoMtx_stack_c::get(), current.angle.GetY());
```
TakaRikka, line 463: this can just be `current.angle.y`, the getters didnt exist in original and should prob be removed

src/d/actor/d_a_obj_mknjd.cpp
```diff
@@ -3,216 +3,1144 @@
...
+            u32 prio = fpcLf_GetPriority(judgeResult);
+            fopDwTg_ToDrawQ(&judgeResult->mDwTg, prio);
+        }
+        else {
+            fopDwTg_DrawQTo(&judgeResult->mDwTg);
+        }
+    }
+
+    u16 makarProc = PROC_NPC_CB1;
+    judgeResult = static_cast<fopAc_ac_c*>(fopAcIt_Judge(fpcSch_JudgeForPName, &makarProc));
```
TakaRikka, line 676: im pretty sure there's an inline like `fopAcM_SearchByID` or something like that that makes these kind of calls look a little prettier

TakaRikka: looks like its `fopAc_ac_c* fopAcM_SearchByName(s16 proc_id)`

include/d/actor/d_a_player.h
```diff
@@ -109,7 +112,7 @@ class daPy_py_c : public fopAc_ac_c {
     virtual void checkCutCharge() const;
     virtual void getBokoFlamePos(cXyz*);
     virtual void checkTactWait() const;
-    virtual void setTactZev(unsigned int, int, char*);
+    virtual void setTactZev(unsigned int, int, const char*);
```
LagoLunatic, line 115: The debug maps show this param isn't const, changing it to be const might break regalloc

include/d/d_particle.h
```diff
@@ -374,7 +374,7 @@ class dPa_control_c {
                               const cXyz* scale, u8 alpha, dPa_levelEcallBack* pCallBack,
                               s8 setupInfo, const GXColor* pPrmColor, const GXColor* pEnvColor,
                               const cXyz* pScale2D) {
-        return set(2, particleID, pos, angle, scale, alpha, pCallBack, setupInfo, pPrmColor, pEnvColor, pScale2D);
+        return set(1, particleID, pos, angle, scale, alpha, pCallBack, setupInfo, pPrmColor, pEnvColor, pScale2D);
```
LagoLunatic, line 377: Checking more carefully, I think yours (2) is correct and mine (1) is wrong for setToon. I think the actual names of the inlines d_a_arrow calls to get 1 are `dComIfGp_particle_setP1` -> `dPa_control_c::setNormalP1`

PR #510, opened 2023-10-08
player OK, started adding player flag enums and inlines, kytag05 OK

LagoLunatic:


include/d/d_com_inf_game.h
```diff
@@ -39,6 +39,141 @@ class camera_class;
 class J2DOrthoGraph;
 class dDemo_actor_c;
 
+enum ALWAYS_RES_FILE_IDS {
```
magcius, line 42: one minor nit -- can you change this to FILE_INDEX? because "file ID" means something different (getObjectResID)

LagoLunatic: Always.arc has the flag I call "Sync File IDs and Indexes" set, meaning its file IDs and indexes are always identical. When the flag is not set, they can be different.

I added a GCFT feature to export enums like these, and if that flag is set it only exports the file IDs enum for the RARC, while if it's not set it will export the file IDs and indexes as two separate enums.

Maybe the enum's name should reflect when the flag is set? Something like "ALWAYS_RES_FILE_IDS_AND_INDEXES"? Since the enum can be used with either getObjectIDRes or getObjectRes.

magcius: that's fine then, though maybe there should be a comment there.

PR #519, opened 2023-10-12
Some Runtime.PPCEABI.H files

Yotona:


Prakxo:
The files have no new line at the end, which causes issues

Yotona:
Should be resolved. Pretty much just copy-paste from Pikmin, which has pretty bizarre formatting, surely these files will need some tending to if tww adopts a style standard.

PR #527, opened 2023-10-16
J3DVisibility OK

MrPolymorph:


src/d/actor/d_a_am.cpp
```diff
@@ -1212,7 +1212,7 @@ static BOOL useHeapInit(fopAc_ac_c* i_actor) {
             /* field_0x0   */ 2,
             /* mJointIndex */ 0x05, // hitomi (pupil) joint
             /* field_0x4   */ 5.0f,
-            /* field_0x8   */ (cXyz*)&cyl2_eye_offset,
+            /* mbActive   */ (cXyz*)&cyl2_eye_offset,
```
LagoLunatic, line 1215: Looks like a find and replace error. __jnt_hit_data_c is not J3DVisibilityManager

LagoLunatic (changes requested):

src/d/actor/d_a_am.cpp
MrPolymorph: Balls, sorry. I used CLions refactor function.

MrPolymorph: Fixed

LagoLunatic (approved):

PR #528, opened 2023-10-19
[Actor] d_a_obj_pbka OK

MrPolymorph:
100% Match


src/d/actor/d_a_obj_pbka.cpp
```diff
@@ -2,52 +2,133 @@
...
+    }
+    return true;
 }
 
-/* 0000015C-000001CC       .text CreateInit__11daObjPbka_cFv */
 void daObjPbka_c::CreateInit() {
-    /* Nonmatching */
+    mCullMtx =  mpModel->getBaseTRMtx();
+    fopAcM_setCullSizeBox(this,-300.0,-300.0,-300.0,300.0,300.0,300.0);
+    mCullSizeFar = 0x3f800000;
```
LagoLunatic, line 55: mCullSizeFar is a float not an int, if ghidra isn't showing it properly you can use https://gregstoll.dyndns.org/~gregstoll/floattohex/ to convert manually and figure out what the actual value is in these cases

MrPolymorph: Nice one, thank you!

PR #533, opened 2023-10-28
d_a_Tbox: Match demoProc(), add modelInfo array

Sage-of-Mirrors:
Match `daTbox_c::demoProc()` and add defines + the modelInfo array.

### DIFF
--- a/src/d/actor/d_a_tbox.cpp
+++ b/src/d/actor/d_a_tbox.cpp
@@ -351,7 +351,6 @@ void daTbox_c::demoProcAppear() {
 
 /* 00001E4C-0000210C       .text demoProc__8daTbox_cFv */
 s32 daTbox_c::demoProc() {
-    /* Nonmatching */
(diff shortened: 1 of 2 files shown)


src/d/actor/d_a_tbox.cpp
```diff
@@ -314,8 +348,98 @@ void daTbox_c::demoProcAppear() {
...
+                mEfLight.mPower = 0.0f;
+                break;
+            case DEMO_PROC_APPEAR:
+                m03F0 |= 0x20;
+                m03EC = -130.0f;
+                
+                setDzb();
+
+                if (getFuncType() == FUNC_TYPE_TACT) {
+                    m03F0 &= -3;
```
LagoLunatic, line 379: if this is a bitfield it might be more readable as `m03F0 &= ~2;` because of two's complement

LagoLunatic (approved):

Issue #193, opened 2023-09-28
d_a_tbox


magcius:
takara, takara2, takara3, takara4, takara5, takara6, takara7, takara8, takaraK, takaraI, takaraM, tkrASw, tkrAGc, tkrAKd, tkrAIk, tkrBMs, tkrCTf, tkrAOc, tkrAOs

PR #536, opened 2023-10-30
d_a_tbox OK

Sage-of-Mirrors:


src/d/actor/d_a_tbox.cpp
```diff
@@ -693,53 +694,200 @@ s32 daTbox_c::CreateInit() {
...
+
+    if (mAppearEvtTimer == 0x04 && mSmokeCB.getEmitter() != NULL) {
+        mSmokeCB.end();
+    }
+
+    if (mAppearEvtTimer != 0x00) {
+        mAppearEvtTimer--;
+    }
+
+    if (mpAppearRegAnm->play() != FALSE) {
```
LagoLunatic, line 887: does `if (mpAppearRegAnm->play()) {` not match?

src/d/actor/d_a_tbox.cpp
```diff
@@ -943,6 +1249,94 @@ s32 daTbox_Draw(daTbox_c* i_tbox) {
...
+
+    if (checkOpen() == FALSE) {
+        dMap_drawPoint(5, current.pos.x, current.pos.y, current.pos.z, mRoomNo, -0x8000, openFlag, mGbaName, 0);
+    }
+
+    mTevStr.mRoomNo = mRoomNo;
+    g_env_light.settingTevStruct(TEV_TYPE_ACTOR, getPositionP(), &mTevStr);
+
+    if (getFuncType() == FUNC_TYPE_TACT) {
+        J3DModelData* platMdlData = mTactPlatformMdl->getModelData();
```
LagoLunatic, line 1273: might wanna rename to `mpTactPlatformMdl` to make it more clear it's a pointer

Sage-of-Mirrors:
`d_a_tbox` is now OK when it comes to data/rodata/text, but there is a weird bug with the relocation tables that I'm not sure how to fix.

Sage-of-Mirrors:
Relocation bug fixed - had some PTMFs in the wrong order.

Closes #193.

PR #539, opened 2023-11-13
Nearly 100% match d_a_swhit0

Sage-of-Mirrors:
d_a_swhit0 almost OK, just missing `daSwhit0_c::getTimer()` and `daSwhit0_Draw()`.


src/d/actor/d_a_swhit0.cpp
```diff
@@ -3,341 +3,491 @@
...
+#include "m_Do/m_Do_mtx.h"
+
+static dCcD_SrcSph l_sph_src = {
+    // dCcD_SrcGObjInf
+    {
+        /* Flags             */ 0,
+        /* SrcObjAt Type     */ 0,
+        /* SrcObjAt Atp      */ 0,
+        /* SrcObjAt SPrm     */ 0,
+        /* SrcObjTg Type     */ 0xFF1DFEFF,
```
LagoLunatic, line 21: I would write this with enums like `~(AT_TYPE_WATER | AT_TYPE...)` so it's more readable

Sage-of-Mirrors: Done!

PR #541, opened 2023-11-24
Automatically generate headers for all actors

LagoLunatic:
* Created headers for everything in d/actor
* Update header include paths in the .cpp files to have the d/actor prefix
* Created a basic skeleton for each actor's class with the appropriate name (did not create any other classes/structs for each actor besides the main one that inherits from fopAc_ac_c)
* Added inline member functions extracted from the debug maps to the skeleton class
* Added member function declarations extracted from the .cpp files to the skeleton class
* Added member variables extracted from the Ghidra repo to the skeleton class (only applies to a few actors that have been documented in Ghidra but not the decomp, for the rest I put a placeholder "Place member variables here" comment)
* Tried to automatically add some common includes needed for the members in the header to compile (e.g. c_phase, m_Do_ext) (did not add includes for the .cpp file)

With these changes the simplest actors that haven't been started yet now compile out of the box (e.g. d_a_wbird), but most still don't, usually because their .cpp files have duplicate weak functions redefined in them.

Anything about the headers that could be improved?


magcius:
It would be nice if the _create/_draw/_execute/_delete functions could have the right types by default, but this is a fantastic start!

PR #544, opened 2023-11-29
Match d_a_lamp

jmerdich:
Matched everything in lamp_class. Some member names might be off, but I gave my best guess.

### DIFF
--- a/src/d/actor/d_a_lamp.cpp
+++ b/src/d/actor/d_a_lamp.cpp
@@ -41,8 +41,8 @@ static dCcD_SrcSph sph_src = {
 static u8 padding[76];
 
 // Need to break these out to get the rodata ordered right
-static const float partHeightOffset = 20.0;
-static const float partMaxFlickerPerTick = 0.02;
+static const float partHeightOffset = 20.0f;
+static const float partMaxFlickerPerTick = 0.02f;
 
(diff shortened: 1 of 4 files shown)


src/d/actor/d_a_lamp.cpp
```diff
@@ -4,39 +4,230 @@
...
+    {
+        /* Center */ 0.0f, 0.0f, 0.0f,
+        /* Radius */ 30.0f,
+    },
+};
+static u8 padding[76];
+
+// Need to break these out to get the rodata ordered right
+static const float partHeightOffset = 20.0;
+static const float partMaxFlickerPerTick = 0.02;
```
magcius, line 45: Please suffix floats with f, 20.0f, 0.02f

src/d/actor/d_a_lamp.cpp
```diff
@@ -4,39 +4,230 @@
...
 
 /* 000000EC-00000158       .text daLamp_Draw__FP10lamp_class */
-void daLamp_Draw(lamp_class*) {
-    /* Nonmatching */
+static BOOL daLamp_Draw(lamp_class* i_lamp) {
+    g_env_light.settingTevStruct(TEV_TYPE_BG0, &i_lamp->current.pos, &i_lamp->mTevStr);
```
magcius, line 49: The preference is to use `i_this` for the C function parameters

PR #546, opened 2023-12-03
d_a_ghostship Mostly Done

SuperDude88:
Only remaining issues are a couple regswaps + unused floats in .data that a bunch of actors have


LagoLunatic:
is there a reason the comments with the mangled symbol names were all removed? personally I think those are a really useful quick reference, since you can't search for e.g. just `create` in the symbol maps and find the function you're looking for since it's too ambiguous when demangled, while the mangled ones are unambiguous.

PR #547, opened 2023-12-07
d_a_arrow_iceeff/lighteff 100%, not OK

SuperDude88:


include/d/actor/d_a_arrow_iceeff.h
```diff
@@ -2,20 +2,32 @@
...
+    J3DModel* field_0x298[30];
+    J3DModel* mpModel;
+    Mtx field_0x314[30];
+    cXyz field_0x8B4[30];
+    J3DAnmTextureSRTKey* field_0xA1C;
+    mDoExt_bckAnm mBck;
+    s32 field_0xA30;
+    f32 field_0xA34;
+    s32 field_0xA38;
+    u8 field_0xA3C;
```
LagoLunatic, line 30: you should add offset comments for every field. if you have the struct defined in ghidra there's a ghidra python script  you can use that generates the list of member variables for the header automatically including the offset comments, which saves a lot of time.

PR #559, opened 2023-12-24
d_a_tag_etc work

Crain-32:


src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,175 @@
...
 void daTag_Etc_c::demoProc() {
-    /* Nonmatching */
+    s16 sVar1;
+    fopAc_ac_c* pfVar2;
+    int staffIdx;
+    u8 cVar3;
+    void* local_18[2];
+
+    local_18[0] = this->field_0x294;
+    pfVar2 = (fopAc_ac_c*)fopAcIt_Judge(fpcSch_JudgeByID, local_18);
```
LagoLunatic, line 89:
```suggestion
    pfVar2 = (daNpc_Md_c*)fopAcM_SearchByID(field_0x294);
```

* field_0x294 is not a void* pointer, it's a process ID
* the fopAcM_SearchByID inline should be used instead of `fopAcIt_Judge(fpcSch_JudgeByID`, you don't need to create the temp stack variable manually that way. you should check the demo debug maps to see what inlines the devs used in this REL.

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,175 @@
...
+    staffIdx = g_dComIfG_gameInfo.play.mEvtManager.getMyStaffId("TAG_ETC_D", (fopAc_ac_c*)0x0, 0);
```
LagoLunatic, line 90:
```suggestion
    staffIdx = dComIfGp_evmng_getMyStaffId("TAG_ETC_D");
```

actors basically never access g_dComIfG_gameInfo directly, you should always use the appropriate dComIfG inline (which you can see in the demo debug maps) unless it doesn't match. same for several other calls like this in this file.

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,175 @@
...
+
+    if (staffIdx != -1) {
+        cVar3 = getType2();
+        switch (cVar3) {
+        case 0:
+            if ((pfVar2 == (fopAc_ac_c*)0x0) ||
+                ((u32)pfVar2[0x13].mBase.mLnTg.mBase.mpTagData & 0x10) == 0)
```
LagoLunatic, line 97: this is totally wrong, this actor is medli, not an array of the actor base class.

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,175 @@
...
 #include "dolphin/types.h"
+#include "f_op/f_op_actor_iter.h"
+#include "f_pc/f_pc_searcher.h"
+#include "global.h"
 
 /* 00000078-00000084       .text getEventNo__11daTag_Etc_cFv */
-void daTag_Etc_c::getEventNo() {
-    /* Nonmatching */
+u8 daTag_Etc_c::getEventNo() {
+    return this->mBase.mParameters >> 0x18;
```
LagoLunatic, line 17:
```suggestion
    return fopAcM_GetParam(this) >> 0x18;
```

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,175 @@
...
-void daTag_Etc_c::rangeCheck(fopAc_ac_c*) {
-    /* Nonmatching */
+BOOL daTag_Etc_c::rangeCheck(fopAc_ac_c* pActor) {
+    BOOL result;
+    float fVar4;
+    float distanceMagnitude;
+    cXyz cxYz_planar_distance;
+    cXyz cXyz_distance;
+
+    cXyz_distance = ((pActor->current).pos - (this->current).pos);
```
LagoLunatic, line 34:
```suggestion
    delta = pActor->current.pos - current.pos;
```

typing `this->` isn't necessary, parentheses aren't necessary when order of operations is obvious, putting the type name in the variable name isn't necessary

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,175 @@
...
+    if (0.0 <= cXyz_distance.x) {
+        cxYz_planar_distance.x = cXyz_distance.x;
+        cxYz_planar_distance.y = 0.0;
+        cxYz_planar_distance.z = cXyz_distance.z;
+        distanceMagnitude = PSVECSquareMag(&cxYz_planar_distance);
+        if (0.0 < distanceMagnitude) {
+            fVar4 = sqrtf(distanceMagnitude);
+        }
```
LagoLunatic, line 42: this whole thing (above 6 lines) is the `absXZ` inline of the cXyz class

Crain-32:
Did some cleanup based on the Feedback. I was surprised how many of my Java Habits/Styles came in, I'll have to make sure to keep an eye on that.

Only comments I have,
I'm still not able to get `rangeCheck` to match, not sure what I'm missing there at this point.
I'm not a fan of `otherCheck`'s look due to the double casting of the Pointer to Medli, but I was unsure if I was allowed to modify the header to check the param type to `daNpc_Md_c`, so I went with the "It matches the diff" approach.

The inlining stuff still throws me off, I was using VSCode for the edits, but I'm going to try out a JetBrains' IDE on it sometime after Christmas, I've found their refactor tooling to be good at finding stuff like that in Java. Otherwise I'll just make a cheat sheet while I learn them all.

LagoLunatic:
> Did some cleanup based on the Feedback. I was surprised how many of my Java Habits/Styles came in, I'll have to make sure to keep an eye on that.
> 
> Only comments I have, I'm still not able to get `rangeCheck` to match, not sure what I'm missing there at this point.

It doesn't look like you pushed any new commits, so I can't check what changed there.

> I'm not a fan of `otherCheck`'s look due to the double casting of the Pointer to Medli, but I was unsure if I was allowed to modify the header to check the param type to `daNpc_Md_c`, so I went with the "It matches the diff" approach.

You shouldn't modify the function argument types, those are extracted directly from the symbol maps.
What do you mean "double casting"? You should have to cast `fopAc_ac_c*` to `daNpc_Md_c*` once, but I don't think a second cast is necessary unless I'm missing something.

> The inlining stuff still throws me off, I was using VSCode for the edits, but I'm going to try out a JetBrains' IDE on it sometime after Christmas, I've found their refactor tooling to be good at finding stuff like that in Java. Otherwise I'll just make a cheat sheet while I learn them all.

I'm not sure what the IDE has to do with inlines, I use VSCode and get inlines in the autocomplete suggestions just fine if that's what you mean.
Are you looking at tag_etc's kiosk demo debug map? Because it should tell you exactly which inlines were actually used in this REL without needing to rely on the editor's guesses based off of names.
The only problem to watch out for is that each function only appears once in each map, the first time it was used (based on link order). So if the same inline gets used in two separate functions it will only show up in the tree underneath one of those functions, not both.

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,145 @@
...
-void daTag_Etc_c::otherCheck(fopAc_ac_c*) {
-    /* Nonmatching */
+BOOL daTag_Etc_c::otherCheck(fopAc_ac_c* pActor) {
+    BOOL result;
+
+    u8 cVar2 = getType2();
+
+    switch (cVar2) {
+    case 0:
+        if (pActor != (daNpc_Md_c*)0x0 && (((daNpc_Md_c*)pActor)->m30F0 & 0x10) != 0) {
```
magcius, line 50:
```suggestion
        if (pActor != NULL && (((daNpc_Md_c*)pActor)->m30F0 & 0x10) != 0) {
```

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,145 @@
...
 }
 
 /* 00000214-000002EC       .text demoProc__11daTag_Etc_cFv */
 void daTag_Etc_c::demoProc() {
-    /* Nonmatching */
+    daNpc_Md_c* pMedli;
+    int staffIdx;
+    u8 cVar3;
+    pMedli = (daNpc_Md_c*)fopAcM_SearchByID(processId);
+    staffIdx = dComIfGp_evmng_getMyStaffId("TAG_ETC_D", (daNpc_Md_c*)0x0, 0);
```
magcius, line 69:
```suggestion
    staffIdx = dComIfGp_evmng_getMyStaffId("TAG_ETC_D", NULL, 0);
```

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,145 @@
...
+
+    if (staffIdx != -1) {
+        cVar3 = getType2();
+        switch (cVar3) {
+        case 0:
+            if ((pMedli == (daNpc_Md_c*)0x0) || (pMedli->m30F0 & 0x10) == 0) {
```
magcius, line 75:
```suggestion
            if ((pMedli == NULL) || (pMedli->m30F0 & 0x10) == 0) {
```

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,145 @@
...
-void daTag_Etc_c::rangeCheck(fopAc_ac_c*) {
-    /* Nonmatching */
+BOOL daTag_Etc_c::rangeCheck(fopAc_ac_c* pActor) {
+    BOOL result;
+    float distanceMagnitude;
+    cXyz delta;
+
+    delta = pActor->current.pos - current.pos;
+    if (delta.y >= 0.0f) {
+        if ((mScale.x * 100) <= delta.absXZ() || (mScale.x * 100) < delta.y) {
```
magcius, line 31:
```suggestion
        if ((mScale.x * 100.0f) <= delta.absXZ() || (mScale.x * 100.0f) < delta.y) {
```

src/d/actor/d_a_tag_etc.cpp
```diff
@@ -4,50 +4,145 @@
...
-    /* Nonmatching */
+    float fVar1;
+    u8 stageEVNTListIndex;
+    u16 uVar2;
+    u8 cVar3;
+
+    fopAcM_SetupActor(this, daTag_Etc_c);
+
+    stageEVNTListIndex = getEventNo();
+    eventIndex = dComIfGp_evmng_getEventIdx((char*)0x0, stageEVNTListIndex);
```
magcius, line 116:
```suggestion
    eventIndex = dComIfGp_evmng_getEventIdx(NULL, stageEVNTListIndex);
```

PR #596, opened 2024-02-01
d_a_obj_doguu_demo

Candygoblen123:
This is my first time doing anything decomposition related, so let me know if I have anything wrong.

### DIFF
--- a/include/d/actor/d_a_obj_doguu_demo.h
+++ b/include/d/actor/d_a_obj_doguu_demo.h
@@ -1,6 +1,8 @@
 #ifndef D_A_OBJ_DOGUU_DEMO_H
 #define D_A_OBJ_DOGUU_DEMO_H
 
+#include "SSystem/SComponent/c_phase.h"
+#include "d/d_bg_w.h"
 #include "f_op/f_op_actor.h"
 
 class daObjDoguuD_c : public fopAc_ac_c {
@@ -10,12 +12,19 @@ public:
     inline BOOL _draw();
     inline BOOL _execute();
 
-    void CreateHeap();
+    BOOL CreateHeap();
     void CreateInit();
     void set_mtx();
 
 public:
-    /* Place member variables here */
+    /* 0x290 */ request_of_phase_process_class mPhs;
+    /* 0x298 */ J3DModel* mpModel;
+    /* 0x29C */ dBgW* mpBgW;
+    /* 0x2A0 */ Mtx mtx;
+    /* 0x2D0 */ u8 mFlag;  // this field seems to be set to 1 only when initalized with CreateInit,
+                           // can probably be a BOOL
+    /* 0x2D1 */ u8 param;
+    /* 0x2D4 */ u32 mShape;
 };
 
 #endif /* D_A_OBJ_DOGUU_DEMO_H */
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_obj_doguu_demo.cpp
```diff
@@ -4,50 +4,145 @@
...
-    /* Nonmatching */
+static BOOL CheckCreateHeap(fopAc_ac_c* i_this) {
+    return ((daObjDoguuD_c*)i_this)->CreateHeap();
 }
 
-/* 00000098-00000214       .text CreateHeap__13daObjDoguuD_cFv */
-void daObjDoguuD_c::CreateHeap() {
-    /* Nonmatching */
+BOOL daObjDoguuD_c::CreateHeap() {
+    this->param = (u8)mBase.mParameters;  // Im not sure why this is here, none of the other classes
```
LagoLunatic, line 21:
```suggestion
    mUnusedParam = fopAcM_GetParam(this) & 0xFF;
```
the parameters field is almost always accessed via fopAcM_GetParam, and if you see what resembles a cast to u8 when parameters are involved, it's probably actually masking out an 8-bit parameter manually like this.

member variables should ideally start with `m` to distinguish them from local variables.
you don't need to type `this->` when assigning to a member variable.

also, as far as I can see this & 0xFF parameter appears unused (correct me if I'm wrong), so it can have unused in its name to make that clearer

src/d/actor/d_a_obj_doguu_demo.cpp
```diff
@@ -4,50 +4,145 @@
...
+        return FALSE;
+
+    return TRUE;
 }
 
-/* 00000214-00000294       .text CreateInit__13daObjDoguuD_cFv */
 void daObjDoguuD_c::CreateInit() {
-    /* Nonmatching */
+    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -30000.0, -5000.0, -30000.0, 30000.0, 40000.0, 30000.0);
```
LagoLunatic, line 49:
```suggestion
    fopAcM_setCullSizeBox(this, -30000.0f, -5000.0f, -30000.0f, 30000.0f, 40000.0f, 30000.0f);
```
single precision floats should be suffixed with `f` to distinguish from double precision floats

include/d/actor/d_a_obj_doguu_demo.h
```diff
@@ -10,12 +12,19 @@ class daObjDoguuD_c : public fopAc_ac_c {
...
     void set_mtx();
 
 public:
-    /* Place member variables here */
+    /* 0x290 */ request_of_phase_process_class mPhs;
+    /* 0x298 */ J3DModel* mpModel;
+    /* 0x29C */ dBgW* mpBgW;
+    /* 0x2A0 */ Mtx mtx;
+    /* 0x2D0 */ u8 mFlag;  // this field seems to be set to 1 only when initalized with CreateInit,
+                           // can probably be a BOOL
```
LagoLunatic, line 25: it can't be `BOOL` (which is a 32-bit type) but it might be `bool` (which is an 8-bit type)
(yes the devs used both types of booleans, and getting the type right affects matches a lot)

LagoLunatic:
if you matched the file fully (I'm assuming so since you removed the `Nonmatching` comments), you can mark the file as matching in configure.py so the match gets checked by the build.

PR #599, opened 2024-02-15
work on d_a_wall

Candygoblen123:
Most things match, with the exception of `_execute` and the missing `dCcD_Tri::__vtable` struct.

### DIFF
--- a/include/d/actor/d_a_wall.h
+++ b/include/d/actor/d_a_wall.h
@@ -10,19 +10,19 @@
 class daWall_c : public fopAc_ac_c {
 public:
     inline daWall_c();
-    BOOL _delete();
+    bool _delete();
     BOOL CreateHeap();
     void CreateInit();
     s32 _create();
     void set_mtx();
     void setMoveBGMtx();
-    BOOL _execute();
+    bool _execute();
     void mode_wait();
     void mode_break();
     void set_tri();
     void set_effect();
     void set_se();
-    BOOL _draw();
+    bool _draw();
 
 public:
     /* 0x290 */ request_of_phase_process_class mPhs;
@@ -34,7 +34,7 @@ public:
     /* 0x5AC */ u8 mState;
     /* 0x5B0 */ dPa_smokeEcallBack mSmokeCb;
     /* 0x5D0 */ JPABaseEmitter* mpEmitter;
-    /* 0x5D4 */ float mpDst;
+    /* 0x5D4 */ float mDst;
     /* 0x5D8 */ u8 mBreakCounter;
     /* 0x5DC */ s32 mSwitchIndex;
     /* 0x5E0 */ u8 mType;
--- a/include/d/actor/d_a_wall.h
+++ b/include/d/actor/d_a_wall.h
@@ -1,12 +1,17 @@
 #ifndef D_A_WALL_H
 #define D_A_WALL_H
 
+#include "SSystem/SComponent/c_phase.h"
+#include "d/d_bg_w.h"
+#include "d/d_cc_d.h"
+#include "d/d_particle.h"
 #include "f_op/f_op_actor.h"
 
 class daWall_c : public fopAc_ac_c {
 public:
+    inline daWall_c();
     BOOL _delete();
-    void CreateHeap();
+    BOOL CreateHeap();
     void CreateInit();
     s32 _create();
     void set_mtx();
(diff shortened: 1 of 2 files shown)


src/d/actor/d_a_wall.cpp
```diff
@@ -4,99 +4,354 @@
...
-    /* Nonmatching */
+static s32 daWall_Create(void* i_this) {
+    return ((daWall_c*)i_this)->_create();
 }
 
-/* 00001064-00001088       .text daWall_Delete__FPv */
-static BOOL daWall_Delete(void*) {
-    /* Nonmatching */
+static BOOL daWall_Delete(void* i_this) {
+    return ((daWall_c*)i_this)->_delete() & 0xFF;
```
LagoLunatic, line 321:
```suggestion
    return ((daWall_c*)i_this)->_delete();
```

when a static method returns a `BOOL` and it has a clrlwi like this, it's probably a sign that the inner method (`_delete`, `_draw`, etc) should return a `bool` as that will create the clrlwi automatically. then you don't need to manually `& 0xFF` which is a fakematch.

src/d/actor/d_a_wall.cpp
```diff
@@ -4,99 +4,354 @@
...
+    case 1:
+    case 2:
+        dComIfGp_particle_setProjection(projection_id[mType], &current.pos, &current.angle, NULL,
+                                        0xFF, NULL, fopAcM_GetRoomNo(this), &mTevStr.mColorK0, &mTevStr.mColorK0);
+        dComIfGp_particle_setProjection(projection_id[mType], &current.pos, &local_28, NULL,
+                                        0xFF, NULL, fopAcM_GetRoomNo(this), &mTevStr.mColorK0, &mTevStr.mColorK0);
+        mpEmitter = dComIfGp_particle_set(particle_id[mType], &current.pos, &current.angle, NULL,
+                                        mpDst, &mSmokeCb, fopAcM_GetRoomNo(this));
+        if (mpEmitter != NULL) {
+            mpEmitter->mFlags = mpEmitter->mFlags | 0x40;
```
LagoLunatic, line 288:
```suggestion
            mpEmitter->becomeImmortalEmitter();
```

include/d/actor/d_a_wall.h
```diff
@@ -20,7 +25,27 @@ class daWall_c : public fopAc_ac_c {
...
+    /* 0x290 */ request_of_phase_process_class mPhs;
+    /* 0x298 */ J3DModel* mpModel;
+    /* 0x29C */ dCcD_Stts mStts;
+    /* 0x2D8 */ dCcD_Tri mTri[2];
+    /* 0x578 */ dBgW* mpBgW;
+    /* 0x57C */ Mtx mMtx;
+    /* 0x5AC */ u8 mState;
+    /* 0x5B0 */ dPa_smokeEcallBack mSmokeCb;
+    /* 0x5D0 */ JPABaseEmitter* mpEmitter;
+    /* 0x5D4 */ float mpDst;
```
LagoLunatic, line 37:
```suggestion
    /* 0x5D4 */ f32 mDst;
```

`mp` is for pointer members. this is a float, not a pointer to a float, so it shouldn't have the p.

src/d/actor/d_a_wall.cpp
```diff
@@ -4,99 +4,354 @@
...
 
 /* 00000D84-00000F74       .text set_effect__8daWall_cFv */
 void daWall_c::set_effect() {
-    /* Nonmatching */
+    u16 projection_id[3] = {0xa16e, 0xa170, 0xa172};
+    u16 particle_id[3] = {0xa16f, 0xa171, 0xa173};
+
+    csXyz local_28 = current.angle;
+    local_28.y += 0x8000;
+    mpDst = 200.00;
```
LagoLunatic, line 275:
```suggestion
    mDst = 200.0f;
```
this is a single not a double so it should have `f`

src/d/actor/d_a_wall.cpp
```diff
@@ -4,99 +4,354 @@
...
 void daWall_c::mode_break() {
-    /* Nonmatching */
+    if (mBreakCounter != 0) {
+        mBreakCounter = mBreakCounter + 1;
+        switch (mType) {
+        case 0:
+        case 1:
+        case 2:
+            int tmp;
+            if ((mBreakCounter > 10) && (tmp = cLib_chaseF(&mpDst, 0.0, 2.222222), tmp != 0)) {
```
LagoLunatic, line 216:
```suggestion
            if ((mBreakCounter > 10) && cLib_chaseF(&mDst, 0.0f, 2.222222f)) {
```
tmp variable shouldn't be necessary I think, unless this fails to match for some reason?
also these are singles not doubles

src/d/actor/d_a_wall.cpp
```diff
@@ -4,99 +4,354 @@
...
+    mStts.Init(0xFF, 0xFF, this);
+    set_tri();
+    dComIfG_Bgsp()->Regist(mpBgW, this);
+    set_mtx();
+    mpBgW->Move();
+    mSwitchIndex = fopAcM_GetParam(this) & 0xFF;
+    mState = false;
 }
 
-/* 00000380-000004EC       .text _create__8daWall_cFv */
```
LagoLunatic, line 29: I'd recommend keeping these comments in, they don't hurt anything and can be good to have as reference sometimes

PR #600, opened 2024-02-17
Initial Cheatsheet created

Crain-32:
Basically just info dumped anything useful that came to mind, or recent conversations we've had.

### DIFF
--- a/cheatsheet.md
+++ b/cheatsheet.md
@@ -1,0 +1,162 @@
+# Cheatsheet
+
+## Common Inlines
+When accessing a field in `g_dComIfG_gameInfo`, there is likely an inline using the format `dComIfG_XXX`,
+check `d_com_inf_game.h`
+
+Common interactions with the `fopAc_ac_c` struct involve the `fopAcM_` inlines. Check `f_op_actor_mng.h`
+Other Actor inlines can be found in `f_pc_manager.h`
+
+Accessing `g_env_light` should use the inline `dKy_getEnvlight()`
+
+The stage based inlines can be found in `d_stage.h`
+
+Audio Inlines are in `m_Do_audio.h`, check when dealing with `mDoAud_zelAudio_c`
+
+The `Mtx` struct has a lot of inlines in the `m_Do_mtx.h` file
+
+*TODO - insert JKR inlines, might be more useful to pair with Ghidra output examples?*
+
+### Finding Inlines
+Make sure to reference the debug map when dealing with functions, especially on the first pass.
+
+This can hint at a local `inline` when you see something like `function__##currentNamespace_varation##currentNamespace`, for example
+`getItemNo__15daSpcItem01_prmFP13daSpcItem01_c`. That same format can hint at another file's inline being referenced.
+
+## Macros
+
+The two main Macros you'll likely see are the `fopAcm_SetupActor` Macro, and the `JUTAssertion` Macros.
+
+### fopAcm_SetupActor
+For `fopAcm_SetupActor` the first hint is being in some sort of `create` function, in Ghidra the macro is already expanded
+so just keep an eye out for something like the following.
+```c++
+if (((this->parent).mCondition & Constructed) == 0) {
+    if(this != (structPointer_c *) 0x0) {
+        fopAc_ac_c::fopAc_ac_c((fopAc_ac_c*) this);
+        // More stuff
+    }
+    (this->parent).mCondition = (this->parent).mCondition | Constructed;
+}
+```
+Everything within the first `if` statement is handled by the `fopAcm_SetupActor` Macro.
+
+### JUTAssertion
+This is more a "family" of Macros, however most of the time we're dealing with `JUT_ASSERT`. Super easy pattern to spot,
+just the following.
+```c++
+if (condition) {
+  tempVar = JUTAssertion::getSDevice();
+  JUTAssertion::showAssert(tempVar, "current_file.cpp", lineNumber, "condition");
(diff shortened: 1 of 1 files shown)


cheatsheet.md
```diff
@@ -0,0 +1,162 @@
...
+b
+;; Aka cmpwi + a lot of branching is likely a small switch statement.
+```
+
+## Register Problems
+When dealing with register mismatching, we've found the following situations to help.
+When dealing with enums, you might have to do a double return. For example
+```c++
+return createInit() ? 4 : 5 
+// Matches, but we know Enums were used in this function.
```
LagoLunatic, line 117: nitpick: technically I don't think we know that 100% for sure (no asserts in this function)

cheatsheet.md
```diff
@@ -0,0 +1,162 @@
...
+// All the values between 0xHigh_Unknown - 0xOffset are still unknown.
+  /* 0xOFFSET  */ u8 field_0xOFFSET;
+  /* 0xOFFSET1 */ u8 field_0xOFFSET1;
+// Similar to the array approach, but defines all the unkown `u8` as their own field instead.
+  /* 0xOFFSET */  type field_0xOFFSET;
+// This field is typed correctly, but has an unknown purpose. `type mOffset` could also convey this. 
+} // Size 0xVALUE
+// ^ Total size of the Struct, could also do the folling
+STATIC_ASSERT(sizeof(MyStruct) == 0xVALUE);
+```
```
LagoLunatic, line 162: this whole code block is really hard to read, I feel like it would be easier for someone to just look at an example of actual code instead of pseudocode, or better yet maybe mention the ghidra script that generates the struct in this format automatically so you don't have to type it out manually

Crain-32:
Been meaning to clean this up, sorry. Hopefully I can get to it today.

Regarding regalloc, it's not a concept I'm familiar with, so I'll try to clean that up as much as I can, but it'll likely be best if I clean it up, we merge, and then someone more familiar with it polishes those sections.

LagoLunatic:
> Regarding regalloc, it's not a concept I'm familiar with, so I'll try to clean that up as much as I can, but it'll likely be best if I clean it up, we merge, and then someone more familiar with it polishes those sections.

[Register allocation](https://en.wikipedia.org/wiki/Register_allocation) is an optimization where the compiler assigns variables to CPU registers to avoid allocating space on the stack and constantly accessing RAM every time a variable is accessed.

This can often cause problems in decomp because the rules for how the compiler determines which variables get assigned to which registers in which order are complex and not fully understood. So you often run into situations where a function has all of its instructions the same as the original function, and it's functionally equivalent to the original function, but which variable got put into which register got all swapped around, so the hash of the function doesn't match. There are a bunch of different patterns you eventually pick up on for knowing how to fix regswaps, maybe I could write a section on it eventually if I ever feel like it.

But the examples currently in the cheatsheet aren't related to registers being wrong, one is related to instructions being out of order, and another is related to the instructions themselves being completely different.

Crain-32:
Updated.

I've honestly never used the Ghidra Script for the Struct definition, so I'd need a reference in order to write that up.

cheatsheet.md
```diff
@@ -0,0 +1,148 @@
...
+
+This can hint at a locally defined `inline` when you see something like `function__##currentNamespace_varation##currentNamespace`,
+for example
+`getItemNo__15daSpcItem01_prmFP13daSpcItem01_c`.
+
+## Macros
+
+The two main Macros you'll likely see are the `fopAcm_SetupActor` Macro, and the `JUTAssertion` Macros.
+
+### fopAcm_SetupActor
```
magcius, line 35: `fopAcM_SetupActor`, beware of case here

cheatsheet.md
```diff
@@ -0,0 +1,162 @@
...
+    }
+    break;
+}
+```
+You can also get single case switch statements instead of `if/else`. For example,
+```c++
+if (m_itemNo == BOKO_BELT) {
+    offsetY = -24.0f;
+}
+// Should be
```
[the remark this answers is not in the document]
magcius: Agreed. I would rephrase this section.

cheatsheet.md
```diff
@@ -0,0 +1,148 @@
...
+    break;
+default:
+    if (mAcch.ChkGroundLanding()) {
+       // stuff 1
+    }
+    break;
+}
+```
+
+You can rarely get single case switch statements instead of `if/else`. For example,
```
magcius, line 100: Please remove this section, or reword it to focus on things like the cmpwi/double branch.

magcius (commented):
All of Lago's feedback still applies here.

LagoLunatic:
Also I feel files with tips like this should go in the ./docs folder instead of root directory

cheatsheet.md
```diff
@@ -0,0 +1,148 @@
...
+You can find the map pinned in the Main Decomp channel in the Discord. You can think of this map as one
+of the last steps of the compiler, where it writes a record of what symbols it used and where it used them. Particularly
+for debugging purposes.
+
+If you go into the `.text section layout` of the Map file, 
+you'll be able to spot what inlines are used, and in what function.
+
+This can hint at a locally defined `inline` when you see something like `function__##currentNamespace_varation##currentNamespace`,
+for example
+`getItemNo__15daSpcItem01_prmFP13daSpcItem01_c`.
```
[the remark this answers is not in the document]
Crain-32: I don't know how to use objdiff's demangler. I'll put in a `TODO` so someone else more familiar could clean that up.

PR #610, opened 2024-03-16
d_a_pw work

Crain-32:
Was checking out Poes for some Jahalla one cycle stuff, but we proved that it is impossible without just manipulating the RNG to have all the Poes move together. Although `action_dousa` is for movement, after `case 0xF` the switch is doing some wacky stuff, so I'm going to leave it to more experienced people.

Here is what I got so far, including `daPW__BehaviorType` from Ghidra.

### DIFF
--- a/src/d/actor/d_a_pw.cpp
+++ b/src/d/actor/d_a_pw.cpp
@@ -4,7 +4,11 @@
 //
 
 #include "d/actor/d_a_pw.h"
+#include "d/actor/d_a_player_main.h"
+#include "d/d_kankyo_rain.h"
 #include "dolphin/types.h"
+#include "f_op/f_op_camera_mng.h"
+#include "f_pc/f_pc_manager.h"
 
 /* 00000078-00000158       .text nodeCallBack__FP7J3DNodei */
 static BOOL nodeCallBack(J3DNode*, int) {
@@ -62,7 +66,7 @@ void BG_check(pw_class*) {
 }
 
 /* 00001990-00001D60       .text Line_check__FP8pw_class4cXyzUc */
-void Line_check(pw_class*, cXyz, unsigned char) {
+BOOL Line_check(pw_class*, cXyz, unsigned char) {
     /* Nonmatching */
 }
 
@@ -87,7 +91,7 @@ void kyori_sub(pw_class*) {
 }
 
 /* 00002714-0000289C       .text hani_check__FP8pw_class */
-void hani_check(pw_class*) {
+BOOL hani_check(pw_class*) {
     /* Nonmatching */
 }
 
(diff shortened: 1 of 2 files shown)


src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
+            // of case 0x14 alone.
+
+        }
+        break;
+    case 0x19:
+    case 0x1A:
+        break;
+    case 0x1B:
+    case 0x1C:
+        if (i_this->actor_status & 0x00100000 == 0) {
```
LagoLunatic, line 320:
```suggestion
        if (!fopAcM_CheckStatus(i_this, fopAcStts_HOOK_CARRY_e)) {
```

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
+        BOOL iVar6 = i_this->m2B4->mFrameCtrl.checkPass(25.0f);
+        if (iVar6 != FALSE) {
+            i_this->m3A4 = 0.0f;
+            first_mode_change(i_this);
+            s16 actorAngleY = fopAcM_searchActorAngleY(i_this, link);
+            i_this->m38C = actorAngleY;
+            JPABaseEmitter* particle =
+                dComIfGp_particle_set(0x82EE, &i_this->m2CC, &i_this->shape_angle);
+            if (particle != NULL) {
+                // JPASetRMtxSTVecfromMtx(i_this->m2B4->mpModel->mpNodeMtxm,
```
LagoLunatic, line 155: probably something like
```suggestion
                `particle->setGlobalSRTMatrix(i_this->m2B4->getModel()->getAnmMtx(0x00));`
```

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
+        if (i_this->m39A > 100) {
+            i_this->m39A = 100;
+        }
+        BOOL tempVar10 = TRUE;
+        if (!(i_this->m2B4->mFrameCtrl.getState() & 1) &&
+            i_this->m2B4->mFrameCtrl.getRate() != 0.0f)
+        {
+            tempVar10 = FALSE;
+        }
+        if (tempVar10) {
```
LagoLunatic, line 173:
```suggestion
        if (i_this->m2B4->isStop()) {
```
can delete the above 6 lines

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
+            i_this->shape_angle.y = i_this->m38C;
+            i_this->m3A4 = -80.0f;
+            i_this->m340 = 1;
+            i_this->m368 += 1;
+        }
+        break;
+
+    case 0x2:
+        i_this->shape_angle.y += 0x1000;
+        cLib_addCalc0(&i_this->m3A4, 1.0, 1.0);
```
LagoLunatic, line 141:
```suggestion
        cLib_addCalc0(&i_this->m3A4, 1.0f, 1.0f);
```

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
 
 /* 00002A90-00003B08       .text action_dousa__FP8pw_class */
-void action_dousa(pw_class*) {
+void action_dousa(pw_class* i_this) {
     /* Nonmatching */
+    daPy_py_c* link = daPy_getPlayerActorClass();
+    float randomValue;
+    switch (i_this->m368) {
+    case 0:
+        anm_init(i_this, 0x23, 7.0, '\x02', 1.0, -1);
```
LagoLunatic, line 120:
```suggestion
        anm_init(i_this, 0x23, 7.0f, J3DFrameCtrl::LOOP_REPEAT_e, 1.0f, -1);
```
all the anm_init calls should probably look something like this. the two floating-pointer numbers are singles not doubles, and the unsigned char argument is the J3DFrameCtrl::Attribute_e enum (the reason I know this is because the anm_init function is very similar across most enemies)

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
+        break;
+
+    case 0x2:
+        i_this->shape_angle.y += 0x1000;
+        cLib_addCalc0(&i_this->m3A4, 1.0, 1.0);
+        i_this->m39A += 7;
+        if (i_this->m39A > 100) {
+            i_this->m39A = 100;
+        }
+        BOOL iVar6 = i_this->m2B4->mFrameCtrl.checkPass(25.0f);
```
LagoLunatic, line 146:
```suggestion
        if (!i_this->m2B4->checkFrame(25.0f)) {
```

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
+        anm_init(i_this, 0x23, 7.0, '\x02', 1.0, -1);
+        i_this->m346 = 0;
+        i_this->m368 += 1;
+        break;
+
+    case 0x1:
+        f32 actor_distance = fopAcM_searchActorDistance(i_this, link);
+        if (actor_distance < 500.0f) {
+            anm_init(i_this, 0x10, 3.0, '\0', 1.0, -1);
+            s16 actorYAngle = fopAcM_searchActorAngleY(i_this, link);
```
LagoLunatic, line 129:
```suggestion
            i_this->m38C = fopAcM_searchActorAngleY(i_this, link);
```
I don't think there's any reason to put this in a temporary local variable before assigning it, unless that's required for this to match?

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +111,253 @@ void first_mode_change(pw_class*) {
...
+            i_this->m38C += (short) randomValue;
+            if(i_this->mPathIndex == 0xFF || i_this->m348 == NULL) {
+                i_this->m368 = 0x10;
+            } else {
+                i_this->m368 = 0x14;
+            }
+        }
+        break;
+    case 0xF:
+        goto PART_OF_CASE_E;
```
LagoLunatic, line 291: FYI this is more likely to be a fallthrough from case E to case F than a goto, I ran into a lot of fallthroughs when doing these massive enemy switch statements before

src/d/actor/d_a_pw.cpp
```diff
@@ -26,7 +36,7 @@ static BOOL daPW_Draw(pw_class*) {
 }
 
 /* 000004D4-00000600       .text anm_init__FP8pw_classifUcfi */
-void anm_init(pw_class*, int, float, unsigned char, float, int) {
+void anm_init(pw_class*, int, float, J3DFrameCtrl::Attribute_e frameCtrl, float, int) {
```
LagoLunatic, line 39: you can't change the types of function arguments from what they originally were as that changes the symbol name.

when I said this argument was the J3DFrameCtrl::Attribute_e enum, I just meant that it can be used at the callsites where `anm_init` is being called, not that `anm_init` itself needs to be changed to enforce the usage of the enum.

the name `frameCtrl` also isn't right, framectrl is the class that owns the attribute enum, not the enum itself.

src/d/actor/d_a_pw.cpp
```diff
@@ -113,21 +123,19 @@ void first_mode_change(pw_class*) {
 /* 00002A90-00003B08       .text action_dousa__FP8pw_class */
 void action_dousa(pw_class* i_this) {
     /* Nonmatching */
-    daPy_py_c* link = daPy_getPlayerActorClass();
+    daPy_py_c* pPlayer = daPy_getPlayerActorClass();
     float randomValue;
     switch (i_this->m368) {
     case 0:
-        anm_init(i_this, 0x23, 7.0, '\x02', 1.0, -1);
+        anm_init(i_this, 0x23, 7.0, J3DFrameCtrl::LOOP_ONCE_RESET_e, 1.0, -1);
```
LagoLunatic, line 130:
```suggestion
        anm_init(i_this, 0x23, 7.0, J3DFrameCtrl::LOOP_REPEAT_e, 1.0, -1);
```

src/d/actor/d_a_pw.cpp
```diff
@@ -107,8 +121,230 @@ void first_mode_change(pw_class*) {
...
+    case 0x9:
+        i_this->m39A = 0;
+        i_this->m39C = 0;
+        i_this->m368 += 1;
+    case 0xA:
+        if (i_this->mBehaviorType == InvisibleAtStart) {
+            i_this->m38C = fopAcM_searchPlayerDistance(i_this);
+        }
+        i_this->m370 = fopAcM_create(PROC_KANTERA, 0xFF000001, &i_this->m2CC, i_this->current.roomNo);
+        if (i_this->m370 != -1) {
```
LagoLunatic, line 206:
```suggestion
        if (i_this->m370 != fpcM_ERROR_PROCESS_ID_e) {
```

Crain-32:
Alright, rebased on master, so this should be 100% up to date now

PR #652, opened 2024-05-19
Minor Miscellaneous Work

SuperDude88:
Couple changes I've had sitting around

### DIFF
--- a/include/d/actor/d_a_player_main_data.h
+++ b/include/d/actor/d_a_player_main_data.h
@@ -38,10 +38,9 @@ extern const Vec l_crawl_lside_front_offset;
 extern const Vec l_crawl_rside_front_offset;
 extern const Vec l_crawl_min_side_offset;
 
-// I believe these won't link properly with the static variables
-//extern dCcD_SrcCyl l_cyl_src;
+extern dCcD_SrcCyl l_cyl_src;
 extern dCcD_SrcCyl l_wind_cyl_src;
 extern dCcD_SrcCyl l_at_cyl_src;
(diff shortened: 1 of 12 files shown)


src/d/actor/d_a_player_main_data.inc
```diff
@@ -73,7 +73,7 @@ static dCcD_SrcCyl l_cyl_src = {
     },
 };
 
-static dCcD_SrcCyl l_wind_cyl_src = {
+dCcD_SrcCyl l_wind_cyl_src = {
```
LagoLunatic, line 76: what's the reasoning for removing this keyword? if you're looking at the green g in objdiff, that's just dtk's assumption for main.dol symbols because framework.map doesn't tell us if it had the static keyword or not. but in RELs they usually do have it.

include/d/actor/d_a_player_main_data.h
```diff
@@ -1,46 +1,47 @@
...
+extern const Vec l_crawl_min_side_offset;
 
-static dCcD_SrcCyl l_cyl_src;
-static dCcD_SrcCyl l_wind_cyl_src;
-static dCcD_SrcCyl l_at_cyl_src;
-static dCcD_SrcCps l_at_cps_src;
-static dCcD_SrcCps l_fan_wind_cps_src;
-static dCcD_SrcSph l_fan_wind_sph_src;
+// I believe these won't link properly with the static variables
+//extern dCcD_SrcCyl l_cyl_src;
```
LagoLunatic, line 42: `extern dCcD_SrcCyl l_cyl_src;` should link fine? I had this fixed locally but forgot to commit it

Issue #389, opened 2023-09-28
d_a_obj_eskban


magcius:
Eskban

LagoLunatic:
Has weak function ordering issues related to dBgS_MoveBgActor::Draw()

PR #653, opened 2024-05-23
d_a_obj_eskban

Aelire:
weak functions are in the right order with `nosyminline on`, but the sections don't match, and there are some destructors (`cCcD_SphAttr`, `cCcD_CylAttr`, `dCcD_Stts`, `cCcD_Stts`, `dBgS_MoveBgActor`) that appear in objdiff but marked UNUSED in the .map files. 
I don't know what's up with that

Had to add some padding for the bss size to match

#389

### DIFF
--- a/include/d/actor/d_a_obj_eskban.h
+++ b/include/d/actor/d_a_obj_eskban.h
@@ -3,26 +3,70 @@
 
 #include "f_op/f_op_actor.h"
 
+#include "d/d_a_obj.h"
+#include "d/d_bg_s_movebg_actor.h"
+#include "d/d_particle.h"
+
 namespace daObjEskban {
-    class Act_c : public fopAc_ac_c {
+
+    static const int DESTROY_VIBRATION_LEN = 35;
+    static const int DESTROY_VIBRATION_SHOCK_FRAME_IDX = 28;
+    static const int DESTROY_SMOKE_ANM_LEN = 20;
+
+    class Act_c : public dBgS_MoveBgActor {
     public:
-        void prm_get_swSave() const {}
-    
-        void CreateHeap();
-        s32 Create();
-        void Mthd_Create();
-        BOOL Delete();
-        void Mthd_Delete();
+        static const char M_arcname[7];
+        static const char M_evname[7];
+        static Mtx M_tmp_mtx;
+
+        enum Prm_e {
+            PRM_SWSAVE_W = 0x08,
+            PRM_SWSAVE_S = 0x00,
+        };
+        s32 param_get_swSave() const {
+            return daObj::PrmAbstract<Prm_e>(this, PRM_SWSAVE_W, PRM_SWSAVE_S);
+        }
+
+        enum actor_state {
+            ST_WAIT = 0,
+            ST_DESTROYED = 1,
+            ST_CUTSCENING = 2,
+            ST_VIBRATING = 3,
+            ST_SMOKING = 4,
+        };
+
+        Act_c();
+
+        int CreateHeap();
+        int Create();
+        s32 Mthd_Create();
+        int Delete();
+        BOOL Mthd_Delete();
         void set_mtx();
         void init_mtx();
         void eff_m_break(unsigned short, unsigned short);
         void eff_b_break(unsigned short);
         void daObjEskban_effect_set();
-        void Execute(float(**)[3][4]);
-        BOOL Draw();
-    
+        int Execute(Mtx**);
+        int Draw();
+
     public:
-        /* Place member variables here */
+        /* 0x2C4 Act_c vtable */
+        /* 0x2C8 */ dPa_smokeEcallBack* M_smoke;  // name used in assert
+        /* 0x2CC */ cXyz mSmokePos;
+        /* 0x2D8 */ request_of_phase_process_class mPhs;
+        /* 0x2E0 */ J3DModel* mpModel;
+        /* 0x2E4 */ dCcD_Stts mCheckStts;
+        /* 0x320 */ dCcD_Cyl mCheckCyl;
+        /* 0x450 */ dCcD_Stts mCameraStts;
+        /* 0x48C */ dCcD_Cyl mCameraCyl;
+        /* 0x5BC */ u32 mActorID;
+        /* 0x5C0 */ dCcD_Stts mCheckSphStts;
+        /* 0x5FC */ dCcD_Sph mCheckSph;
+        /* 0x728 */ s32 mActorState;
+        /* 0x72C */ s32 mRemainingSmokeAnm;
+        /* 0x730 */ s32 mRemainingVibration;
+        /* 0x734 */ u8 mIsVisible;
     };
 };
 
--- a/include/d/d_com_inf_game.h
+++ b/include/d/d_com_inf_game.h
@@ -3163,6 +3163,10 @@ inline void dComIfGp_particle_calcMenu() {
     g_dComIfG_gameInfo.play.getParticle()->calcMenu();
 }
 
+inline int dComIfGp_particle_addModelEmitter(dPa_modelEmitter_c* emitter){
+    return g_dComIfG_gameInfo.play.getParticle()->addModelEmitter(emitter);
+}
+
 inline void dComIfGp_particle_drawModelParticle() {
     g_dComIfG_gameInfo.play.getParticle()->drawModelParticle();
 }
--- a/src/d/actor/d_a_obj_eskban.cpp
+++ b/src/d/actor/d_a_obj_eskban.cpp
@@ -226,7 +226,7 @@ void daObjEskban::Act_c::eff_m_break(u16 particleID, u16 prm_b) {
     if (!pMdlEmtr) {
         return;
     }
-    cLs_Addition(g_dComIfG_gameInfo.play.getParticle()->mModelControl, pMdlEmtr);
+    dComIfGp_particle_addModelEmitter(pMdlEmtr);
 }
 
 /* 00000C80-00000D44       .text eff_b_break__Q211daObjEskban5Act_cFUs */
--- a/src/d/actor/d_a_obj_eskban.cpp
+++ b/src/d/actor/d_a_obj_eskban.cpp
@@ -3,94 +3,416 @@
 // Translation Unit: d_a_obj_eskban.cpp
 //
 
-#include "d/actor/d_a_obj_eskban.h"
+#include "f_op/f_op_actor_mng.h"
+
 #include "d/d_procname.h"
+#include "d/d_com_inf_game.h"
+#include "d/actor/d_a_obj_eskban.h"
+#include "d/res/res_eskban.h"
+
+/* .bss alignment */
+static u8 dummy[0x4c];
+
+namespace daObjEskban {
+/* .data alignment, seemingly unused by this unit */
+static Vec _data_align[4] = {
+    {1, 1, 1},
+    {1, 1, 1},
+    {9.40453e-38 /* possibly not actually a float? */, 0, 2.125},
+    {0, 1.75, 0},
+};
+
+Mtx Act_c::M_tmp_mtx;
+const char Act_c::M_arcname[7] = "Eskban";
+const char Act_c::M_evname[7] = "Eskban";
+
+}  // namespace daObjEskban
+
+static dCcD_SrcCyl cyl_check_src = {
+    // dCcD_SrcGObjInf
+    {
+        /* Flags             */ 0,
+        /* SrcObjAt  Type    */ 0,
+        /* SrcObjAt  Atp     */ 0,
+        /* SrcObjAt  SPrm    */ 0,
+        /* SrcObjTg  Type    */ AT_TYPE_BOMB,
+        /* SrcObjTg  SPrm    */ TG_SPRM_SET | TG_SPRM_IS_OTHER,
+        /* SrcObjCo  SPrm    */ 0,
+        /* SrcGObjAt Se      */ 0,
+        /* SrcGObjAt HitMark */ 0,
+        /* SrcGObjAt Spl     */ 0,
+        /* SrcGObjAt Mtrl    */ 0,
+        /* SrcGObjAt SPrm    */ 0,
+        /* SrcGObjTg Se      */ 0,
+        /* SrcGObjTg HitMark */ 0,
+        /* SrcGObjTg Spl     */ 0,
+        /* SrcGObjTg Mtrl    */ 0,
+        /* SrcGObjTg SPrm    */ 0,
+        /* SrcGObjCo SPrm    */ 0,
+    },
+    // cM3dGCylS
+    {{
+        /* Center */ {0.0f, 0.0f, 0.0f},
+        /* Radius */ 200.0f,
+        /* Height */ 600.0f,
+    }},
+};
+static dCcD_SrcCyl cyl_camera_src = {
+    // dCcD_SrcGObjInf
+    {
+        /* Flags             */ 0,
+        /* SrcObjAt  Type    */ 0,
+        /* SrcObjAt  Atp     */ 0,
+        /* SrcObjAt  SPrm    */ 0,
+        /* SrcObjTg  Type    */ 0,
+        /* SrcObjTg  SPrm    */ 0,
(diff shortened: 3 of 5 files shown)


LagoLunatic:
Have you tried marking it as "Matching" in configure.py? In theory it should match fine if all the functions are at 100% and they're in the right order. The section splitting being different doesn't affect the binary, and those destructors you listed are stripped out by the linker.

Aelire:
I hadn't tried that, didn't realize it added more checks
Unfortunately it doesn't match, I misread and there is actually one function in the wrong place, `dBgS_MoveBgActor::Draw` and `daObjEskban::<unnamed>::Mthd_IsDelete` are swapped.

thanks for the explanations for the destructors btw, so if I understand correctly objdiff is comparing a built but not linked object, with an object reconstructed from debug info + the linked .rel, so anything thrown out by the linker would show as a mismatch in objdiff I guess?
But I could bindiff the final rel inside `build/` with the original extracted(+decompressed) rel to check matching to be sure

LagoLunatic:
> hadn't tried that, didn't realize it added more checks
> Unfortunately it doesn't match, I misread and there is actually one function in the wrong place, `dBgS_MoveBgActor::Draw` and `daObjEskban::<unnamed>::Mthd_IsDelete` are swapped.

same issue as obj_ladder and obj_mtest then

> thanks for the explanations for the destructors btw, so if I understand correctly objdiff is comparing a built but not linked object, with an object reconstructed from debug info + the linked .rel, so anything thrown out by the linker would show as a mismatch in objdiff I guess?

yes

> But I could bindiff the final rel inside `build/` with the original extracted(+decompressed) rel to check matching to be sure

yep, for RELs that's how you tell what differences are introduced during linking.
for main.dol, you can just run `ninja diff`

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,105 +3,451 @@
...
-    /* Nonmatching */
+void Mthd_Draw(void* i_this) {
+    ((Act_c*)i_this)->MoveBGDraw();
 }
 
-/* 00001540-0000156C       .text Mthd_IsDelete__Q211daObjEskban28@unnamed@d_a_obj_eskban_cpp@FPv */
-void Mthd_IsDelete(void*) {
-    /* Nonmatching */
+/* 00001540-0000156C       .text Mthd_IsDelete__Q211daObjEskban28@unnamed@d_a_obj_eskban_cpp@FPv
+ */
```
LagoLunatic, line 439: can you disable the autoformatter that's randomly adding/removing linebreaks in autogenerated lines of code like this? I often do regex searches on these and it's easier if they don't get modified from how they were originally

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,105 +3,451 @@
...
+        {},         // dCcD_SrcGobjTg mGObjTg
+        {},         // dCcD_SrcGObjCo mGObjCo
+    },
+    {
+        {
+            {0, 0, 0},  // mCenter
+            300,        // mRadius
+        }  // cM3dGSphS mSph
+    }  // cCcD_SrcSphAttr mSphAttr
+};
```
LagoLunatic, line 138: I feel like these srcsphs/srccyls are difficult to read with this much nesting and indentation?
I usually put the attributes in a flat list with the field names on the left instead of the right (e.g. in d_a_nh.cpp)

Aelire: this was initially to help me learn the structs but definitely too verbose - I'll follow your example

src/d/actor/d_a_obj_eskban.cpp
Aelire: I'll disable it for this block with a comment (`// clang-format off`, I see there are already a couple other instances of it in the repo), but I think it could be globally deactivated for these comments that start with an offset range by adding it to [CommentPragmas](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#commentpragmas)) in the `.clang-format` https://github.com/zeldaret/tww/blob/37d09d16577fdefb9bafa50dbb3da07980bc4e62/.clang-format#L39
```diff
-CommentPragmas:  '^ (IWYU pragma:|NOLINT)'
+CommentPragmas:  '^ (IWYU pragma:|NOLINT|\d{8}-\d{8}\b)'
```

LagoLunatic: I think the clang stuff was copied from TP decomp, I don't think we use that here at all

magcius: I removed it.

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,421 @@
...
+        mDoAud_seStart(JA_SE_READ_RIDDLE_1);
+        daObjEskban_effect_set();
+        break;
+    case ST_SMOKING: /* 0x350 */
+        if (mRemainingSmokeAnm > 0) {
+            mRemainingSmokeAnm--;
+            break;
+        } else {
+            mActorState = ST_WAIT;
+            dComIfGp_event_onEventFlag(DESTROY_EVT_FLAG);
```
magcius, line 347: dComIfGp_event_reset

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,421 @@
...
+            dComIfGp_getVibration().StartQuake(4, ~0x20, cXyz(0, 1, 0));
+        }
+        if (mRemainingSmokeAnm > 0) {
+            mRemainingSmokeAnm--;
+            break;
+        }
+        if (dComIfGp_evmng_endCheck("Eskban") && (this->M_smoke->isEnd())) {
+            dComIfGp_getVibration().StopQuake(-1);
+            fopAcM_delete(this);
+            dComIfGp_event_onEventFlag(DESTROY_EVT_FLAG);
```
magcius, line 374: dComIfGp_event_reset

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,421 @@
...
 }
 
 /* 00000A68-00000B00       .text set_mtx__Q211daObjEskban5Act_cFv */
 void daObjEskban::Act_c::set_mtx() {
-    /* Nonmatching */
+    this->shape_angle = this->current.angle;
+    mDoMtx_stack_c::transS(this->current.pos);
+    mDoMtx_stack_c::ZXYrotM(this->shape_angle);
+    mpModel->setBaseTRMtx(mDoMtx_stack_c::get());
+    mDoMtx_copy(mDoMtx_stack_c::get(), this->M_tmp_mtx);
```
magcius, line 205: Remove unnecessary this->

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,421 @@
...
+    M_smoke = new dPa_smokeEcallBack();
+    JUT_ASSERT(264, M_smoke != 0);
+    return mpModel != NULL;
 }
 
 /* 000001FC-00000368       .text Create__Q211daObjEskban5Act_cFv */
-s32 daObjEskban::Act_c::Create() {
-    /* Nonmatching */
+int daObjEskban::Act_c::Create() {
+    this->cullMtx = mpModel->getBaseTRMtx();
```
magcius, line 135: fopAcM_SetMtx

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,421 @@
...
+    }
+    return phase_state;
 }
 
+daObjEskban::Act_c::Act_c() {}
+
 /* 000009C0-00000A10       .text Delete__Q211daObjEskban5Act_cFv */
 BOOL daObjEskban::Act_c::Delete() {
-    /* Nonmatching */
+    if (this->M_smoke) {
```
magcius, line 183: Remove unnecessary this->

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,421 @@
...
+    if (!pBEmtr) {
+        return;
+    }
+    pBEmtr->setGlobalRTMatrix(mpModel->getBaseTRMtx());
+    dPa_J3DmodelEmitter_c* pMdlEmtr =
+        new dPa_J3DmodelEmitter_c(pBEmtr, mdlData, this->tevStr, txPattern, prm_b, 0);
+    if (!pMdlEmtr) {
+        return;
+    }
+    cLs_Addition(g_dComIfG_gameInfo.play.getParticle()->mModelControl, pMdlEmtr);
```
magcius, line 233: I think this should be in a dComIfGp_particle_addModelEmitter wrapper.

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,421 @@
...
+    pBEmtr->setGlobalParticleScale(p_scale);
 }
 
 /* 00000EF0-00001400       .text Execute__Q211daObjEskban5Act_cFPPA3_A4_f */
-void daObjEskban::Act_c::Execute(float(**)[3][4]) {
-    /* Nonmatching */
+int daObjEskban::Act_c::Execute(Mtx** pMtx) {
+    dComIfG_Ccsp()->Set(&this->mCheckCyl);
+    dComIfG_Ccsp()->Set(&this->mCameraCyl);
+    dComIfG_Ccsp()->Set(&this->mCheckSph);
```
magcius, line 282: Remove unnecessary this->

src/d/actor/d_a_obj_eskban.cpp
Aelire: I removed it for `M_tmp_mtx` and everything that is an obvious class member

For the snake_case unprefixed variables inherited from `fop_Ac_c`, it feels useful to not confuse them with local variables, but I can also remove those if you prefer

src/d/actor/d_a_obj_eskban.cpp
Aelire: ah ok, I removed the changes it made to the skeleton and disabled it locally then

include/d/d_com_inf_game.h
```diff
@@ -3163,6 +3163,10 @@ inline void dComIfGp_particle_calcMenu() {
     g_dComIfG_gameInfo.play.getParticle()->calcMenu();
 }
 
+inline int dComIfGp_particle_addModelEmitter(dPa_J3DmodelEmitter_c* emitter){
+    return cLs_Addition(g_dComIfG_gameInfo.play.getParticle()->mModelControl, emitter);
+}
```
LagoLunatic, line 3168: FYI, if you check the debug map for this actor, `d_a_obj_eskbanD.map`, you can see three nested inlines are used here:
```cpp
           10] dComIfGp_particle_addModelEmitter__FP18dPa_modelEmitter_c (func,weak) found in d_a_obj_eskban.o 
            11] addModelEmitter__13dPa_control_cFP18dPa_modelEmitter_c (func,weak) found in d_a_obj_eskban.o 
             12] add__18dPa_modelControl_cFP18dPa_modelEmitter_c (func,weak) found in d_a_obj_eskban.o 
>>> SYMBOL NOT FOUND: cLs_Addition__FP15node_list_classP10node_class
```

So `dComIfGp_particle_addModelEmitter` would call `g_dComIfG_gameInfo.play.getParticle()->addModelEmitter`, then that would call `mModelControl->add`, and that would call `cLs_Addition`.

In some cases like this using the inlines the same way as the debug map doesn't really matter as it doesn't affect the match. But there are also some cases where you have to copy the way they used inlines exactly for the function to match 100% so I usually do that.

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,416 @@
...
+        break;
+    case ST_DESTROYED: /* 0x1d8 */
+        if (eventInfo.checkCommandDemoAccrpt()) {
+            mActorState = ST_CUTSCENING;
+            break;
+        }
+        fopAcM_orderOtherEvent(this, "Eskban");
+        break;
+    case ST_CUTSCENING: /* 0x214 */
+        if (!fopAcM_SearchByID(mActorID)) {
```
LagoLunatic, line 319:
```suggestion
        if (fopAcM_SearchByID(mActorID) == NULL) {
```
this returns a pointer to an actor, not a boolean

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,416 @@
...
-    /* Nonmatching */
+int daObjEskban::Act_c::Execute(Mtx** pMtx) {
+    dComIfG_Ccsp()->Set(&mCheckCyl);
+    dComIfG_Ccsp()->Set(&mCameraCyl);
+    dComIfG_Ccsp()->Set(&mCheckSph);
+    if (mCheckSph.ChkCoHit()) {
+        cCcD_Obj* hitObj = mCheckSph.GetCoHitObj();
+        if (hitObj) {
+            fopAc_ac_c* hitAct = hitObj->GetAc();
+            if (hitAct && fpcM_GetName(hitAct) == PROC_NPC_MD) {
```
LagoLunatic, line 283:
```suggestion
            if (hitAct && fopAcM_GetName(hitAct) == PROC_NPC_MD) {
```

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,416 @@
...
+        if (!mCameraCyl.ChkCoHit()) {
+            break;
+        }
+        cCcD_Obj* hitObj;
+        if (!(hitObj = mCameraCyl.GetCoHitObj())) {
+            break;
+        }
+        fopAc_ac_c* hitAct = hitObj->GetAc();
+        if (hitAct && fpcM_GetName(hitAct) == PROC_Bomb2) {
+            mActorID = fpcM_GetID(hitAct);
```
LagoLunatic, line 306:
```suggestion
            mActorID = fopAcM_GetID(hitAct);
```

src/d/actor/d_a_obj_eskban.cpp
```diff
@@ -3,94 +3,416 @@
...
+    case ST_WAIT: /* 0x148 */
+        if (!mCameraCyl.ChkCoHit()) {
+            break;
+        }
+        cCcD_Obj* hitObj;
+        if (!(hitObj = mCameraCyl.GetCoHitObj())) {
+            break;
+        }
+        fopAc_ac_c* hitAct = hitObj->GetAc();
+        if (hitAct && fpcM_GetName(hitAct) == PROC_Bomb2) {
```
LagoLunatic, line 305:
```suggestion
        if (hitAct && fopAcM_GetName(hitAct) == PROC_Bomb2) {
```

include/d/actor/d_a_obj_eskban.h
```diff
@@ -3,26 +3,70 @@
...
+        /* 0x2C4 Act_c vtable */
+        /* 0x2C8 */ dPa_smokeEcallBack* M_smoke;  // name used in assert
+        /* 0x2CC */ cXyz mSmokePos;
+        /* 0x2D8 */ request_of_phase_process_class mPhs;
+        /* 0x2E0 */ J3DModel* mpModel;
+        /* 0x2E4 */ dCcD_Stts mCheckStts;
+        /* 0x320 */ dCcD_Cyl mCheckCyl;
+        /* 0x450 */ dCcD_Stts mCameraStts;
+        /* 0x48C */ dCcD_Cyl mCameraCyl;
+        /* 0x5BC */ s32 mActorID;
```
LagoLunatic, line 63:
```suggestion
        /* 0x5BC */ uint mActorID;
```

src/d/actor/d_a_obj_eskban.cpp
Aelire: checking for nullness with `!ptr` is pretty normal c++(98); I do it in other places in this file (and many other files do the same). Do you want the explicit `== NULL` check because this is a return value from a function?

tbh I'm having a bit of a hard time figuring out style expectations without written guidelines even when looking at other files, so apologies if a lot of this doesn't look like it should - hopefully it'll be smoother next time

LagoLunatic: ah alright, I just assumed it was a mistake. in asserts the original devs pretty much always check `var != NULL` (which ends up as `var != 0` in the strings because of double defines), but for if statements it shouldn't matter I think

PR #672, opened 2024-07-24
d_vibration & d_vib_pattern matching

Aelire:
I'm not sure what's up with `dolzel.pch` and whether the class is from this header or imported from somewhere else. I can't really find references to it anywhere else

There seems to be only one inline method and it's not completely obvious to me what precisely it could do, so I took a guess but I kinda doubt it's correct; should I leave it empty instead (it doesn't affect matching whether it's used or not)?


include/d/d_vibration.h
```diff
@@ -1,45 +1,73 @@
...
+        RUMBLE_QUAKE = 0x2,
+    };
+
+    /*
+    These weak functions in the debug maps point to an anonymous struct from a precompiled header ?
+    dVibration_c::@class$3553dolzel_pch::@class$3553dolzel_pch()
+    dVibration_c::@class$3553dolzel_pch::@class$3554dolzel_pch::@class$3554dolzel_pch()
+    dVibration_c::@class$3553dolzel_pch::@class$3554dolzel_pch::~@class$3554dolzel_pch()
+    dVibration_c::@class$3553dolzel_pch::~@class$3553dolzel_pch()
+    */
```
LagoLunatic, line 32: these anonymous structs seem to be present in TP debug with only minor differences to WW retail, they seem pretty similar to your camera_rumble/motor_rumble except I think they're nested structs inside of each other or something. not really sure exactly what "dolzel.pch" would have been supposed to be though

include/d/d_vibration.h
```diff
@@ -49,6 +77,8 @@ class dVibration_c {
     void Init();
     void Pause();
     void Remove();
+
+    inline bool testShake() { return mMotorQuake.mPatternIdx != -1; }
```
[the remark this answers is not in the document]
Aelire: I'll leave it empty then, I don't have access to TP debug

include/d/d_vibration.h
LagoLunatic: for TP debug the cXyz seems to be at offset 0x18 in the inner struct, and the outer struct contains two instances of the inner struct, rather than one.
image
image

it looks like TP decomp has it like this:
https://github.com/zeldaret/tp/blob/6bf2f3cc5578f3e3108df8337a462ee4c618915a/include/d/d_vibration.h#L27-L40

so maybe something like this:
```cpp
    struct {
        struct {
            /* 0x00 */ int  mPatternIdx;
            /* 0x04 */ u32  mPattern;
            /* 0x08 */ s32  mLength;
            /* 0x0C */ s32  mRounds;
            /* 0x10 */ s32  mFlags;
            /* 0x14 */ cXyz mCoord;
            /* 0x20 */ s32  mCurrentFrame;
        } mShock, mQuake;
    } /* 0x00 */ mCamera;
```

this seems to get the constructors and destructors very close to what debug shows.

and then for consistency, motor was probably like this (even though it doesn't show in debug because there's no cXyz to construct/deconstruct):
```cpp
    struct {
        struct {
          /* 0x00 */ int mPatternIdx;
          /* 0x04 */ u32 mPattern;
          /* 0x08 */ s32 mLength;
          /* 0x0C */ s32 mRounds;
          /* 0x10 */ s32 mCurrentFrame;
          /* 0x14 */ s32 mStopFrame; /* different from length for looping rumble */
        } mShock, mQuake;
    } /* 0x48 */ mMotor;
```

Aelire: Makes a lot more sense, thanks for finding the reference! I've updated accordingly

PR #677, opened 2024-09-04
Progress categories support

encounter:
Was using tww as a test-bed for https://github.com/encounter/dtk-template/pull/33

I can revert the changes to the libraries and set them on the individual Objects instead (e.g. `Object("foo", progress_category="game")`), so let me know.

CI now uses the built-in objdiff report functionality in dtk-template. I assume your website script doesn't care about the report artifact name, but if it does, I can revert that change. (My upcoming progress service does match by artifact name)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,20 +21,19 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        submodules: recursive
     - name: Git config
       run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
     - name: Prepare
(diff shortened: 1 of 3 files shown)


LagoLunatic:
> I assume your website script doesn't care about the report artifact name, but if it does, I can revert that change. (My upcoming progress service does match by artifact name)

It does, if you scroll down in build.yml it matches on the name being `GZLE01-report-${{ github.sha }}`. I included the hash because I thought download-artifact needs you to specify exactly which one to download, but I think I might have misunderstood the documentation?

You can probably remove the `-${{ github.sha }}` from there too and we can see if it still works.

encounter:
Don't know how I missed that! Updated. upload/download-artifact are scoped to the current workflow run, so no sha is needed.

PR #678, opened 2024-09-08
d_a_sail 100%

LDAsuku:
it matches

most of the variable names totally suck, but I've taken care to use inlines and avoid obvious fakematches.

the unused bss pattern seems identical to the one in `d_a_tag_msg` and `d_a_movie_player` so I copied it from there.

lmk what you think

### DIFF
--- a/include/d/actor/d_a_sail.h
+++ b/include/d/actor/d_a_sail.h
@@ -31,7 +31,10 @@ public:
     virtual void draw();
 
 public:
-    /* 0x0010 */ u8 m0010[0x00A4 - 0x0010];
+    /* 0x0010 */ Mtx m0010;
+    /* 0x0040 */ Mtx m0040;
+    /* 0x0070 */ Mtx m0070;
+    /* 0x00A0 */ dKy_tevstr_c *mA0;
     /* 0x00A4 */ cXyz m00A4[0x54 * 2];
     /* 0x0884 */ cXyz m0884[0x54];
     /* 0x0C74 */ cXyz m0C74[0x54 * 2];
@@ -42,9 +45,25 @@ public:
     /* 0x1C3A */ u8 m1C3A;
     /* 0x1C3B */ u8 m1C3B;
     /* 0x1C3C */ J3DModel* m1C3C;
-    /* 0x1C40 */ u8 m1C40[0x1C44 - 0x1C40];
+    /* 0x1C40 */ s32 m1C40;
     /* 0x1C44 */ f32 m1C44;
     /* 0x1C48 */ f32 m1C48;
+    /* 0x1C4C */ f32 m1C4C;
+    /* 0x1C50 */ s16 m1C50;
+    /* 0x1C52 */ s16 m1C52;
+
+    inline Mtx *get_m0010() { return &this->m0010; }
+    inline Mtx *get_m0070() { return &this->m0070; }
+
+    inline cXyz* get_pos(s32 idx) {
+        return this->m00A4 + 0x54 * this->m1C3A + 0x54 * idx;
+    }
+    inline cXyz* get_nrm1(s32 idx) {
+        return this->m0C74 + 0x54 * this->m1C3A + 0x54 * idx;
+    }
+    inline cXyz* get_nrm2(s32 idx) {
+        return this->m1454 + 0x54 * this->m1C3A + 0x54 * idx;
+    }
 };
 
 class sail_class : public fopAc_ac_c {
@@ -52,7 +71,6 @@ public:
     /* 0x0290 */ request_of_phase_process_class mClothPhase;
     /* 0x0298 */ request_of_phase_process_class mKaizokusenPhase;
     /* 0x02A0 */ daSail_packet_c mSailPacket;
-    /* 0x1EEC */ u8 m1EEC[0x1EF4 - 0x1EEC];
 };
 
 class daSail_HIO_c : public JORReflexible {
@@ -73,7 +91,9 @@ public:
     /* 0x05 */ u8 m05;
     /* 0x06 */ u8 m06;
     /* 0x07 */ u8 m07;
-    /* 0x08 */ u8 m08[0x10 - 0x08];
+    /* 0x08 */ u8 m08;
+    /* 0x09 */ u8 m09[0x0C - 0x09];
+    /* 0x0C */ f32 m0C;
     /* 0x10 */ f32 m10;
 };
 
--- a/include/d/actor/d_a_sail.h
+++ b/include/d/actor/d_a_sail.h
@@ -15,26 +15,27 @@ public:
         m1C36 = 0;
         m1C3B = 1;
     }
-    
-    void getMtx() {}
-    void getNrm() {}
-    void getPos() {}
-    void getPosSpd() {}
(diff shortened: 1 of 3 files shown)


include/d/actor/d_a_sail.h
```diff
@@ -42,17 +45,32 @@ class daSail_packet_c : public J3DPacket {
...
+
+    inline cXyz* get_pos(s32 idx) {
+        return this->m00A4 + 0x54 * this->m1C3A + 0x54 * idx;
+    }
+    inline cXyz* get_nrm1(s32 idx) {
+        return this->m0C74 + 0x54 * this->m1C3A + 0x54 * idx;
+    }
+    inline cXyz* get_nrm2(s32 idx) {
+        return this->m1454 + 0x54 * this->m1C3A + 0x54 * idx;
+    }
```
LagoLunatic, line 66: instead of coming up with custom names for inlines, it's usually preferred to use the official names from the debug maps.
for this class, I've already added stub inlines with the correct names higher up in the file, starting at line 19.

also, typing `this->` is unnecessary (applies in several other places in this actor too)

include/d/actor/d_a_sail.h
```diff
@@ -42,17 +45,32 @@ class daSail_packet_c : public J3DPacket {
...
 };
 
 class sail_class : public fopAc_ac_c {
 public:
     /* 0x0290 */ request_of_phase_process_class mClothPhase;
     /* 0x0298 */ request_of_phase_process_class mKaizokusenPhase;
     /* 0x02A0 */ daSail_packet_c mSailPacket;
-    /* 0x1EEC */ u8 m1EEC[0x1EF4 - 0x1EEC];
```
LagoLunatic, line 55: this placeholder field was added to pad the size of `sail_class` to the correct size in its actor profile. removing this will make the size be incorrect.

src/d/actor/d_a_sail.cpp
```diff
@@ -438,8 +456,39 @@ u8 l_AmastDL[] ALIGN_DECL(32) = {
 };
 
 /* 000000EC-00000364       .text setCorrectNrmAngle__15daSail_packet_cFsf */
-void daSail_packet_c::setCorrectNrmAngle(short, float) {
-    /* Nonmatching */
+void daSail_packet_c::setCorrectNrmAngle(s16 param_0, f32 param_1) {
+    s16 rnd = 900 + (s32)cM_rndF(200.0f);
+    this->m1C38 += rnd;
```
LagoLunatic, line 461: is the temp variable necessary for the match? or does this work too
```suggestion
    this->m1C38 += 900 + (s32)cM_rndF(200.0f);
```

src/d/actor/d_a_sail.cpp
```diff
@@ -438,8 +456,39 @@ u8 l_AmastDL[] ALIGN_DECL(32) = {
...
+
+    this->m1C34 = 300.0f * JMASSin(this->m1C38);
+
+    s16 r28 = param_0 + 0x8000;
+    s32 r27 = param_0;
+
+    s16 r26 = l_HIO.m10 * (1.0f - 0.5f * param_1);
+    s32 r0 = abs(r28);
+    s16 r30 = 1.25f * r26 * 182.04445f;
```
LagoLunatic, line 470: do either of these inlines match for all the conversions here? would make it a bit more readable if so
```suggestion
    s16 r30 = 1.25f * cAngle::d2s(r26);
```
```suggestion
    s16 r30 = 1.25f * cM_deg2s(r26);
```

src/d/actor/d_a_sail.cpp
```diff
@@ -449,39 +498,445 @@ void daSail_packet_c::setNrmMtx() {
 
 /* 00000398-00000424       .text setBackNrm__15daSail_packet_cFv */
 void daSail_packet_c::setBackNrm() {
-    /* Nonmatching */
+    cXyz* nrm1 = this->get_nrm1(0);
+    cXyz* nrm2 = this->get_nrm2(0);
+    for (int i = 0; i < 0x54; i++) {
+        nrm2->setall(0.0f);
+        VECSubtract(nrm2, nrm1, nrm2);
```
LagoLunatic, line 505:
```suggestion
        nrm2 -= nrm1;
```

src/d/actor/d_a_sail.cpp
```diff
@@ -449,39 +498,445 @@ void daSail_packet_c::setNrmMtx() {
...
+    spC8.set(vtxPos[param_1 + idx]);
+    spD4.setall(0.0f);
+
+    if (param_1 != 0) {
+        spF8 = vtxPos[idx - 1 + param_1] - spC8;
+        if (param_2 != 0) {
+            spEC = vtxPos[(param_2 - 1) * 7 + param_1] - spC8;
+            spE0 = spEC.outprod(spF8);
+            spE0 = spE0.normZC();
+            VECAdd(&spD4, &spE0, &spD4);
```
LagoLunatic, line 529:
```suggestion
            spD4 += spE0;
```
also applies to other instances of VECAdd in this function

src/d/actor/d_a_sail.cpp
```diff
@@ -438,8 +456,39 @@ u8 l_AmastDL[] ALIGN_DECL(32) = {
 };
 
 /* 000000EC-00000364       .text setCorrectNrmAngle__15daSail_packet_cFsf */
-void daSail_packet_c::setCorrectNrmAngle(short, float) {
-    /* Nonmatching */
+void daSail_packet_c::setCorrectNrmAngle(s16 param_0, f32 param_1) {
+    s16 rnd = 900 + (s32)cM_rndF(200.0f);
+    this->m1C38 += rnd;
+
+    this->m1C34 = 300.0f * JMASSin(this->m1C38);
```
LagoLunatic, line 463:
```suggestion
    this->m1C34 = 300.0f * cM_ssin(this->m1C38);
```
they almost always use the cM version of the sin/cos inlines

src/d/actor/d_a_sail.cpp
```diff
@@ -449,39 +498,445 @@ void daSail_packet_c::setNrmMtx() {
...
+                if (f15 < f16) {
+                    sp34[j] -= 100.0f;
+                } else {
+                    f16 = f15;
+                    sp34[j] = f24;
+                }
+            }
+
+            f32 f14 = 100.0f - std::sqrtf((125.0f * 125.0f) - (f16 * f16));
+            sp6C[j] += f14 > -10.0f ? f14 : -10.0f; // this looks like cLib_minLimit but it does not match
```
LagoLunatic, line 810:
```suggestion
            sp6C[j] += f14 > -10.0f ? f14 : -10.0f; 
```
template functions like cLib_minLimit appear in the debug map for every unit they were used in (unlike regular functions that appear for just the first one). cLib_minLimit doesn't appear for d_a_sail.o, so they didn't use it here, and must have typed this manually

LagoLunatic (commented):
nice, looks pretty good! by the way, if the unit is 100% matching including weak function order, make sure to edit configure.py to mark it as matching for the build system.

LDAsuku:
I've addressed everything in your review. the weak functions are out of order so I've not updated configure.py

PR #680, opened 2024-10-19
First Person Item Procs

SuperDude88:
Matched these couple while looking into HD code for first person items


src/d/actor/d_a_player_main.cpp
```diff
@@ -2086,7 +2087,8 @@ BOOL daPy_lk_c::checkScopeEnd() {
 
 /* 80112100-8011215C       .text setSubjectMode__9daPy_lk_cFv */
 void daPy_lk_c::setSubjectMode() {
-    /* Nonmatching */
+    dComIfGp_setPlayerStatus0(0, daPyStts0_UNK2000_e);
+    mDoAud_seStart(0x8FA);
```
LagoLunatic, line 2091: there's an enum in JAZelAudio_SE.h that can be used for the 6 seStart calls in this PR

src/d/actor/d_a_player_main.cpp
```diff
@@ -2305,13 +2316,43 @@ BOOL daPy_lk_c::procScope() {
...
+
+    return true;
 }
 
 /* 80112B18-80112BEC       .text procSubjectivity__9daPy_lk_cFv */
 BOOL daPy_lk_c::procSubjectivity() {
-    /* Nonmatching */
+    dComIfGp_setAStatus(7);
+
+    if(checkSubjectEnd(0) != 0 || (m3570 != 0 && (m34C9 & 0x40) == 0)) {
```
LagoLunatic, line 2339:
```suggestion
    if(checkSubjectEnd(0) || (m3570 != 0 && (m34C9 & 0x40) == 0)) {
```
boolean

include/d/actor/d_a_player_main.h
```diff
@@ -43,6 +43,10 @@ class daPy_sightPacket_c : public dDlst_base_c {
     void setSightTex(void* sightTex) { mpSightTex = sightTex; }
     void setLockTex(void* lockTex) { mpLockTex = lockTex; }
     void setImage(ResTIMG* image) { mpImg = image; }
+    u8 checkSEFrame() { return field_0x6; }
+    void incFrame() { field_0x6++; }
+    // This inline isn't in the map but I couldn't think of another clean way to do it
+    void resetFrame() { field_0x6 = 0; }
```
LagoLunatic, line 49: most likely some of the other inlines are supposed to be doing multiple things at once. some of the ones above I implemented might be incomplete since I filled them in based on name, I'll take a closer look later.

SuperDude88: Took the guess that a bit more was inside `incFrame()`, there's only 1 other time the value gets incremented (in `setHookshotSight()`) and it seems to follow the pattern in the new implementation

LagoLunatic: for future reference, by checking the exact size in bytes of the inlines in the debug maps, I'm pretty sure these two inlines are supposed to be like this:
```cpp
void onLockFlg() {
    mLockFlag = true;
    mFrame = 0;
}
void incFrame() {
    mFrame++;
    if(mFrame == 26) {
        mFrame = 0;
    }
}
```

PR #682, opened 2024-10-22
d_wood (bushes) matching

themikelester:
Lago, I've been out of this scene for 5 years, but excited to be back and contributing. Please let me know if you see anything egregious in here, or anything I can do to match your style better.


include/d/d_wood.h
```diff
@@ -2,50 +2,80 @@
...
-    /* 0x0A8 */ Mtx field_0x0a8;
+    /* 0x014 */ s32 mAnmIdx;
+    /* 0x018 */ Mtx mModelViewMtx;
+    /* 0x048 */ Mtx mTrunkModelViewMtx;
+    /* 0x078 */ Mtx mShadowModelMtx;
+    /* 0x0A8 */ Mtx mShadowModelViewMtx;
     /* 0x0D8 */ Unit_c* mpNext;
-    /* 0x0DC */ u8 field_0xdc[0x18C - 0xDC];
+
+    dKy_tevstr_c mTevStr;
```
magcius, line 53: Missing `/* 0x0DC */` here.

include/d/d_tree.h
```diff
@@ -9,6 +9,22 @@ class cCcD_Obj;
...
+    extern cXy l_texCoord[];
+    
+    extern const u32 g_dTree_shadowMatDL_SIZE;
+    extern const u32 g_dTree_Oba_kage_32DL_SIZE;
+
+    extern u8 l_matDL[];
+    extern u8 g_dTree_shadowMatDL[];
+    extern u8 g_dTree_Oba_kage_32DL[];
+};
+
```
magcius, line 27: This shouldn't be necessary.

src/d/d_wood.cpp
```diff
@@ -4,184 +4,917 @@
...
+
+  PSMTXIdentity(mModelMtx);
+  PSMTXIdentity(mTrunkModelMtx);
+  mMode = Mode_Max;
+  this->mCountdown = 0;
+  this->mWindDir = 0;
+  this->mWindPow = 0.0;
+  this->mPosOffsetY = 0.0;
+  this->mPosOffsetZ = 0.0;
+  this->mVelY = 0.0;
```
magcius, line 133: Remove `this->`

src/d/d_wood.cpp
```diff
@@ -4,184 +4,917 @@
...
+static s32 _M_init_num = 0;
+
 /* 800BDED0-800BDF5C       .text mode_norm_init__Q25dWood5Anm_cFv */
 void dWood::Anm_c::mode_norm_init() {
-    /* Nonmatching */
+  this->mMode = Mode_Norm;
+
+  for (u32 i = 0; i < 2; i++) {
+    this->mRotY[i] = (short)(_M_init_num << 0xd);
+    this->mRotX[i] = (short)(_M_init_num << 0xd);
```
magcius, line 274: I suspect this is `_M_init_num * 0x2000`

src/d/d_wood.cpp
```diff
@@ -4,184 +4,917 @@
...
+static const Attr_c L_attr = {};
+
+//-----------------------------------------
+// Helpers
+//-----------------------------------------
+
+// This is a 3-step sqrt estimate, but it also contains an extra multiply by the
+// supplied magnitude. I am not 100% sure what it is accomplishing.
+static inline double inv_sqrt(float mag) {
+  if (mag > 0.0f) {
```
magcius, line 104: Is this not `std::sqrtf`?

themikelester: Whoa, it is! I'm surprised to learn that it's implemented that way. Is there any way to have Ghidra spot that that's just a simple std::sqrtf?

include/d/d_tree.h
themikelester: Not necessary (I could use externs), but d_wood does use these symbols from the d_tree namespace (and so does d_tree). I figured it'd be cleaner to keep them in the header. 

Do you think I should just remove them from the header?

src/d/d_wood.cpp
themikelester: Is there a way to have Ghidra not add "this->" in the decompile window?

include/d/d_tree.h
LagoLunatic: d_tree isn't a namespace, it's just the object name. keep the extern declarations in the header but remove the namespace - the `g_` prefix means they're global variables.

LagoLunatic: oh but the `l_` ones are local, so those should be removed from the header, yes.

src/d/d_wood.cpp
LagoLunatic: I doubt it, ghidra is effectively showing C++ as if it was pseudo-C code, and typing `this->` is necessary is C.

src/d/d_wood.cpp
LagoLunatic: no, you have to recognize what it looks like manually.

magcius: Not sure. I use my eyeballs. Maybe RootCubed could some detection magic here.

src/d/d_wood.cpp
```diff
@@ -4,184 +4,899 @@
...
 #include "dolphin/types.h"
+#include "m_Do/m_Do_lib.h"
+#include "m_Do/m_Do_mtx.h"
+
+// These belong in the d_tree compilation unit, in the d_tree namespace. But in
+// order to match, they must be in this compilation unit so they can have a
+// 16-bit offset, allowing us load their pointer using a single load immediate
+// instruction
+u8 d_tree::g_dTree_shadowTexCoord[8 /* TODO: real length */] =
+    {}; // Supposed to be 2 byte UV, but I can't find a matching type
```
LagoLunatic: there doesn't necessarily have to be a struct type for this fyi. it could just be a two-dimensional array of u8s.

src/d/d_wood.cpp
```diff
@@ -4,184 +4,899 @@
...
+//-----------------------------------------
 
 /* 800BD678-800BD710       .text __ct__Q25dWood5Anm_cFv */
 dWood::Anm_c::Anm_c() {
-    /* Nonmatching */
+  int iVar1;
+  int iVar2;
+
+  PSMTXIdentity(mModelMtx);
+  PSMTXIdentity(mTrunkModelMtx);
```
LagoLunatic, line 108:
```suggestion
  MTXIdentity(mModelMtx);
  MTXIdentity(mTrunkModelMtx);
```

src/d/d_wood.cpp
```diff
@@ -4,184 +4,899 @@
...
+  if (mAnimCooldown > 0) {
+    mAnimCooldown -= 1;
+  }
+
+  // Evaluate for attacks that will not cut us down
+  if ((ret & 1)) {
+    cCcD_Obj *atHitObj = inf.GetAtHitObj();
+    if (atHitObj != NULL &&
+        // @TODO: I think the cCcD_ObjAtType enum is reversed. E.g. the first
+        // one should be Machete (bit 21) not Wind
```
LagoLunatic, line 463: the enum definitely isn't reversed. the machete actor is using AT_TYPE_MACHETE and the fan actor is using AT_TYPE_WIND currently. why do you believe it's backwards?

src/d/d_wood.cpp
```diff
@@ -4,184 +4,899 @@
...
+
+      // Clear the hit bit so that we don't get cut down
+      ret &= ~0x01;
+
+      if (actor != NULL && mAnimCooldown == 0) {
+        animIdx = packet->search_anm(Anm_c::Mode_PushInto);
+        mAnimCooldown = 20;
+
+        // Play the cut sound
+        JAIZelBasic::getInterface()->seStart(JA_SE_OBJ_TREE_SWING, &mPos,
```
LagoLunatic: this is the mDoAud_seStart inline, same for the other seStart calls here

include/d/d_tree.h
themikelester: Ohhhh gotcha

src/d/d_wood.cpp
```diff
@@ -4,184 +4,899 @@
...
+          mAnmIdx = animIdx;
+        }
+
+        // If we were able to allocate an animation (or we already have one),
+        // start the "PushInto" (shrinking) animation
+        if ((mAnmIdx >= 8) &&
+            packet->get_anm(mAnmIdx)->get_mode() >=
+                Anm_c::Mode_PushInto) {
+          s16 targetAngle =
+              cLib_targetAngleY(&(actor->current).pos, &mPos);
```
LagoLunatic:
```suggestion
              cLib_targetAngleY(&actor->current.pos, &mPos);
```
parentheses aren't necessary here. same for the other 15 places in this file if you ctrl+f for `).`

src/d/d_wood.cpp
themikelester: Is there a way to have Ghidra pick up on this?

src/d/d_wood.cpp
themikelester: Ah, I was wrong, old comment

src/d/d_wood.cpp
magcius: It's a `#define`, so, no.

LagoLunatic: no, the reason you don't use the `PS` ones are because those are only used in the release build, while the `C_` ones are used in the debug build. the defines with no prefix like `MTXIdentity` are used to swap between them depending on whether building in debug mode or release mode.

src/d/d_wood.cpp
themikelester: Oo gotcha. Where do the mDo* inline functions come from? Are they just patterns that you see in multiple files and decide to make a mDo* function for? 

Is there any easy way to spot them in Ghidra?

LagoLunatic: inline functions are official from the original devs, we don't add any they didn't make (except in *very* rare cases where it's required to match a function).

there's no way for ghidra to notice them because we only have a release build of the game, which has them all inlined away. but you can check the kiosk demo debug maps (pinned in discord) and they list all of the inlines in the game (except ones that don't exist in the demo but do in the final game, but those are also rare).

themikelester: Ohhhhh that's so good to know

LagoLunatic:
some general notes about the style:
* you're using two spaces for indentation, this decomp uses four spaces.
* you should keep the "Nonmatching" comments in until the function is matched. you said the TU is only 73% matched, so I assume several of the functions aren't at 100% yet, but all of the comments were removed.
* it looks like you're using an autoformatter that limits line length, based on the comments containing the mangled symbol names (like `cc_hit_before_cut__Q25dWood6Unit_cFPQ25dWood8Packet_c`) getting split onto multiple lines instead of one. I'd recommend turning this off, we don't usually use an autoformatter and I find it easier if those automatically-generated comments don't get edited.

themikelester:
Thanks! Style replies:
- A mistake. It appears something got tweaked in my vscode settings 
- Ah, I'll definitely keep Nonmatching comments until complete, good idea
- Interesting! The line trimming was bothering me too 😅. I actually thought it was using the project's default settings because there is a default formatter set in settings.json (which I see that you just changed this week). I'll tweak my settings to avoid line trimming

LagoLunatic:
> I actually thought it was using the project's default settings because there is a default formatter set in settings.json (which I see that you just changed this week).

oh, I thought I removed that after copying it from dtk-template but I must have just disabled it locally, my mistake. I removed it now.

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+            }
+        }
+    }
 }
 
 /* 800BEF78-800BEF84       .text __ct__Q25dWood6Room_cFv */
-dWood::Room_c::Room_c() {
-    /* Nonmatching */
-}
+dWood::Room_c::Room_c() { mpUnit = (Unit_c *)0x0; }
```
LagoLunatic, line 679:
```suggestion
dWood::Room_c::Room_c() {
    mpUnit = NULL;
}
```

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+    unit->mpNext = mpUnit;
+    mpUnit = unit;
+    return;
 }
 
 /* 800BEF94-800BEFF0       .text delete_all_unit__Q25dWood6Room_cFv */
 void dWood::Room_c::delete_all_unit() {
-    /* Nonmatching */
+    Unit_c *unit;
+    while (unit = mpUnit, unit != (Unit_c *)0x0) {
```
LagoLunatic, line 692:
```suggestion
    while (unit = mpUnit, unit != NULL) {
```

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+#include "dolphin/gx/GXAttr.h"
+#include "dolphin/gx/GXDisplayList.h"
 #include "dolphin/types.h"
+#include "m_Do/m_Do_lib.h"
+#include "m_Do/m_Do_mtx.h"
+
+// These belong in the d_tree compilation unit. But in order to match, they must
+// be in this compilation unit so they can have a 16-bit offset, allowing us
+// load their pointer using a single load immediate instruction
+// u8 g_dTree_shadowTexCoord[2];
```
LagoLunatic, line 33: outdated comment?

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+
+//-----------------------------------------
+// Types
+//-----------------------------------------
+typedef u8 ALIGN_DECL(32) jData;
```
LagoLunatic, line 38: what does the name "jData" mean? is this official or did you name it?

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+
+    MTXIdentity(mModelMtx);
+    MTXIdentity(mTrunkModelMtx);
+    mMode = Mode_Max;
+    mCountdown = 0;
+    mWindDir = 0;
+    mWindPow = 0.0;
+    mPosOffsetY = 0.0;
+    mPosOffsetZ = 0.0;
+    mVelY = 0.0;
```
LagoLunatic, line 193:
```suggestion
    mWindPow = 0.0f;
    mPosOffsetY = 0.0f;
    mPosOffsetZ = 0.0f;
    mVelY = 0.0f;
```

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+        mPhaseX[iVar1] = 0;
+        mAmpY[iVar1] = 0;
+        mAmpX[iVar1] = 0;
+        iVar1 = iVar1 + 1;
+    }
+
+    mWindDir = targetAngle;
+    mVelY = L_attr.kCutInitVelY;
+    mPosOffsetY = 0.0;
+    mPosOffsetZ = 0.0;
```
LagoLunatic, line 253:
```suggestion
    mPosOffsetY = 0.0f;
    mPosOffsetZ = 0.0f;
```

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+    mNormAnimIdx = AnmID_Norm0;
+    mAlphaScale = 0xff;
+    return;
 }
 
 /* 800BD710-800BD800       .text play__Q25dWood5Anm_cFPQ25dWood8Packet_c */
-void dWood::Anm_c::play(dWood::Packet_c*) {
-    /* Nonmatching */
+void dWood::Anm_c::play(dWood::Packet_c *i_packet) {
+    if (mMode != 6) {
```
LagoLunatic, line 211:
```suggestion
    if (mMode != Mode_Max) {
```

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+    mVelY = 0.0;
+
+    iVar1 = 0;
+    for (u32 i = 0; i < 2; i++) {
+        mPhaseY[iVar1] = 0;
+        mPhaseX[iVar1] = 0;
+        mAmpY[iVar1] = 0;
+        mAmpX[iVar1] = 0;
+        iVar1 = iVar1 + 1;
+    }
```
LagoLunatic, line 202: I notice most of the loops have two separate index variables, is there a reason for this? does it not match if you do:
```suggestion
    for (u32 i = 0; i < 2; i++) {
        mPhaseY[i] = 0;
        mPhaseX[i] = 0;
        mAmpY[i] = 0;
        mAmpX[i] = 0;
    }
```

src/d/d_wood.cpp
```diff
@@ -4,184 +4,970 @@
...
+    mDoMtx_stack_c::now[1][3] = mPos.y;
+    mDoMtx_stack_c::now[2][3] = mPos.z;
+    mDoMtx_concat(j3dSys.getViewMtx(), mDoMtx_stack_c::get(),
+                  mTrunkModelViewMtx);
+
+    mDoMtx_concat(j3dSys.getViewMtx(), mShadowModelMtx, mShadowModelViewMtx);
 }
 
+/* 800BEA28-800BEA50       .text clear__Q25dWood6Unit_cFv */
+void dWood::Unit_c::clear() { cLib_memSet(this, 0, 0x18c); }
```
LagoLunatic, line 520:
```suggestion
void dWood::Unit_c::clear() {
    cLib_memSet(this, 0, sizeof(Unit_c));
}
```

src/d/d_wood.cpp
themikelester: Good catch. That's just copy/paste Ghidra detritus. Fixed!

src/d/d_wood.cpp
themikelester: Ah, that was something I was experimenting with as a way to avoid manualy setting the alignment of several of the large types such as display lists. I named it. 

What do you think? Is it worth keeping around and renaming to something better?

themikelester:
Last bit that isn't matching is that the .data section is too long due to the vtables from collision system objects. I'm unsure as to why this is. See [this post](https://discord.com/channels/727908905392275526/873250400483024976/1300677249300693025)

LagoLunatic:
I think the build is failing because an enum was moved on the main branch - try adding an include for d_cc_d.h and see if it fixes it

LagoLunatic:
looks like there's a version difference - if you don't have the japanese version to check what it is you can use `MatchingFor("GZLE01", "GZLP01")` in configure.py to mark it as matching for only the non-japanese versions of the game for now.

PR #685, opened 2024-12-29
f_op_msg_mng Work

SuperDude88:
Still quite far from finished, just putting this up as draft to reduce chances of duplicate work


LagoLunatic:
I think this should merged now, the TU doesn't need to be finished. Can you fix the conflict and wrap the `dComIfGs_getPalLanguage` calls in VERSION != VERSION_JPN ifdefs?

PR #686, opened 2025-01-08
d_a_obj_rforce match

Azurpourpre:
#437 d_a_obj_rforce

matching 100%

### DIFF
--- a/src/d/actor/d_a_obj_rforce.cpp
+++ b/src/d/actor/d_a_obj_rforce.cpp
@@ -7,36 +7,36 @@
 #include "d/d_bg_s_movebg_actor.h"
 #include "d/d_com_inf_game.h"
 #include "d/d_procname.h"
-#include "d/res/res_vdora.h"
+#include "d/res/res_stptetu.h"
 
 const char daObjRforce::Act_c::M_arcname[] = "StpTetu";
 
 /* 00000078-0000009C       .text solidHeapCB__Q211daObjRforce5Act_cFP10fopAc_ac_c */
 BOOL daObjRforce::Act_c::solidHeapCB(fopAc_ac_c* this_i) {
-    return ((Act_c*)this_i)->create_heap() & 0xFF;
+    return static_cast<Act_c*>(this_i)->create_heap();
 }
 
 /* 0000009C-00000220       .text create_heap__Q211daObjRforce5Act_cFv */
-BOOL daObjRforce::Act_c::create_heap() {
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_obj_rforce.cpp
```diff
@@ -4,68 +4,128 @@
...
+    dComIfGd_setListBG();
+    
+    g_env_light.setLightTevColorType(mpModel, &tevStr);
+    mDoExt_modelUpdateDL(mpModel);
+    dComIfGd_setList();
+
+    return TRUE;
 }
 
+daObjRforce::Act_c::~Act_c(){}
```
LagoLunatic, line 102: You can usually put empty destructors in the header instead of the cpp (i.e. `virtual ~Act_c() {}` in the class body)

src/d/actor/d_a_obj_rforce.cpp
```diff
@@ -4,68 +4,128 @@
...
+
 namespace daObjRforce {
 namespace {
 /* 00000508-00000528       .text Mthd_Create__Q211daObjRforce28@unnamed@d_a_obj_rforce_cpp@FPv */
-void Mthd_Create(void*) {
-    /* Nonmatching */
+void Mthd_Create(void* obj) {
+    ((daObjRforce::Act_c*)obj)->_create();
```
LagoLunatic, line 108:
```suggestion
s32 Mthd_Create(void* obj) {
    return ((daObjRforce::Act_c*)obj)->_create();
```

src/d/actor/d_a_obj_rforce.cpp
```diff
@@ -4,68 +4,128 @@
...
 }
 
 /* 00000528-0000054C       .text Mthd_Delete__Q211daObjRforce28@unnamed@d_a_obj_rforce_cpp@FPv */
-void Mthd_Delete(void*) {
-    /* Nonmatching */
+int Mthd_Delete(void* obj) {
+    return ((daObjRforce::Act_c*)obj)->_delete() & 0xFF;
```
LagoLunatic, line 113:
```suggestion
BOOL Mthd_Delete(void* obj) {
    return ((daObjRforce::Act_c*)obj)->_delete();
```
The thing I said above about create_heap returning `bool` also applies to these other functions like `_delete`

LagoLunatic (commented):
Nice. Also, once a TU matches 100% you can mark it as `Matching` in configure.py.

PR #688, opened 2025-01-31
d_a_obj_hha 100%

Azurpourpre:
100% matching object except the rel file is not matching probably because of weak function ordering

### DIFF
--- a/src/d/actor/d_a_obj_hha.cpp
+++ b/src/d/actor/d_a_obj_hha.cpp
@@ -79,8 +79,18 @@ void daObjHhaPart_c::draw_normal(daObjHha_c*) {
 }
 
 /* 0000056C-00000698       .text create_s__16daObjHhaSplash_cFUsP4cXyzffP5csXyz */
-void daObjHhaSplash_c::create_s(unsigned short, cXyz*, float, float, csXyz*) {
-    /* Nonmatching */
+void daObjHhaSplash_c::create_s(unsigned short param1, cXyz param2, float param3, float param4, csXyz param5) {
+    cXyz acStack38;
+
+    this->m20.set(param2.x, param2.y + param3, param2.z);
+    mDoMtx_stack_c::YrotS(param5.y);
+    mDoMtx_stack_c::multVec(&cXyz::BaseZ, &acStack38);
+    acStack38 *= param4;
+    this->m20 += acStack38;
+    this->m14 = this->m20;
+    this->m2C = param5;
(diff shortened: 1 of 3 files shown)


LagoLunatic:
> daObjHha_c gets dCcD_Stts vtable

What does this mean? Screenshot?

src/d/actor/d_a_obj_hha.cpp
```diff
@@ -4,182 +4,690 @@
...
+    mYgush.draw();
+    return TRUE;
 }
 
 namespace {
 /* 000029F4-00002A14       .text Mthd_Create__25@unnamed@d_a_obj_hha_cpp@FPv */
-void Mthd_Create(void*) {
-    /* Nonmatching */
+void Mthd_Create(void* this_i) {
+    static_cast<daObjHha_c*>(this_i)->_create();
```
LagoLunatic, line 670: Missing return

Azurpourpre:
> > daObjHha_c gets dCcD_Stts vtable
> 
> What does this mean? Screenshot?

image

The "issue" i was mentioning was that the memory layout are not the same (in objdiff), because as you can see, i have dCcD_GStts vtable which is shifting every other vtable after it, and i thought that this was causing some differences between original and decomp, when calling some functions via vtable.

However, I can't find any differences in the assembly. Also, the rel from decomp and original rel files have the same .data section size, so i really doubt now that this is the issue.

LagoLunatic:
> The "issue" i was mentioning was that the memory layout are not the same (in objdiff), because as you can see, i have dCcD_GStts vtable which is shifting every other vtable after it, and i thought that this was causing some differences between original and decomp, when calling some functions via vtable.
> 
> However, I can't find any differences in the assembly. Also, the rel from decomp and original rel files have the same .data section size, so i really doubt now that this is the issue.

vtables being present on the right but not the left isn't an issue. The linker will strip them out, objdiff is just showing the object on the right before it gets linked. If you check the original symbol map from the vanilla game, you can see that those vtables are present, just marked "UNUSED" by the linker:
```
  UNUSED   00000c ........ __vt__10cCcD_GStts d_a_obj_hha.o 
  UNUSED   00000c ........ __vt__10dCcD_GStts d_a_obj_hha.o 
```

The real reason it doesn't match is probably weak function ordering. Several functions in this TU return a `cXyz` directly, which for some bizarre reason we haven't figured out causes this compiler to put them in the wrong .text section for us compared to the vanilla game. You can revert the object to `NonMatching` in configure.py and it should be fine to merge.

PR #689, opened 2025-02-19
d_a_tori_flag 100%

mattias-blum:
- Had to change the return type for the constructor of `dCloth_packet_c`
- May want to reconsider the naming of certain members of `daTori_Flag_HIO_c` now or in the future

### DIFF
--- a/include/d/actor/d_a_tori_flag.h
+++ b/include/d/actor/d_a_tori_flag.h
@@ -6,6 +6,22 @@
 #include "SSystem/SComponent/c_phase.h"
 #include "d/d_cloth_packet.h"
 
+class daTori_Flag_HIO_c {
+    public:
+        // Constructor
+        daTori_Flag_HIO_c();
+
+        // Virtual Destructor
+        virtual ~daTori_Flag_HIO_c(){}
+        
+    public:
+        /* Place member variables here */
+        /* 0x00 */ //vtable
+        /* 0x04 */ signed char flag;
+        /* 0x08 */ float someFloat;
+        /* 0x0c */ signed short someShort;
+};
+
 class daTori_Flag_c : public fopAc_ac_c {
 public:
     inline s32 _create();
@@ -14,8 +30,10 @@ public:
     inline bool _execute();
 
     void set_mtx();
-    void CreateHeap();
-    void CreateInit();
+    BOOL CreateHeap();
+    s32 CreateInit();
+
+    static const char M_arcname[7];
 
 public:
     /* 0x290 */ request_of_phase_process_class mPhsTrflag;
@@ -29,12 +47,7 @@ public:
     /* 0x4D0 */ dCcD_Cyl mCyl2;
 };
 
-class daTori_Flag_HIO_c {
-public:
-    daTori_Flag_HIO_c();
-
-public:
-    /* Place member variables here */
-};
+extern daTori_Flag_HIO_c l_HIO;
+extern cXyz l_flag_offset;
 
 #endif /* D_A_TORI_FLAG_H */
--- a/include/d/actor/d_a_tori_flag.h
+++ b/include/d/actor/d_a_tori_flag.h
@@ -15,9 +15,8 @@ class daTori_Flag_HIO_c {
         virtual ~daTori_Flag_HIO_c(){}
         
     public:
-        /* Place member variables here */
         /* 0x00 */ //vtable
-        /* 0x04 */ signed char flag;
+        /* 0x04 */ signed char mNo;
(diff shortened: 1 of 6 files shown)


include/d/actor/d_a_tori_flag.h
```diff
@@ -6,6 +6,22 @@
...
+class daTori_Flag_HIO_c {
+    public:
+        // Constructor
+        daTori_Flag_HIO_c();
+
+        // Virtual Destructor
+        virtual ~daTori_Flag_HIO_c(){}
+        
+    public:
+        /* Place member variables here */
```
LagoLunatic, line 18: You can remove this comment since once the member variables are added

include/d/actor/d_a_tori_flag.h
```diff
@@ -29,12 +47,7 @@ class daTori_Flag_c : public fopAc_ac_c {
...
 
-class daTori_Flag_HIO_c {
-public:
-    daTori_Flag_HIO_c();
-
-public:
-    /* Place member variables here */
-};
+extern daTori_Flag_HIO_c l_HIO;
+extern cXyz l_flag_offset;
```
LagoLunatic, line 51: The `l_` prefix means local, so these don't need to go in the header since they're local to this one TU and won't be included for use in other TUs

src/d/actor/d_a_tori_flag.cpp
```diff
@@ -4,66 +4,201 @@
...
+        /* Center */ 0.0f, 0.0f, 0.0f,
+        /* Radius */ 30.0f,
+        /* Height */ 400.0f,
+    },
+};
 
 /* 000000EC-00000118       .text __ct__17daTori_Flag_HIO_cFv */
 daTori_Flag_HIO_c::daTori_Flag_HIO_c() {
-    /* Nonmatching */
+    flag = -0x1;
```
LagoLunatic, line 53:
```suggestion
    mNo = -0x1;
```
The first field of HIO classes (this s8 that defaults to -1) always seems to be the HIO child index - which TP debug strings indicate was named something like `mNo`, so I like to name them the same thing in TWW (even though it's not really used outside of debug builds)

src/d/actor/d_a_tori_flag.cpp
```diff
@@ -4,66 +4,201 @@
...
+    someFloat = 0.0;
```
LagoLunatic, line 54:
```suggestion
    someFloat = 0.0f;
```

src/d/actor/d_a_tori_flag.cpp
```diff
@@ -4,66 +4,201 @@
...
+    mStts.Init(0xFF,0xFF,this);
+    mCyl.Set(l_cyl_src);
+    mCyl.SetStts(&mStts);
+    cXyz* wind_vec = dKyw_get_wind_vec();
+    mWindvec.x = wind_vec->x;
+    mWindvec.y = wind_vec->y;
+    mWindvec.z = wind_vec->z;
+    set_mtx();
+    dKy_tevstr_init(&mClothTevStr, current.roomNo, 0xFF);
+    cullMtx = mpModel->mBaseTransformMtx;
```
LagoLunatic, line 110:
```suggestion
    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
```
inline

src/d/actor/d_a_tori_flag.cpp
```diff
@@ -4,66 +4,201 @@
...
+    if (windStrength > currStrength) {
+        mWindvec.x = wind.x;
+        mWindvec.y = wind.y;
+        mWindvec.z = wind.z;
+    }
+    else {
+        cLib_addCalcPos2(&mWindvec, wind, 0.05f, 0.05f);
+    }
+    mCyl.SetC(current.pos);
+    g_dComIfG_gameInfo.play.mCcS.Set(&mCyl);
```
LagoLunatic, line 172:
```suggestion
    dComIfG_Ccsp()->Set(&mCyl);
```

mattias-blum:
Made the suggested changes.

LagoLunatic:
> Made the suggested changes.

It looks like the inlines I suggested still aren't used, did the TU no longer match with them in use? If so you can leave it as is but put a comment above those lines mentioning what inlines are supposed to be used there but break the codegen.

Otherwise everything looks good!

mattias-blum:
Sorry forgot about that one, I'll change it real quick

mattias-blum:
Okay should be good to go now

mattias-blum:
Okay should be good now

PR #690, opened 2025-02-20
d_a_steam_tag 100%

mattias-blum:
- Added inline functions to `d_point_wind` that were referenced in this TU
- This line `mCreateTimer = 0x1e;` in `daSteamTag_c::execute()` seems like it's referencing some constant declared in another file , not sure what it would be but it's likely related to timers somehow.
- I think the constant referenced in this line is correct (the value of it is) `checkItemGet(dItem_PEARL_DIN_e, TRUE)` in `daSteamTag_c::create()` but since I'm not sure what this actor is I don't know if Din's Pearl makes sense as the item here.

### DIFF
--- a/include/d/actor/d_a_steam_tag.h
+++ b/include/d/actor/d_a_steam_tag.h
@@ -3,29 +3,47 @@
 
 #include "f_op/f_op_actor.h"
 #include "d/d_cc_d.h"
+#include "d/d_particle.h"
 #include "d/d_point_wind.h"
+#include "f_op/f_op_actor_mng.h"
 
 class JPABaseEmitter;
 
+struct daSteamTag_mData {
+    /* 0x00 */ s16 field_0x00;
+    /* 0x02 */ s16 field_0x02;
+    /* 0x04 */ s16 mCreateTimer_seed;
+    /* 0x06 */ s16 mCreateTimer_offset;
+    /* 0x08 */ s16 mEmitTimer_offset;
+    /* 0x0A */ s16 mEmitTimer_seed;
+    /* 0x0C */ u8  mSteamAlpha;
+    /* 0x0D */ u8  field_0x0D;
+    /* 0x0E */ u8  field_0x0E;
+    /* 0x0F */ u8  field_0x0F;
+};
+
 class daSteamTag_c : public fopAc_ac_c {
 public:
+    ~daSteamTag_c();
     inline s32 create();
     inline BOOL draw();
     inline BOOL execute();
 
-    void getData();
-    void CreateInit();
-    void createEmitter();
-    void endEmitter();
+    const daSteamTag_mData* getData();
+    s32 CreateInit();
+    BOOL createEmitter();
+    bool endEmitter();
 
     static void init();
+
+    static daSteamTag_mData mData;
 
 public:
     /* 0x290 */ JPABaseEmitter* mpEmitter;
     /* 0x294 */ u8 m294[0x296 - 0x294];
     /* 0x296 */ s16 mEmitTimer;
     /* 0x298 */ s16 mCreateTimer;
-    /* 0x29A */ u8 m29A[0x29B - 0x29A];
+    /* 0x29A */ u8 m29A;
     /* 0x29B */ u8 m29B;
     /* 0x29C */ dCcD_Stts mGStts;
     /* 0x2D8 */ dCcD_Cyl mCyl;
@@ -35,6 +53,10 @@ public:
 
 public:
     static u8 mEmitterNum;
+};
+
+namespace daSteamTag_prm {
+    inline u8 getSchBit(daSteamTag_c* item) { return (fopAcM_GetParam(item) >> 2) & 0xFF; }
 };
 
 #endif /* D_A_STEAM_TAG_H */
--- a/include/d/actor/d_a_steam_tag.h
+++ b/include/d/actor/d_a_steam_tag.h
@@ -12,14 +12,13 @@ class JPABaseEmitter;
 struct daSteamTag_mData {
     /* 0x00 */ s16 field_0x00;
     /* 0x02 */ s16 field_0x02;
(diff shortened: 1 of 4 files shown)


LagoLunatic:
> This line mCreateTimer = 0x1e; in daSteamTag_c::execute() seems like it's referencing some constant declared in another file , not sure what it would be but it's likely related to timers somehow.

Is there something specific that leads you to believe it's defined in another file? I would assume it's just a one second timer constant (0x1e = decimal 30, Wind Waker runs at 30 frames per second, and the timer counts down once per frame), which are pretty common to see in game code.

> I think the constant referenced in this line is correct (the value of it is) checkItemGet(dItem_PEARL_DIN_e, TRUE) in daSteamTag_c::create() but since I'm not sure what this actor is I don't know if Din's Pearl makes sense as the item here.

It looks like they added a hardcoded check to make the steam disappear from DRC's pre-dungeon room when you beat the dungeon by checking if the player owns Din's Pearl (since that's the reward for completing DRC).

include/d/actor/d_a_steam_tag.h
```diff
@@ -3,29 +3,47 @@
...
 class JPABaseEmitter;
 
+struct daSteamTag_mData {
+    /* 0x00 */ s16 field_0x00;
+    /* 0x02 */ s16 field_0x02;
+    /* 0x04 */ s16 mCreateTimer_seed;
+    /* 0x06 */ s16 mCreateTimer_offset;
+    /* 0x08 */ s16 mEmitTimer_offset;
+    /* 0x0A */ s16 mEmitTimer_seed;
+    /* 0x0C */ u8  mSteamAlpha;
```
LagoLunatic, line 19: The word "seed" isn't right here, `cM_rndF`'s argument affects the maximum range of values it can return, not a random seed. e.g. `cM_rndF(5.0f)` -> anywhere in the range of 0.0f to 5.0f randomly. (I don't know why we have that parameter simply called "f" in c_math.cpp, but in c_math.h it's more clearly called "max".)

(Also, this isn't a hard rule, but I think the devs tended to generally use `mThisNamingStyle` for classes and `this_naming_style` for structs?)

So I would name these more along these lines:
```suggestion
    /* 0x04 */ s16 create_time_range;
    /* 0x06 */ s16 create_time_min;
    /* 0x08 */ s16 emit_time_min;
    /* 0x0A */ s16 emit_time_range;
    /* 0x0C */ u8  steam_alpha;
```

src/d/actor/d_a_steam_tag.cpp
```diff
@@ -4,56 +4,209 @@
...
+        /* SrcGObjTg Spl     */ 0,
+        /* SrcGObjTg Mtrl    */ 0,
+        /* SrcGObjTg SPrm    */ dCcG_TgSPrm_NoHitMark_e,
+        /* SrcGObjCo SPrm    */ 0,
+    },
+    // cM3dGCpsS
+    {
+        /* P0 */ 0.0f, 0.0f, 0.0f,
+        /* P1 */ 0.0f, 0.0f, 0.0f,
+        /* Height */ 200.0f,
```
LagoLunatic, line 42: The official names for these fields are "Start" "End" and "Radius" from the maps (I think P0/P1/Height were what we had in ghidra)

src/d/actor/d_a_steam_tag.cpp
```diff
@@ -4,56 +4,209 @@
...
+      fVar2 = 50.0f;
+    }
+    mDoMtx_stack_c::transM(0.0f, fVar2, 0.0f);
+    mDoMtx_stack_c::multVec(&cXyz::Zero, &mPntWindCpsS.mEnd);
+    mPntWindCpsS.mRadius = 50.0;
+    mCps.cM3dGCps::Set(mPntWindCpsS);
+    mCps.CalcAtVec();
+    mPointWind.set_pwind_init(&mPntWindCpsS);
+    fopAcM_offDraw(this);
+    return 1;
```
LagoLunatic, line 80:
```suggestion
    return TRUE;
```

CreateInit usually returns `BOOL` (don't confuse with Create, which returns a phase state)

src/d/actor/d_a_steam_tag.cpp
```diff
@@ -4,56 +4,209 @@
...
+    if (mpEmitter != NULL) {
+      mpEmitter->becomeInvalidEmitter();
+      mpEmitter = NULL;
+    }
+    mPointWind.set_pwind_delete();
+    return;
+}
+
+s32 daSteamTag_c::create() {
+    int result;
```
LagoLunatic, line 194:
```suggestion
    int phase_state;
```
We know the exact name they liked to use for this local variable because it appears in debug asserts for dozens of actors

src/d/actor/d_a_steam_tag.cpp
```diff
@@ -4,56 +4,209 @@
...
+    },
+    // cM3dGCpsS
+    {
+        /* P0 */ 0.0f, 0.0f, 0.0f,
+        /* P1 */ 0.0f, 0.0f, 0.0f,
+        /* Height */ 200.0f,
+    },
+};
+
+daSteamTag_mData daSteamTag_c::mData = {15861, 49807, 300, 100, 50, 150, 0xB4, 0x00, 0x00};
```
LagoLunatic, line 46: For struct initializers I usually like to put each field on a separate line and add a comment with the name of the field (like `l_cps_src` has above) because it makes it more clear to someone reading the code where certain values are defined (such as these timer ranges here)

mattias-blum:
>Is there something specific that leads you to believe it's defined in another file? I would assume it's just a one second timer constant (0x1e = decimal 30, Wind Waker runs at 30 frames per second, and the timer counts down once per frame), which are pretty common to see in game code.

Good point I didn't think about that, usually when I see some hex constant in the code somewhere I assume it's defined elsewhere but here that's clearly not the case.

mattias-blum:
Requested PR changes have been made

PR #691, opened 2025-02-22
d_a_shutter 100%

mattias-blum:
Wasn't sure what to name some of the used member variables that have yet to be assigned names (ex. `field_0x339`) but if you have an intuitive guess based on the code/previous actors I'd be happy to change them.

### DIFF
--- a/include/d/actor/d_a_shutter.h
+++ b/include/d/actor/d_a_shutter.h
@@ -41,23 +41,23 @@ public:
     /* 0x2A8 */ Mtx mMtx[2];
     /* 0x308 */ cXyz mcXyz[2];
     /* 0x320 */ s32 field_0x320;
-    /* 0x324 */ s32 mNo;
+    /* 0x324 */ s32 mSwitchNo;
     /* 0x328 */ s32 field_0x328;
     /* 0x32C */ s32 mStaffId;
     /* 0x330 */ s16 mOpenEventIdx;
     /* 0x332 */ s16 mCloseEventIdx;
     /* 0x334 */ u8 field_0x334;
     /* 0x335 */ u8 mType;
-    /* 0x336 */ s8 mSwitch;
-    /* 0x337 */ s8 mCount;
-    /* 0x338 */ s8 mTimer;
+    /* 0x336 */ s8 mbIsSwitch;
+    /* 0x337 */ s8 mFrameTimer;
+    /* 0x338 */ u8 mTimer;
     /* 0x339 */ u8 field_0x339;
     /* 0x33A */ u8 field_0x33A;
     /* 0x33B */ u8 field_0x33B;
 };
 
 namespace daShutter_prm {
-    inline u8 getmNo(daShutter_c* item) { return (fopAcM_GetParam(item) >> 0) & 0xFF; }
+    inline u8 getSwitchNo(daShutter_c* item) { return (fopAcM_GetParam(item) >> 0) & 0xFF; }
     inline u8 getType(daShutter_c* item) { return (fopAcM_GetParam(item) >> 8) & 0x0F; }
 }
 
--- a/include/d/actor/d_a_shutter.h
+++ b/include/d/actor/d_a_shutter.h
@@ -48,7 +48,7 @@ public:
     /* 0x332 */ s16 mCloseEventIdx;
     /* 0x334 */ u8 field_0x334;
     /* 0x335 */ u8 mType;
-    /* 0x336 */ s8 mbIsSwitch;
+    /* 0x336 */ u8 mbIsSwitch;
     /* 0x337 */ s8 mFrameTimer;
     /* 0x338 */ u8 mTimer;
     /* 0x339 */ u8 field_0x339;
--- a/include/d/actor/d_a_shutter.h
+++ b/include/d/actor/d_a_shutter.h
@@ -2,12 +2,14 @@
 #define D_A_SHUTTER_H
 
 #include "f_op/f_op_actor.h"
-#include "SSystem/SComponent/c_phase.h"
+#include "f_op/f_op_actor_mng.h"
+
+class dBgW;
 
 class daShutter_c : public fopAc_ac_c {
 public:
     bool _delete();
-    void CreateHeap();
+    BOOL CreateHeap();
     s32 Create();
     s32 _create();
     void set_mtx();
@@ -16,13 +18,47 @@ public:
     void demo();
     bool _draw();
 
+    /* 0x00 */ static const float m_max_speed[2];
+    /* 0x08 */ static const float m_min_speed[2];
+    /* 0x10 */ static const float m_move_len[2];
+    /* 0x18 */ static const float m_width[2];
+    /* 0x20 */ static const float m_height[2];
+    /* 0x28 */ static const s16 m_bdlidx[2];
+    /* 0x2C */ static const s16 m_dzbidx[2];
+    /* 0x30 */ static const s32 m_heapsize[2];
+    /* 0x38 */ static const Vec m_cull_min[2];
+    /* 0x50 */ static const Vec m_cull_max[2];
+
+    static char* m_arcname[2];
+    static char* m_open_ev_name[2];
+    static char* m_close_ev_name[2];
+    static char* m_staff_name[2];
+    
 public:
     /* 0x290 */ request_of_phase_process_class mPhs;
-    /* 0x298 */ u8 m298[0x308 - 0x298];
-    /* 0x308 */ cXyz m308[2];
-    /* 0x320 */ u8 m320[0x335 - 0x320];
+    /* 0x298 */ J3DModel* mpModel[2];
+    /* 0x2A0 */ dBgW* mdBgW[2];
+    /* 0x2A8 */ Mtx mMtx[2];
+    /* 0x308 */ cXyz mcXyz[2];
+    /* 0x320 */ s32 field_0x320;
+    /* 0x324 */ s32 mNo;
+    /* 0x328 */ s32 field_0x328;
+    /* 0x32C */ s32 mStaffId;
+    /* 0x330 */ s16 mOpenEventIdx;
+    /* 0x332 */ s16 mCloseEventIdx;
+    /* 0x334 */ u8 field_0x334;
     /* 0x335 */ u8 mType;
-    /* 0x336 */ u8 m336[0x33C - 0x336];
+    /* 0x336 */ s8 mSwitch;
+    /* 0x337 */ s8 mCount;
+    /* 0x338 */ s8 mTimer;
+    /* 0x339 */ u8 field_0x339;
+    /* 0x33A */ u8 field_0x33A;
+    /* 0x33B */ u8 field_0x33B;
 };
+
+namespace daShutter_prm {
+    inline u8 getmNo(daShutter_c* item) { return (fopAcM_GetParam(item) >> 0) & 0xFF; }
+    inline u8 getType(daShutter_c* item) { return (fopAcM_GetParam(item) >> 8) & 0x0F; }
+}
 
 #endif /* D_A_SHUTTER_H */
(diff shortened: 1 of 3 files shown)


include/d/actor/d_a_shutter.h
```diff
@@ -16,13 +18,47 @@ class daShutter_c : public fopAc_ac_c {
...
+    /* 0x336 */ s8 mSwitch;
+    /* 0x337 */ s8 mCount;
+    /* 0x338 */ s8 mTimer;
+    /* 0x339 */ u8 field_0x339;
+    /* 0x33A */ u8 field_0x33A;
+    /* 0x33B */ u8 field_0x33B;
 };
 
+namespace daShutter_prm {
+    inline u8 getmNo(daShutter_c* item) { return (fopAcM_GetParam(item) >> 0) & 0xFF; }
```
LagoLunatic, line 60:
```suggestion
    inline u8 getSwitchNo(daShutter_c* item) { return (fopAcM_GetParam(item) >> 0) & 0xFF; }
```
Official name from the debug map

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
 }
 
 /* 000002B8-000004B4       .text Create__11daShutter_cFv */
 s32 daShutter_c::Create() {
-    /* Nonmatching */
+    fopAcM_SetMtx(this, mpModel[0]->getBaseTRMtx());
+    Vec cullMin = m_cull_min[mType];
+    Vec cullMax = m_cull_max[mType];
+    fopAcM_setCullSizeBox(this, cullMin.x, cullMin.y, cullMin.z, cullMax.x, cullMax.y, cullMax.z);
+    mNo = daShutter_prm::getmNo(this);
```
LagoLunatic, line 71:
```suggestion
    mSwitchNo = daShutter_prm::getSwitchNo(this);
```
Member name of params should generally also match the inline name if it makes sense

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
+    int iVar3;
+    if (mCount >= 0) {
+        mCount--;
+    }
+    demo();
+    set_mtx();
+    for (int i = 0; i < (int)ARRAY_SIZE(mMtx); i++) {
+        mdBgW[i]->Move();
+    }
+    mSwitch = fopAcM_isSwitch(this, mNo);
```
LagoLunatic, line 139:
```suggestion
    mbIsSwitch = fopAcM_isSwitch(this, mSwitchNo);
```
isSwitch doesn't return a switch, it takes a switch and returns a boolean for whether the switch is currently set or not.

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
+    }
+    dComIfGp_evmng_cutEnd(mStaffId);
 }
 
 /* 00000B14-00000CF0       .text demo__11daShutter_cFv */
 void daShutter_c::demo() {
-    /* Nonmatching */
+    u8 bVar3 = fopAcM_isSwitch(this, mNo);
+    if (field_0x320 == 0) {
+        if ((bVar3 != (u8)mSwitch) && (mCount < 0)) {
```
LagoLunatic, line 222:
```suggestion
    bool isSwitch = fopAcM_isSwitch(this, mSwitchNo);
    if (field_0x320 == 0) {
        if ((isSwitch != mbIsSwitch) && (mCount < 0)) {
```
Does this still match if you change mbIsSwitch from s8 -> bool and the local variable from u8 -> bool as well? There are some cases where you might need to use `u8` for a boolean to get it to match, but most of the time `bool` or `BOOL` would be correct.

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
+        case 0: //WAIT
+        {
+            mTimer = 0xf;
+            field_0x33A = 0;
+            dComIfGp_evmng_cutEnd(mStaffId);
+            return;
+        }
+        case 1: //WAIT02
+        {
+            if (!cLib_calcTimer((u8*)&mTimer)) {
```
LagoLunatic, line 168:
```suggestion
            if (!cLib_calcTimer(&mTimer)) {
```
This indicates `mTimer`'s type is supposed to be `u8`, not `s8`.

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
+    mSwitch = fopAcM_isSwitch(this, mNo);
+    return TRUE;
 }
 
+const char* M_act_table[4] = {"WAIT", "WAIT02", "OPEN", "CLOSE"};
+
 /* 00000788-00000B14       .text shutter_move__11daShutter_cFv */
 void daShutter_c::shutter_move() {
-    /* Nonmatching */
+    int actionIndex = dComIfGp_evmng_getMyActIdx(mStaffId, (char**)M_act_table, ARRAY_SIZE(M_act_table), FALSE, 0);
```
LagoLunatic, line 147:
```cpp
/* 00000788-00000B14       .text shutter_move__11daShutter_cFv */
void daShutter_c::shutter_move() {
    static char* action_table[] = {"WAIT", "WAIT02", "OPEN", "CLOSE"};
    int actionIndex = dComIfGp_evmng_getMyActIdx(mStaffId, action_table, ARRAY_SIZE(action_table), FALSE, 0);
```

In the debug map, this array of action names is called `action_table$4442`, which other than telling us that the name is `action_table`, the `$4442` part at the end of a variable name also tells us that it's a local static (it goes inside a function, but is marked `static` so ).
(Also, the const can be removed to get rid of the `(char**)` cast.)

This doesn't affect the match in this case but it's good to be aware of because iirc there are weirder cases involving inlines where you need to use local statics a certain way to get it to match.

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
 
 /* 00000CF0-00000DD8       .text _draw__11daShutter_cFv */
 bool daShutter_c::_draw() {
-    /* Nonmatching */
+    cXyz actorPos;
+    for (int i = 0; i < (int)ARRAY_SIZE(mMtx); i++) {
+        actorPos.x = current.pos.x;
+        actorPos.y = current.pos.y;
+        actorPos.z = current.pos.z;
+        PSVECAdd(&actorPos, &mcXyz[i], &actorPos);
```
LagoLunatic, line 261:
```suggestion
        actorPos += mcXyz[i];
```
Looks like the += inline.
Operator inlines have weird names in the debug map, you can throw them into objdiff -> Tools -> Demangle to see what they really are

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
 }
 
 /* 00000CF0-00000DD8       .text _draw__11daShutter_cFv */
 bool daShutter_c::_draw() {
-    /* Nonmatching */
+    cXyz actorPos;
+    for (int i = 0; i < (int)ARRAY_SIZE(mMtx); i++) {
+        actorPos.x = current.pos.x;
+        actorPos.y = current.pos.y;
+        actorPos.z = current.pos.z;
```
LagoLunatic, line 260:
```suggestion
        actorPos = current.pos;
```
Might be the = inline? Ignore if you already tried that and it didn't match

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
 }
 
 /* 000006F0-00000788       .text _execute__11daShutter_cFv */
 bool daShutter_c::_execute() {
-    /* Nonmatching */
+    bool bVar1;
+    int iVar3;
+    if (mCount >= 0) {
+        mCount--;
+    }
```
LagoLunatic, line 133: This looks like another timer counting down, not a count.

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,292 @@
...
+        mStaffId = dComIfGp_evmng_getMyStaffId(m_staff_name[mType], NULL, 0);
+        shutter_move();
+    }
+    else if ((field_0x320 == 1) && (mOpenEventIdx != 0)) {
+        fopAcM_orderOtherEventId(this, mOpenEventIdx, 0xff, 0xffff, 0, 1);
+        eventInfo.mCondition = eventInfo.mCondition | 2;
+    }
+    else if ((field_0x320 == 2) && (mCloseEventIdx != 0)) {
+        fopAcM_orderOtherEventId(this, mCloseEventIdx, 0xff, 0xffff, 0, 1);
+        eventInfo.mCondition = eventInfo.mCondition | 2;
```
LagoLunatic, line 250: There's an inline for this, and an enum for these flags (2) here

src/d/actor/d_a_shutter.cpp
mattias-blum: No, I couldn't get it to match when setting both to either `bool` or `BOOL`

LagoLunatic: Interesting, does it match with mbIsSwitch changed from s8 -> u8? (Instead of casting it to u8)

mattias-blum:
Made the PR changes.

src/d/actor/d_a_shutter.cpp
mattias-blum: I named it `mFrameTimer` in the latest commit since it's initialized to 30 and - assuming `_execute()` is called every frame - counts down every frame.

src/d/actor/d_a_shutter.cpp
mattias-blum: This is good to know since this was an issue with `d_a_shutter2` not matching.

src/d/actor/d_a_shutter.cpp
mattias-blum: Yep. Just committed that change

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,284 @@
...
 #include "d/d_procname.h"
+#include "d/d_com_inf_game.h"
+
+const float daShutter_c::m_max_speed[2] = {3.0f, 3.0f};
+const float daShutter_c::m_min_speed[2] = {1.0f, 1.0f};
+const float daShutter_c::m_move_len[2] = {200.0f, 84.0f};
+const float daShutter_c::m_width[2] = {220.0f, 82.0f};
+const float daShutter_c::m_height[2] = {200.0f, 100.0f};
+const s16 daShutter_c::m_bdlidx[2] = {4, 4};
+const s16 daShutter_c::m_dzbidx[2] = {7, 7};
```
LagoLunatic, line 16: These are file indexes used to load files, you can find enums for them in the headers `res_htobi1.h` (for [0]) and `res_htobi2.h` (for [1])

src/d/actor/d_a_shutter.cpp
```diff
@@ -5,80 +5,284 @@
...
+const s32 daShutter_c::m_heapsize[2] = {0x1140, 0x3000};
+const Vec daShutter_c::m_cull_min[2] = {{-500.0f, -100.0f, -50.0f}, {-150.0f, -100.0f, -50.0f}};
+const Vec daShutter_c::m_cull_max[2] = {{500.0f, 250.0f, 50.0f}, {150.0f, 250.0f, 50.0f}};
+
+char* daShutter_c::m_arcname[2] = {"Htobi1", "Htobi2"};
+char* daShutter_c::m_open_ev_name[2] = {"HYSDOOROPEN", "R03DOOROPEN"};
+char* daShutter_c::m_close_ev_name[2] = {0, "R03DOORCLOSE"};
```
LagoLunatic, line 23:
```suggestion
char* daShutter_c::m_close_ev_name[2] = {NULL, "R03DOORCLOSE"};
```

LagoLunatic (commented):
I missed these the first time

mattias-blum:
Changes have been made

PR #692, opened 2025-02-23
d_a_shutter2 100%

mattias-blum:
- Code should be consistent with `d_a_shutter`
- A couple of variables still remain unnamed

### DIFF
--- a/include/d/actor/d_a_shutter2.h
+++ b/include/d/actor/d_a_shutter2.h
@@ -35,7 +35,7 @@ public:
     /* 0x2C8 */ request_of_phase_process_class mPhs;
     /* 0x2D0 */ J3DModel* mpModel;
     /* 0x2D4 */ Mtx mMtx;
-    /* 0x304 */ s32 m304;
+    /* 0x304 */ s32 mActionIndex;
     /* 0x308 */ s32 mSwitchNo;
     /* 0x30C */ s32 m30C;
     /* 0x310 */ s32 mStaffId;
@@ -49,7 +49,6 @@ public:
 
 namespace daShutter2_prm {
     inline u8 getSwitchNo(daShutter2_c* item) { return (fopAcM_GetParam(item) >> 0) & 0xFF; }
-    inline u8 getType(daShutter2_c* item) { return (fopAcM_GetParam(item) >> 8) & 0x0F; }
 }
 
 #endif /* D_A_SHUTTER2_H */
--- a/include/d/actor/d_a_shutter2.h
+++ b/include/d/actor/d_a_shutter2.h
@@ -1,22 +1,55 @@
 #ifndef D_A_SHUTTER2_H
 #define D_A_SHUTTER2_H
 
-#include "f_op/f_op_actor.h"
+#include "d/d_bg_s_movebg_actor.h"
 
-class daShutter2_c : public fopAc_ac_c {
+class daShutter2_c : public dBgS_MoveBgActor {
 public:
     BOOL Delete();
-    void CreateHeap();
-    s32 Create();
+    int CreateHeap();
+    int Create();
     s32 _create();
     void set_mtx();
-    void Execute(float(**)[3][4]);
+    int Execute(Mtx**);
     void shutter_move();
     void demo();
     BOOL Draw();
 
+    /* 0x00 */ static const float m_max_speed[1];
+    /* 0x04 */ static const float m_min_speed[1];
+    /* 0x08 */ static const float m_move_len[1];
+    /* 0x0C */ static const float m_width[1];
+    /* 0x10 */ static const float m_height[1];
+    /* 0x14 */ static const s16 m_bdlidx[1];
+    /* 0x16 */ static const s16 m_dzbidx[1];
+    /* 0x18 */ static const s32 m_heapsize[1];
+    /* 0x1C */ static const Vec m_cull_min[1];
+    /* 0x28 */ static const Vec m_cull_max[1];
+
+    static char* m_arcname[1];
+    static char* m_open_ev_name[1];
+    static char* m_close_ev_name[1];
+    static char* m_staff_name[1];
+
 public:
-    /* Place member variables here */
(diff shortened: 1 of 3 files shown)


mattias-blum:
Something went wrong with the commit, will need to fix

mattias-blum:
Was an issue with `configure.py`, git shenanigans adding some garbage to it

include/d/actor/d_a_shutter2.h
```diff
@@ -1,22 +1,55 @@
...
+    /* 0x316 */ s16 mCloseEventIdx;
+    /* 0x318 */ u8 m318;
+    /* 0x319 */ u8 mType;
+    /* 0x31A */ u8 mbIsSwitch;
+    /* 0x31B */ u8 mbIsNearEnemy;
 };
 
+namespace daShutter2_prm {
+    inline u8 getSwitchNo(daShutter2_c* item) { return (fopAcM_GetParam(item) >> 0) & 0xFF; }
+    inline u8 getType(daShutter2_c* item) { return (fopAcM_GetParam(item) >> 8) & 0x0F; }
```
LagoLunatic, line 52:
```suggestion
```
Copypasted, this inline doesn't appear in the debug map

src/d/actor/d_a_shutter2.cpp
```diff
@@ -5,75 +5,230 @@
...
 }
 
 /* 00000350-0000041C       .text _create__12daShutter2_cFv */
 s32 daShutter2_c::_create() {
-    /* Nonmatching */
+    fopAcM_SetupActor(this, daShutter2_c);
+    mType = 0;
+    int result = dComIfG_resLoad(&mPhs, m_arcname[mType]);
+    if (result == cPhs_COMPLEATE_e) {
+        if (MoveBGCreate(m_arcname[mType], m_dzbidx[mType], NULL, m_heapsize[mType]) == 5) {
```
LagoLunatic, line 78:
```suggestion
        if (MoveBGCreate(m_arcname[mType], m_dzbidx[mType], NULL, m_heapsize[mType]) == cPhs_COMPLEATE_e) {
```

src/d/actor/d_a_shutter2.cpp
```diff
@@ -5,75 +5,230 @@
...
+            result = cPhs_ERROR_e;
+        }
+        else {
+            return;
```
LagoLunatic, line 82: Missing return value in a non-void function

src/d/actor/d_a_shutter2.cpp
```diff
@@ -5,75 +5,230 @@
...
+    mDoMtx_stack_c::YrotM(current.angle.y);
+    mpModel->setBaseTRMtx(mDoMtx_stack_c::get());
+    MTXCopy(mDoMtx_stack_c::get(), mMtx);
+    return;
 }
 
 /* 000004AC-00000528       .text Execute__12daShutter2_cFPPA3_A4_f */
-void daShutter2_c::Execute(float(**)[3][4]) {
-    /* Nonmatching */
+BOOL daShutter2_c::Execute(Mtx** i_this) {
```
LagoLunatic, line 99:
```suggestion
BOOL daShutter2_c::Execute(Mtx** pMtx) {
```

src/d/actor/d_a_shutter2.cpp
```diff
@@ -5,75 +5,230 @@
...
 
 /* 000008B0-00000950       .text Draw__12daShutter2_cFv */
 BOOL daShutter2_c::Draw() {
-    /* Nonmatching */
+    g_env_light.settingTevStruct(TEV_TYPE_BG0, &current.pos, &tevStr);
+    g_env_light.setLightTevColorType(mpModel, &tevStr);
+    dComIfGd_setListBG();
+    mDoExt_modelUpdateDL(mpModel);
+    dComIfGd_setList();
+    return 1;
```
LagoLunatic, line 206:
```suggestion
    return TRUE;
```

src/d/actor/d_a_shutter2.cpp
```diff
@@ -5,75 +5,230 @@
...
+    }
+    dComIfGp_evmng_cutEnd(mStaffId);
 }
 
 /* 00000698-000008B0       .text demo__12daShutter2_cFv */
 void daShutter2_c::demo() {
-    /* Nonmatching */
+    u8 isSwitch = fopAcM_isSwitch(this, mSwitchNo);
+    u8 isNearEnemy = fopAcM_myRoomSearchEnemy(fopAcM_GetRoomNo(this)) == NULL;
+    if (m304 == NULL) {
```
LagoLunatic, line 155:
```suggestion
    if (m304 == 0) {
```
Not a pointer

src/d/actor/d_a_shutter2.cpp
```diff
@@ -5,75 +5,230 @@
...
+        case 2: //CLOSE
+        {
+            fVar3 = cLib_addCalc(&current.pos.y, home.pos.y, 0.1f, maxVel, minVel);
+            if (fVar3 == 0.0f){
+                dComIfGp_evmng_cutEnd(mStaffId);
+            }
+            return;
+        }
+    }
+    dComIfGp_evmng_cutEnd(mStaffId);
```
LagoLunatic, line 148: I think all the `return` statements inside this switch are probably supposed to be `break`s. They happen to look identical in this function since the switch is at the end of the function, but I've seen other places where cutEnd went inside a default case along these lines:
```suggestion
            break;
        }
        default:
        {
            dComIfGp_evmng_cutEnd(mStaffId);
            break;
        }
    }
```

src/d/actor/d_a_shutter2.cpp
mattias-blum: This seems like an oversight by the developers, I cleaned up this code in my latest commit but adding any sort of return statement after the if statement causes it not to match. You would think it should return `cPhs_COMPLEATE_e` in the default case but it just doesn't apparently.

mattias-blum:
PR changes made

PR #693, opened 2025-02-24
d_a_leaflift 100%

mattias-blum:


mattias-blum:
I'm going to add `d_a_leaflift_static` to this as well since the methods are defined in `d_a_leaflift.h`

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
 }
 
+static void rideCallBack(dBgW* param1, fopAc_ac_c* i_this, fopAc_ac_c* i_other);
+
 /* 000001E0-00000338       .text CreateHeap__9daLlift_cFv */
-void daLlift_c::CreateHeap() {
-    /* Nonmatching */
-}
+BOOL daLlift_c::CreateHeap() {
+    J3DModelData* modelData = (J3DModelData *)dComIfG_getObjectRes(m_arcname, TRFLAG_BDL_ETHATA);
```
LagoLunatic, line 78: Wrong file ID enum used here, arcname's value is "olift" not "trflag'

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
+    mpModel = mDoExt_J3DModel__create(modelData, 0x80000, 0x11000022);
+    if (!mpModel) {
+        return FALSE;
+    }
+
+    mpModel->setUserArea((u32)this);
+    mcBgW = new dBgW();
+    if (mcBgW) {
+        cBgD_t* pData = (cBgD_t *)dComIfG_getObjectRes(m_arcname, TRFLAG_BTI_ETHATA);
+        if (mcBgW->Set(pData, 0x1, &mMtx) == 1) {
```
LagoLunatic, line 90:
```suggestion
        if (mcBgW->Set(pData, cBgW::MOVE_BG_e, &mMtx) == true) {
```

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
+        fVar2 = posOffset.abs();
+        if (!posOffset.normalizeRS() ) {
+          return;
+        }
+        cLib_addCalcAngleS2(&param2->m440, -fVar2 * fVar1, 8, 0x200);
+        fVar1 = cM_ssin(param2->m440);
+        param2->mZeroQuat.x = posOffset.x * fVar1;
+        param2->mZeroQuat.y = posOffset.y * fVar1;
+        param2->mZeroQuat.z = posOffset.z * fVar1;
+        param2->mZeroQuat.w = cM_scos(param2->m440);
```
LagoLunatic, line 219: `mZeroQuat` is a misleading name for this member. It's initialized to zero, but here it's being set to a nonzero value.

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
+    if (fopAcM_GetName(param3) == 0xa9) {
+        param2->m43C = 1;
+        param2->m43E = 1;
+        posOffset = posOffset.outprod(up_vec);
+        mDoMtx_stack_c::YrotS(-param2->current.angle.y);
+        mDoMtx_stack_c::multVec(&posOffset, &posOffset);
+        if (param2->m43D) {
+            posOffset.x = param3->current.pos.x;
+            posOffset.y = param3->current.pos.y;
+            posOffset.z = param3->current.pos.z;
```
LagoLunatic, line 203:
```suggestion
            posOffset = param3->current.pos;
```

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
+static void rideCallBack(dBgW* param1, fopAc_ac_c* i_this, fopAc_ac_c* i_other) {
+    f32 fVar1;
+    f32 fVar2;
+    daLlift_c* param2 = (daLlift_c*)i_this;
+    daLlift_c* param3 = (daLlift_c*)i_other;
+
+    cXyz posOffset = param3->current.pos - param2->current.pos;
+    
+    fVar1 = 2.0f;
+    if (fopAcM_GetName(param3) == 0xa9) {
```
LagoLunatic, line 194: fopAcM_GetName returns the enum in `d_procname.h`

src/d/d_a_leaflift_static.cpp
```diff
@@ -4,19 +4,43 @@
...
 
 const f32 daLlift_c::m_height = 560.0f;
 const f32 daLlift_c::m_max_speed = 10.0f;
 const f32 daLlift_c::m_min_speed = 5.0f;
 
 /* 800690E4-80069100       .text checkEndDownLift__9daLlift_cFv */
-void daLlift_c::checkEndDownLift() {
-    /* Nonmatching */
+s32 daLlift_c::checkEndDownLift() {
+    return (((current.pos.y <= home.pos.y) << 1) << 0x1c) >> 0x1d;
```
LagoLunatic, line 16: This looks like a simple comparison that the compiler did a branchless optimization to. Have you tried using https://roeming.github.io/C-Math-Evaluator/ to help you simplify it? Keep in mind you may also need to mess with the return type of this function too, like BOOL or bool

src/d/d_a_leaflift_static.cpp
```diff
@@ -4,19 +4,43 @@
...
 }
 
 /* 80069100-800692C4       .text MoveUpLift__9daLlift_cFv */
-void daLlift_c::MoveUpLift() {
-    /* Nonmatching */
+s32 daLlift_c::MoveUpLift() {
+    cXyz upLiftPos[2];
+    u8 res = 0;
```
LagoLunatic, line 22:
```suggestion
BOOL daLlift_c::MoveUpLift() {
    cXyz upLiftPos[2];
    bool res = false;
```
Does this match? (and res = true below)

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
-    /* Nonmatching */
+    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -300.0f, -600.0f, -300.0f, 300.0f, 100.0f, 300.0f);
+    fopAcM_setCullSizeFar(this, 1.0f);
+    mStts.Init(0xFF,0xFF,this);
+    mCyl.Set(l_cyl_src);
+    mCyl.SetStts(&mStts);
+    cXyz local_28 = current.pos;
+    local_28.y += 200.0f;
+    mUpLift = dBgS_ObjGndChk_Wtr_Func(local_28);
```
LagoLunatic, line 114:
```suggestion
    mWaterY = dBgS_ObjGndChk_Wtr_Func(local_28);
```
This func returns the y position of the nearest water collision located below the given point, so maybe a name like this would be more clear

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
 /* 00000FE0-000010E0       .text MoveDownLift__9daLlift_cFv */
-void daLlift_c::MoveDownLift() {
-    /* Nonmatching */
+BOOL daLlift_c::MoveDownLift() {
+    float maxSpeed = m_max_speed;
+    float minSpeed = m_min_speed;
+    if (!m468) {
+      return TRUE;
+    }
+    float fVar4 = cLib_addCalc(&current.pos.y, home.pos.y, 0.1f, maxSpeed, minSpeed);
```
LagoLunatic, line 304: Could name the local like this to match the name `upVel` in the other func
```suggestion
    float downVel = cLib_addCalc(&current.pos.y, home.pos.y, 0.1f, maxSpeed, minSpeed);
```

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,352 @@
...
+    mCyl.SetC(local_28);
+    dComIfG_Ccsp()->Set(&mCyl);
+    emitterCtrl();
+    return TRUE;
 }
 
 /* 00000F24-00000FE0       .text emitterCtrl__9daLlift_cFv */
 void daLlift_c::emitterCtrl() {
-    /* Nonmatching */
+    int iVar1;
```
LagoLunatic, line 276: Unused

include/d/actor/d_a_leaflift.h
```diff
@@ -2,29 +2,66 @@
...
 
+    static const char m_arcname[6];
+
 public:
-    /* Place member variables here */
+    /* 0x290 */ request_of_phase_process_class mPhs;
+    /* 0x298 */ J3DModel* mpModel;
+    /* 0x29C */	dCcD_Stts mStts;
+    /* 0x2D8 */ dCcD_Cyl mCyl;
+    /* 0x408 */ dBgW* mcBgW;
```
LagoLunatic, line 39: This is a dBgW not a cBgW.

I normally name these `mpBgW` since the c/d prefix just indicates which library they're a part of and don't add any information in this context (c = SSystem library, d = dolzel aka zelda-specific subclasses)

src/d/d_a_leaflift_static.cpp
LagoLunatic: This still applies

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,348 @@
...
+        posOffset = posOffset.outprod(up_vec);
+        mDoMtx_stack_c::YrotS(-param2->current.angle.y);
+        mDoMtx_stack_c::multVec(&posOffset, &posOffset);
+        if (param2->m43D) {
+            posOffset = param3->current.pos;
+            posOffset.x += 200.0f;
+            posOffset = posOffset.outprod(up_vec);
+            mDoMtx_stack_c::YrotS(-param2->current.angle.y);
+            mDoMtx_stack_c::multVec(&posOffset, &posOffset);
+            tiltFactor = cM_ssin((param2->mGlideFrameCounter & 0x1F) << 11) * 2.0f * 0.25f * (cM_rndFX(0.2f) + 1.0f);
```
LagoLunatic, line 206: This looks like an optimization too, come to think of it. Try something like:
```suggestion
            s16 angle = (param2->mGlideFrameCounter % 32) * (0x10000 / 32);
            tiltFactor = cM_ssin(angle * 2.0f * 0.25f * (cM_rndFX(0.2f) + 1.0f);
```
If I'm understanding this math right, I think this angle is basically cycling over all 360 degrees (0x10000 cSAngles) over the course of every 32 frames?

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,348 @@
...
+    return;
 }
 
 /* 00000D9C-00000F24       .text _execute__9daLlift_cFv */
 bool daLlift_c::_execute() {
-    /* Nonmatching */
+    Quaternion interpolatedRotation;
+    cXyz adjustedPosition;
+    float distXZ = fopAcM_searchActorDistanceXZ(this, dComIfGp_getPlayer(0));
+    mGlideFrameCounter++;
```
LagoLunatic, line 246: Why is this called "glide counter"? I might be misunderstanding something but I don't think gliding is relevant here, it looks like generic timer that counts the total lifetime of the actor.

If you're not sure what an actor is you can download the latest [GCFT nightly build](https://nightly.link/LagoLunatic/GCFT/workflows/python-app/master?preview) and open the actor .arc up to view what its model looks like:
image
"leaflift" is the big lilypad-esque platform that elevates Link up to the great deku tree's height.
(The deku leaf is referred to as a "fan" in the game's code (e.g. `daPy_lk_c::procFanGlide_init`), not a "leaf", if that's what you were thinking.)

src/d/actor/d_a_leaflift.cpp
```diff
@@ -4,98 +4,348 @@
...
 }
 
 /* 000009C4-00000CAC       .text rideCallBack__FP4dBgWP10fopAc_ac_cP10fopAc_ac_c */
-void rideCallBack(dBgW*, fopAc_ac_c*, fopAc_ac_c*) {
-    /* Nonmatching */
+static void rideCallBack(dBgW* param1, fopAc_ac_c* i_this, fopAc_ac_c* i_other) {
+    f32 tiltFactor;
+    f32 offsetMagnitude;
+    daLlift_c* param2 = (daLlift_c*)i_this;
+    daLlift_c* param3 = (daLlift_c*)i_other;
```
LagoLunatic, line 189: Incorrect cast, the `fopAc_ac_c* i_other` param to a ride callback can be any type of actor that goes on top of this platform, and it's unlikely that a lilypad platform would land on another lilypad platform.

If you want to cast it, then inside the `if (fopAcM_GetName(param3) == PROC_PLAYER) {` block below you could cast i_other to be of type `daPy_lk_c*` (aka Link's class) because the game explicitly checks that it's Link in that block. (Not really necessary though, since no Link-specific methods are called on this param, so leaving it as `fopAc_ac_c*` is also fine.)

src/d/actor/d_a_leaflift.cpp
```diff
@@ -298,21 +294,21 @@ void daLlift_c::emitterCtrl() {
...
         if (mEmitter3) {
             mEmitter3->playCreateParticle();
         }
         return TRUE;
     }
-    if (fVar4 != 0.0f && !m469) {
+    if (downVel != 0.0f && !mbIsUpdraftBoosted) {
         fopAcM_seStart(this, JA_SE_OBJ_LOTUS_LIFT_DOWN, 0);
-        m469 = 1;
+        mbIsUpdraftBoosted = TRUE;
```
LagoLunatic, line 311: Similar to "glide counter", "updraft boosted" sounds like something that would be related to the deku leaf, not a lilypad platform

src/d/actor/d_a_leaflift.cpp
mattias-blum: Okay that clears up a lot because I figured this actor was Link with the Deku Leaf out but clearly that was the wrong assumption. This tool is super helpful thanks for sharing that. I'll use it in the future to confirm what actors I'm working on.

src/d/actor/d_a_leaflift.cpp
mattias-blum: Yeah same deal here as above, just a wrong assumption as to what this actor was.

src/d/actor/d_a_leaflift.cpp
LagoLunatic: Yeah Link is just a single actor, and the code for his various states is organized by splitting them up into different `.inc` files, which are not standalone actors (e.g. `d_a_player_fan.inc` has Link's deku leaf related methods in it).

mattias-blum: Is there a tutorial on how to use GCFT? I tried loading the .iso of the game but it gave the error in the image
Screenshot 2025-02-25 at 6 12 56 PM

LagoLunatic: I've never seen that error before, but did you do either of the follow things?

* Forget to extract the zip file and run the exe from within the archive
* Extract the zip file to some protected system location like C:/Program Files that GCFT doesn't have permission to write to

(GCFT is trying to create settings.txt next to the executable and it's failing)

mattias-blum: I'm running it on Mac so perhaps it's a permissions issue

LagoLunatic: Oh right... when I copied the mac build stuff from wwrando to gcft I must have not properly handled settings.txt. You're probably the first person to try running it on a mac so I guess it's just been broken this whole time.

I'll try to fix mac builds at some point, but for now I think you could just GCFT from source if you want and that should work even on mac

mattias-blum: Okay I'll try that

src/d/actor/d_a_leaflift.cpp
mattias-blum: That is the correct equivalent function but the code doesn't match when doing that.

LagoLunatic: Ok, for future reference it turned out the reason my suggestion didn't match was because the 32 literal was signed (not sure why that matters...). Using `32U` instead of `32` matches:
```cpp
s16 angle = (i_this->mActorLifetimeFrameCount % 32U) * (0x10000 / 32);
```

PR #694, opened 2025-02-26
d_a_tag_so 100%

mattias-blum:
Small one, nothing to comment on

### DIFF
--- a/src/d/actor/d_a_tag_so.cpp
+++ b/src/d/actor/d_a_tag_so.cpp
@@ -4,66 +4,92 @@
 //
 
 #include "d/actor/d_a_tag_so.h"
+#include "d/d_lib.h"
 #include "d/d_procname.h"
 
 /* 000000EC-0000010C       .text __ct__14daTag_So_HIO_cFv */
 daTag_So_HIO_c::daTag_So_HIO_c() {
-    /* Nonmatching */
+    mNo = -1;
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_tag_so.cpp
```diff
@@ -4,66 +4,92 @@
...
 }
 
 /* 000001C4-00000224       .text getArg__10daTag_So_cFv */
 void daTag_So_c::getArg() {
-    /* Nonmatching */
+    s32 param;
+    s32 paramBit;
+    
+    param = fopAcM_GetParam(this);
+    mParam = param;
```
LagoLunatic, line 46:
```suggestion
    m290 = fopAcM_GetParamBit(param, 0, 8);
```
Because the other 2 params use this inline, we can probably assume this one does too instead of relying the implicit cast to mask it. And "mParam" is too vague of a name so it's better to leave it unnamed

src/d/actor/d_a_tag_so.cpp
```diff
@@ -4,66 +4,92 @@
...
+    paramBit = fopAcM_GetParamBit(param, 8, 8);
```
LagoLunatic, line 47:
```suggestion
    paramRadius = fopAcM_GetParamBit(param, 8, 8);
```
Too vague of a name

PR #696, opened 2025-02-26
d_a_obj_hat 100%

njulrik:


LagoLunatic:
You accidentally added the zip of debug maps into the repo.

Also, is this actor fully matched, or still a WIP? If it's fully matched, mark it as `Matching` in configure.py and remove all the `/* Nonmatching */` comments.

include/d/actor/d_a_obj_hat.h
```diff
@@ -1,25 +1,60 @@
...
+
+    /* 0x290 */ request_of_phase_process_class mPhs;
+    /* 0x298 */ J3DModel* model;
+    /* 0x29c */ mDoExt_McaMorf* mpMorf;
+    /* 0x2a0 */ Mtx mMtx;
+
+    /* 0x2d0 */ dBgS_ObjAcch acch;
+    /* 0x494 */ dBgS_AcchCir acch_cir;
+    /* 0x4d4 */ dCcD_Stts stts;
+    /* 0x510 */ dCcD_Cyl cyl;
```
LagoLunatic, line 50:
```suggestion
    /* 0x2d0 */ dBgS_ObjAcch mpAcch;
    /* 0x494 */ dBgS_AcchCir mAcchCir;
    /* 0x4d4 */ dCcD_Stts mStts;
    /* 0x510 */ dCcD_Cyl mCyl;
```

I don't think it's a hard rule, but generally it seems like the original devs used `this_naming_style` only for C-style actors, while they used `mThisNamingStyle` for C++-style actors. This one is a C++ style actor. (Same suggestion applies to the other fields in this actor.)

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
-    /* Nonmatching */
+static BOOL daSampleDraw(void* arg) {
+    static_cast<daObjHat_c*>(arg)->_draw();
 }
 
 /* 00000AA0-00000AA8       .text daSampleIsDelete__FPv */
-static BOOL daSampleIsDelete(void*) {
-    /* Nonmatching */
+static BOOL daSampleIsDelete(void* arg) {
+    return 1;
```
LagoLunatic, line 238:
```suggestion
    return TRUE;
```

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
+    speed.y = 50.0f;
+    speed.z *= 20.0f;
+
+    fopAcM_SetSpeed(this, speed.x, speed.y, speed.z);
 }
 
 /* 00000A20-00000A40       .text daSampleCreate__FPv */
-static s32 daSampleCreate(void*) {
-    /* Nonmatching */
+static s32 daSampleCreate(void* arg) {
```
LagoLunatic, line 217:
```suggestion
static s32 daSampleCreate(void* i_this) {
```

The parameter holding the current actor is pretty much always named `i_this`. Same suggestion applies to other funcs

include/d/actor/d_a_obj_hat.h
```diff
@@ -1,25 +1,60 @@
...
+
+    /* 0x640 */ cXyz angle;
```
LagoLunatic, line 52:
```suggestion
    /* 0x640 */ cXyz mSpeed;
```

This doesn't make sense, angles are stored in s16 or csXyz, while cXyz holds a vector of floats. Based on this being set in the function `setSpeed` below, this could probably be named `mSpeed`.

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
+    &daObjHat_c::executeNormal,
+};
+
+
 /* 0000073C-00000884       .text _execute__10daObjHat_cFv */
 bool daObjHat_c::_execute() {
-    /* Nonmatching */
+    // dCcD_Sph::~dCcD_Sp
+
+    (this->*moveProc[this->procid])();
```
LagoLunatic, line 149: "Proc ID" means something different in this codebase (see `fpc_ProcID`). This should be called the state or similar.
```suggestion
    (this->*moveProc[this->mState])();
```

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
+    return FALSE;
 }
 
 /* 000005D4-000006AC       .text createInit__10daObjHat_cFv */
-void daObjHat_c::createInit() {
-    /* Nonmatching */
+s32 daObjHat_c::createInit() {
+    stts.Init(2, 0xff, this);
+    cyl.Set(l_cyl_src);
+    this->cyl.SetStts(&this->stts);
```
LagoLunatic, line 112:
```suggestion
    mCyl.SetStts(&mStts);
```
Typing `this->` is unnecessary. This also applies everywhere else in this file.

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
+    this->cullMtx = this->model->mBaseTransformMtx;
```
LagoLunatic, line 113:
```suggestion
    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
```
Missing two inlines here. You can see inlines by referencing the debug map for this actor.

include/d/actor/d_a_obj_hat.h
```diff
@@ -1,25 +1,60 @@
...
+    /* 0x64c */ u8 procid;
+    /* 0x64d */ u8 id;
+    /* 0x64e */ u8 byte2;
+    /* 0x64f */ u8 byte3;
```
LagoLunatic, line 56:
```suggestion
```
You don't need to add unused members to pad the struct in this particular case, the compiler will pad it to 0x4 byte alignment so it will stay 0x650 bytes with these removed.

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
 /* 000008B4-00000964       .text setMtx__10daObjHat_cFv */
 void daObjHat_c::setMtx() {
-    /* Nonmatching */
+    model->mBaseScale = this->scale;
+    MTXTrans(mDoMtx_stack_c::now, this->current.pos.x, this->current.pos.y,
+             this->current.pos.z);
+    mDoMtx_YrotM(mDoMtx_stack_c::now, this->current.angle.y + 0x4000);
+    mDoMtx_XrotM(mDoMtx_stack_c::now, this->current.angle.x);
+    mDoMtx_ZrotM(mDoMtx_stack_c::now, this->current.angle.z + 0x4000);
+    MTXCopy(mDoMtx_stack_c::now, this->model->mBaseTransformMtx);
```
LagoLunatic, line 197: Every line in this function should be using an inline. Look at the debug map.

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
+#include "d/d_kankyo.h"
 #include "d/d_procname.h"
+#include "d/res/res_ro.h"
+
+// const char daObjHat_c::m_arcname[] = "ObjHat";
 
 /* 00000078-00000184       .text __ct__10daObjHat_cFv */
 daObjHat_c::daObjHat_c() {
+    u32 hatno = getPrmHatNo();
+    this->id = hatno;
```
LagoLunatic, line 19:
```suggestion
    this->mHatNo = getPrmHatNo();
```
Since we already know the official name is "HatNo", might as well name the field that instead of "id".

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,238 @@
...
 
 #include "d/actor/d_a_obj_hat.h"
+#include "d/actor/d_a_npc_roten.h"
+#include "d/d_a_obj.h"
+#include "d/d_com_inf_game.h"
+#include "d/d_kankyo.h"
 #include "d/d_procname.h"
+#include "d/res/res_ro.h"
+
+// const char daObjHat_c::m_arcname[] = "ObjHat";
```
LagoLunatic, line 14:
```suggestion
// const char daObjHat_c::m_arcname[] = "ObjHat";
```
Why is this commented out?

include/d/actor/d_a_obj_hat.h
njulrik: If I'm reading the code correctly, the variable really holds a normal vector, which is scaled by `speedF`  in `_execute`. Is there a more descriptive name for this (e.g. `mNormalSpeed` or something) or is just `mSpeed` fine?

LagoLunatic: Oh I missed the call that normalizes it. In that case it could be called `mMoveNorm` or similar, since it only holds the direction and not the speed.

LagoLunatic:
Looks like the TU doesn't 100% match. Do any of the sections in objdiff show < 100%? If not, try adding the nosyminline or sym off flags in configure.py to see if it's a weak function ordering issue. You can check if it matches locally (before pushing to GitHub) by running `ninja`.

njulrik:
The build doesn't succeed with either of those flags. Could it be related to inlines? I made much of this before I found the debug maps, so I can see a number of inlines that I didn't use.

njulrik:
Alright, pushed those changes. Hopefully merge-ready now.

configure.py
```diff
@@ -1656,7 +1656,7 @@ def MatchingFor(*versions):
     ActorRel(NonMatching, "d_a_obj_hami2"),
     ActorRel(NonMatching, "d_a_obj_hami3"),
     ActorRel(NonMatching, "d_a_obj_hami4"),
-    ActorRel(NonMatching, "d_a_obj_hat"),
+    ActorRel(Matching,    "d_a_obj_hat"),
```
LagoLunatic, line 1659: Still need to revert this

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+#include "d/actor/d_a_npc_roten.h"
+#include "d/d_a_obj.h"
+#include "d/d_com_inf_game.h"
+#include "d/d_kankyo.h"
 #include "d/d_procname.h"
+#include "d/res/res_ro.h"
+#include "JSystem/J3DGraphAnimator/J3DModel.h"
+
+static const int l_bmd_ix_tbl[] = {RO_BDL_RO_HAT, RO_BDL_RO_HAT2, RO_BDL_RO_HAT3, RO_BDL_RO_HAT};
+static const int l_bck_ix_tbl[] = {RO_INDEX_BCK_HAT_WID, RO_BCK_HAT2_WIND, RO_BCK_HAT3_WID,
```
LagoLunatic, line 16: The BCK list shouldn't be using the `INDEX` enum here, as its being passed to `dComIfG_getObjectIDRes` later. `RO_RES_FILE_ID` gets passed to `dComIfG_getObjectIDRes`, while `RO_RES_FILE_INDEX` gets passed to dComIfG_getObjectRes`.

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+                                   RO_BCK_HAT_WID};
+static dCcD_SrcCyl l_cyl_src = {
+    // dCcD_SrcGObjInf
+    {
+        /* Flags             */ 0,
+        /* SrcObjAt  Type    */ 0,//AT_TYPE_UNK1000,
```
LagoLunatic, line 22: Old comment?

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+        /* SrcObjAt  Atp     */ 0,
+        /* SrcObjAt  SPrm    */ 0,
+        /* SrcObjTg  Type    */ AT_TYPE_WIND,
+        /* SrcObjTg  SPrm    */ cCcD_TgSPrm_Set_e | cCcD_TgSPrm_IsEnemy_e,
+        /* SrcObjCo  SPrm    */ cCcD_CoSPrm_Set_e | cCcD_CoSPrm_IsOther_e | cCcD_CoSPrm_VsGrpAll_e,
+        /* SrcGObjAt Se      */ 0,
+        /* SrcGObjAt HitMark */ 0,
+        /* SrcGObjAt Spl     */ 0,
+        /* SrcGObjAt Mtrl    */ 0,
+        /* SrcGObjAt SPrm    */ 0,//AT_TYPE_WIND,
```
LagoLunatic, line 32: Old comment?

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
-static BOOL CheckCreateHeap(fopAc_ac_c*) {
-    /* Nonmatching */
+static BOOL CheckCreateHeap(fopAc_ac_c* actor) {
+    return ((daObjHat_c*)actor)->createHeap();
 }
 
 /* 000003CC-0000045C       .text _create__10daObjHat_cFv */
 s32 daObjHat_c::_create() {
-    /* Nonmatching */
+    // if (this->actor_condition & )
```
LagoLunatic, line 66: Old comment

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+    fopAcM_SetupActor(this, daObjHat_c);
+
+    int var = dComIfG_resLoad(&mPhs, "Ro");
```
LagoLunatic, line 69:
```suggestion
    int phase_state = dComIfG_resLoad(&mPhs, "Ro");
```

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+
+    if (var == cPhs_COMPLEATE_e) {
+        if (!fopAcM_entrySolidHeap((fopAc_ac_c*)this, &CheckCreateHeap, 0uL)) {
```
LagoLunatic, line 72:
```suggestion
        if (!fopAcM_entrySolidHeap(this, &CheckCreateHeap, 0uL)) {
```
This cast shouldn't be necessary (unless you added it to fix regalloc)

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+            return cPhs_ERROR_e;
+        } else {
+            return createInit();
+        }
+    }
 }
```
LagoLunatic, line 78: Missing return in a non-void function. You probably want `return phase_state;`

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+    J3DModelData* pModelData = (J3DModelData*)dComIfG_getObjectIDRes("Ro", l_bmd_ix_tbl[this->mHatNo]);
+    if (pModelData == NULL) {
+        return FALSE;
+    } else {
+        mDoExt_McaMorf* morf = new mDoExt_McaMorf(
+            pModelData, NULL, NULL,
+            (J3DAnmTransformKey*)dComIfG_getObjectIDRes("Ro", l_bck_ix_tbl[mHatNo]),
+            J3DFrameCtrl::LOOP_REPEAT_e, 1.0f, 0, -1, 1, NULL, 0x80000, 0x37441422);
+
+        this->mpMorf = morf;
```
LagoLunatic, line 91:
```suggestion
        mpMorf = morf;
```
As mentioned earlier, typing `this->` out explicitly isn't required in C++

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+    mStts.Init(2, 0xff, this);
+    mCyl.Set(l_cyl_src);
+    mCyl.SetStts(&mStts);
+    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -50.0f, 0.0f, -50.0f, 50.0f, 200.0f, 50.0f);
+    gravity = -5.0f;
+
+    fopAc_ac_c* parent;
+    int res = fopAcM_SearchByID(fopAcM_GetLinkId(this), &parent);
+    if ((res != 0) && (parent != NULL)) {
```
LagoLunatic, line 122:
```suggestion
    if (fopAcM_SearchByID(fopAcM_GetLinkId(this), &parent) && (parent != NULL)) {
```

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+        fopAcM_SetSpeedF(this, 0);
+    }
+    if (mAcch.ChkGroundHit()) {
+        cLib_addCalc(&this->speedF, 0.0, 0.3, 1000.0, 1.0);
+    }
+
+    mCyl.SetC(this->current.pos);
+    dComIfG_Ccsp()->Set(&mCyl);
+    u32 res = mCyl.ChkTgHit();
+    if (res != 0) {
```
LagoLunatic, line 175:
```suggestion
    if (mCyl.ChkTgHit()) {
```

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+    // because downwards vector quantity has negative value
+    if (ymove < fopAcM_GetMaxFallSpeed(this)) {
+        ymove = fopAcM_GetMaxFallSpeed(this);
+    }
+    fopAcM_SetSpeed(this, xspeed, ymove, zspeed);
+
+    fopAcM_posMove(this, mStts.GetCCMoveP());
+    mAcch.CrrPos(g_dComIfG_gameInfo.play.mBgS);
+    if (mAcch.ChkWallHit()) {
+        fopAcM_SetSpeedF(this, 0);
```
LagoLunatic, line 166:
```suggestion
        fopAcM_SetSpeedF(this, 0.0f);
```

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+    }
+    if (mAcch.ChkGroundHit()) {
+        cLib_addCalc(&this->speedF, 0.0, 0.3, 1000.0, 1.0);
```
LagoLunatic, line 169:
```suggestion
        cLib_addCalc(&this->speedF, 0.0f, 0.3f, 1000.0f, 1.0f);
```

src/d/actor/d_a_obj_hat.cpp
```diff
@@ -4,91 +4,241 @@
...
+
+    setMtx();
+    return TRUE;
 }
 
 /* 00000884-00000888       .text executeNormal__10daObjHat_cFv */
-void daObjHat_c::executeNormal() {
-    /* Nonmatching */
-}
+void daObjHat_c::executeNormal() { /* empty? */ }
```
LagoLunatic, line 184:
```suggestion
void daObjHat_c::executeNormal() { }
```
Yep it's empty

PR #697, opened 2025-02-28
d_a_tornado

ross3102:
this is my first time doing this, so I'm not sure if there's anything specific I'm supposed to do other than get 100% binary match

my concerns:
* there's some inlines in d_a_tornado.h that are still unimplemented. looking at the debug maps it seems these are called only from daShip_c, but I couldn't figure out what they're meant to return based on the daShip_c ghidra output.
* I guessed the member variable names as best I could, but not 100% sure if they make sense
* I left some bad local names, and used extra temp vars in order to get to 100% in `execute`
* I'm not sure if there's a way to verify return types (e.g. BOOL vs int)

### DIFF
--- a/include/d/actor/d_a_ship.h
+++ b/include/d/actor/d_a_ship.h
@@ -64,7 +64,7 @@ public:
     void offFantomGanonBattle() {}
     void offStateFlg(daSHIP_SFLG) {}
     void offTornadoFlg() {
-        mTornadoID = -1;
+        mTornadoID = fpcM_ERROR_PROCESS_ID_e;
         mTornadoActor = NULL;
     }
     void offWhirlFlg() {}
@@ -77,7 +77,7 @@ public:
     void onSceneChange() {}
     void onShortHitFlg() {}
     void onStateFlg(daSHIP_SFLG) {}
-    void onTornadoFlg(int tornadoID) { mTornadoID = tornadoID; }
+    void onTornadoFlg(u32 tornadoID) { mTornadoID = tornadoID; }
     void onWhirlFlg(unsigned long, short) {}
     void onWhirlFlgDirect(unsigned long, short) {}
     void setAtnPos(const cXyz*) {}
@@ -241,7 +241,7 @@ public:
     /* 0x040C */ f32 m040C;
     /* 0x0410 */ int mGridId;
     /* 0x0414 */ daGrid_c* mpGrid;
-    /* 0x0418 */ int mTornadoID;
+    /* 0x0418 */ u32 mTornadoID;
     /* 0x041C */ fopAc_ac_c* mTornadoActor;
     /* 0x0420 */ u32 m0420;
     /* 0x0424 */ fopAc_ac_c* m0424;
--- a/include/d/actor/d_a_ship.h
+++ b/include/d/actor/d_a_ship.h
@@ -23,7 +23,7 @@ public:
     u32 getTactWarpID() { return mTactWarpID; }
     void setTactWarpID(u32 warpID) { mTactWarpID = warpID; }
     
-    bool unknown_inline_TODO() const { return m041C || m0424; }
+    bool unknown_inline_TODO() const { return mTornadoActor || m0424; }
     
     void checkCraneMode() const {}
     void checkCraneUpEnd() const {}
@@ -58,12 +58,15 @@ public:
     void getTactJntMtx() {}
     void getTillerAngleRate() {}
     void getTillerTopPosP() {}
-    void getTornadoActor() {}
+    fopAc_ac_c* getTornadoActor() { return mTornadoActor; }
     void getWhirlActor() {}
     void offCraneHookFlg() {}
     void offFantomGanonBattle() {}
     void offStateFlg(daSHIP_SFLG) {}
-    void offTornadoFlg() {}
+    void offTornadoFlg() {
+        mTornadoID = -1;
+        mTornadoActor = NULL;
+    }
     void offWhirlFlg() {}
     void onCb1Ride() {}
     void onCraneHookFlg() {}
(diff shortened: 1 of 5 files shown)


LagoLunatic:
Looks like the binary doesn't 100% match. Do any of the sections in objdiff show < 100%? If not, try adding the nosyminline or sym off flags in configure.py. You can check if it matches locally by running `ninja`.

> there's some inlines in d_a_tornado.h that are still unimplemented. looking at the debug maps it seems these are called only from daShip_c, but I couldn't figure out what they're meant to return based on the daShip_c ghidra output.

That's normal, implementing inlines can be almost impossible when the code using them hasn't been decompiled yet. Whoever does ship will get them.

> I'm not sure if there's a way to verify return types (e.g. BOOL vs int)

BOOL is a typedef for int so there's no difference in terms of matching. But you can guess if an int return value should be BOOL if it only returns 0 or 1, *and* the callers all only check if it's 0. Not a big deal to leave it as int if you're unsure.

include/d/actor/d_a_ship.h
```diff
@@ -74,7 +77,7 @@ class daShip_c : public fopAc_ac_c {
     void onSceneChange() {}
     void onShortHitFlg() {}
     void onStateFlg(daSHIP_SFLG) {}
-    void onTornadoFlg(unsigned long) {}
+    void onTornadoFlg(int tornadoID) { mTornadoID = tornadoID; }
```
LagoLunatic, line 80:
```suggestion
    void onTornadoFlg(u32 tornadoID) { mTornadoID = tornadoID; }
```
We know this inline takes a `u32` (aka unsigned long) from the debug maps. You shouldn't change function parameter types, only return types and field types.

include/d/actor/d_a_ship.h
```diff
@@ -238,8 +241,8 @@ class daShip_c : public fopAc_ac_c {
     /* 0x040C */ f32 m040C;
     /* 0x0410 */ int mGridId;
     /* 0x0414 */ daGrid_c* mpGrid;
-    /* 0x0418 */ int m0418;
-    /* 0x041C */ fopAc_ac_c* m041C;
+    /* 0x0418 */ int mTornadoID;
```
LagoLunatic, line 244:
```suggestion
    /* 0x0418 */ u32 mTornadoID;
```
Proc IDs normally use the `fpc_ProcID` type (aka uint), but based on onTornadoFlg taking a u32 as its parameter this one may be typed as `u32`. Either way, I don't think `int` is correct, that's just a guess we had in Ghidra because this field was unknown, you might need to change these initial guesses to match the debug maps.

include/d/actor/d_a_ship.h
```diff
@@ -58,12 +58,15 @@ class daShip_c : public fopAc_ac_c {
...
     void getTillerTopPosP() {}
-    void getTornadoActor() {}
+    fopAc_ac_c* getTornadoActor() { return mTornadoActor; }
     void getWhirlActor() {}
     void offCraneHookFlg() {}
     void offFantomGanonBattle() {}
     void offStateFlg(daSHIP_SFLG) {}
-    void offTornadoFlg() {}
+    void offTornadoFlg() {
+        mTornadoID = -1;
```
LagoLunatic, line 67:
```suggestion
        mTornadoID = fpcM_ERROR_PROCESS_ID_e;
```

src/d/actor/d_a_tornado.cpp
```diff
@@ -4,72 +4,342 @@
...
+                ship->onTornadoFlg(fopAcM_GetID(this));
+                speedF = 0.0f;
+            }
+        }
+
+        if (!dComIfGp_event_runCheck()) {
+            fopAcM_posMoveF(this, NULL);
+            cXyz diff = current.pos - home.pos;
+            diff.y = 0;
+            if (diff.abs2XZ() > 56250000) {
```
LagoLunatic, line 169:
```suggestion
            if (diff.abs2XZ() > 7500.0f*7500.0f) {
```
abs2 means it returns a squared distance, so it's more readable to write the literal out as squared too (the compiler optimizes it)

src/d/actor/d_a_tornado.cpp
```diff
@@ -4,72 +4,342 @@
...
+    if (param == 0) {
+        if (!dComIfGp_event_runCheck()) {
+            short target = fopAcM_searchActorAngleY(this, dComIfGp_getLinkPlayer());
+            cLib_addCalcAngleS(&current.angle.y, target, 10, 0x1000, 0x100);
+            cLib_chaseF(&speedF, 20.0f, 0.2f);
+        }
+        if (dComIfGp_checkPlayerStatus0(0, daPyStts0_SHIP_RIDE_e) && dComIfGp_getShipActor() != NULL) {
+            daShip_c* ship = dComIfGp_getShipActor();
+            cXyz diff = ship->current.pos - current.pos;
+            if (diff.abs2XZ() < 100000000.0f) {
```
LagoLunatic, line 159:
```suggestion
            if (diff.abs2XZ() < 10000.0f*10000.0f) {
```

src/d/actor/d_a_tornado.cpp
```diff
@@ -4,72 +4,342 @@
...
+        if (mPtclTimer != 0) {
+            mPtclTimer -= 1;
+        } else {
+            mPtclTimer = 10;
+            dComIfGp_particle_set(0x8213, &current.pos, NULL, (cXyz*)&wind_scale);
+        }
+
+        fopAcM_seStartCurrent(this, JA_SE_OBJ_TORNADE_SUS, 100);
+
+        if (dComIfGs_isEventBit(10000)) {
```
LagoLunatic, line 184:
```suggestion
        if (dComIfGs_isEventBit(0x2710)) {
```
Memory bits should be hex (it's a coincidence that it looks like a round decimal number)

src/d/actor/d_a_tornado.cpp
```diff
@@ -4,72 +4,342 @@
...
+            mpModelUnder->setBaseScale(under_small_scale);
+        } else if (param == 2) {
+            mpModel->setBaseScale(small_scale);
+            mpModelUnder->setBaseScale(under_small_scale);
+            for (int i=0; i<11; i++) {
+                mJointScale[i] = l_joint_scale[i];
+            }
+        } else {
+            fopAcM_SetParam(this, 0);
+            if (dComIfGs_isEventBit(10000)) {
```
LagoLunatic, line 319:
```suggestion
            if (dComIfGs_isEventBit(0x2710)) {
```

PR #698, opened 2025-03-02
d_a_obj_shelf equivalent

ross3102:
* in a few cases I couldn't get a 100% match without doing some weird casting - I imagine there's a better way


src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
+            /* 0x08 */ short mVibXSpeed;
+            /* 0x0a */ short mVibZSpeed;
+            /* 0x0c */ float mVibXMagnitude;
+            /* 0x10 */ float mVibZMagnitude;
+            /* 0x14 */ float mRotAccel;
+            /* 0x18 */ float mRotDecay;
+            /* 0x1c */ float mBounceFactor;
+            /* 0x20 */ u8 mBounceNum;
+            /* 0x21 */ u8 mRotWaitDuration;
+            /* 0x22 */ short m22;
```
ross3102, line 32: this is unused?

LagoLunatic: It's probably just alignment padding. mRotWaitDuration ends at offset 0x22, but mInitSpeed2 is a float, which has alignment 4. So the compiler places mInitSpeed2 at offset 0x24 for alignment, and these two bytes in here are just padding. You can remove this line.

src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
+        static const Attr_c L_attr = {
+            4,
+            30000, 4.0f,
+            0x7724,0x7148,
+            200.0f,200.0f,
+            600.0f,0.02f,
+            0.5f,
+            5,0,
+            0,
+            -3500.0f,-2500.0f
```
LagoLunatic, line 46: Since you already named all the fields in the Attr_c struct anyway, you could make the L_attr definition more readable by adding comments on the left hand side with the name of each field, e.g. `/* mVibDuration */ 4,` instead of `4,`. (e.g. like we have `l_cyl_src` fields commented throughout the repo)

src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
-    /* Nonmatching */
+void Mthd_Draw(void* i_this) {
+    static_cast<Act_c*>(i_this)->Draw();
 }
 
 /* 00000ADC-00000B08       .text Mthd_IsDelete__Q210daObjShelf27@unnamed@d_a_obj_shelf_cpp@FPv */
-void Mthd_IsDelete(void*) {
-    /* Nonmatching */
+void Mthd_IsDelete(void* i_this) {
+    static_cast<Act_c*>(i_this)->IsDelete();
```
LagoLunatic, line 288:
```suggestion
BOOL Mthd_IsDelete(void* i_this) {
    return static_cast<Act_c*>(i_this)->IsDelete();
```

src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
-    /* Nonmatching */
+int Mthd_Execute(void* i_this) {
+    return static_cast<Act_c*>(i_this)->MoveBGExecute();
 }
 
 /* 00000AB0-00000ADC       .text Mthd_Draw__Q210daObjShelf27@unnamed@d_a_obj_shelf_cpp@FPv */
-void Mthd_Draw(void*) {
-    /* Nonmatching */
+void Mthd_Draw(void* i_this) {
+    static_cast<Act_c*>(i_this)->Draw();
```
LagoLunatic, line 283:
```suggestion
BOOL Mthd_Draw(void* i_this) {
    return static_cast<Act_c*>(i_this)->Draw();
```

src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
+                mode_wait_init();
+            } else {
+                mode_fell_init();
+            }
+        } else {
+            // Bounce
+            shape_angle.x = mTargetAngle - (int)((short)(int)(shape_angle.x - mTargetAngle) * attr().mBounceFactor);
+            mRotSpeed = mRotSpeed * -attr().mBounceFactor;
+            if (mCurBounce == attr().mBounceNum) {
+                fopAcM_seStart(this, 0x696d, 0);
```
LagoLunatic, line 204:
```suggestion
                fopAcM_seStart(this, JA_SE_OBJ_RACK_BREAK, 0);
```
JAZelAudio_SE.h has an enum of all the official sound effect names

src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
+        phase_state = MoveBGCreate(M_arcname, 7, dBgS_MoveBGProc_Trans, 0xb00);
+        JUT_ASSERT(0x15b, (phase_state == cPhs_COMPLEATE_e) || (phase_state == cPhs_ERROR_e));
+    }
+    return phase_state;
 }
 
 /* 0000029C-000002A4       .text Delete__Q210daObjShelf5Act_cFv */
 BOOL daObjShelf::Act_c::Delete() {
-    /* Nonmatching */
+    return 1;
```
LagoLunatic, line 86:
```suggestion
    return TRUE;
```

src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
 
 /* 000001A0-0000029C       .text Mthd_Create__Q210daObjShelf5Act_cFv */
-void daObjShelf::Act_c::Mthd_Create() {
-    /* Nonmatching */
+s32 daObjShelf::Act_c::Mthd_Create() {
+    fopAcM_SetupActor(this, Act_c);
+
+    s32 phase_state = dComIfG_resLoad(&mPhs, M_arcname);
+    if (phase_state == cPhs_COMPLEATE_e) {
+        phase_state = MoveBGCreate(M_arcname, 7, dBgS_MoveBGProc_Trans, 0xb00);
```
LagoLunatic, line 78:
```suggestion
        phase_state = MoveBGCreate(M_arcname, OTANA_DZB_OTANA, dBgS_MoveBGProc_Trans, 0xb00);
```
This argument is the file index of the collision

src/d/actor/d_a_obj_shelf.cpp
```diff
@@ -4,133 +4,288 @@
...
+}
+
+const char daObjShelf::Act_c::M_arcname[] = "Otana";
 
 /* 00000078-0000012C       .text CreateHeap__Q210daObjShelf5Act_cFv */
-void daObjShelf::Act_c::CreateHeap() {
-    /* Nonmatching */
+int daObjShelf::Act_c::CreateHeap() {
+    J3DModelData* model_data = (J3DModelData *)dComIfG_getObjectRes(M_arcname, OTANA_BDL_OTANA);
+    JUT_ASSERT(0x12c, model_data != 0);
```
LagoLunatic, line 58:
```suggestion
    JUT_ASSERT(0x12c, model_data != NULL);
```
It appears as "0" in the string, but the programmers likely wrote NULL and the compiler converted it like that due to nested macros.

PR #699, opened 2025-03-06
d_a_tag_kf1

njulrik:
`Equivalent`, appears to only mismatch on data offset.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1779,7 +1779,7 @@ config.libs = [
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_swtact"),
     ActorRel(Matching,    "d_a_tag_etc"),
     ActorRel(Matching,    "d_a_tag_island"),
-    ActorRel(NonMatching, "d_a_tag_kf1"),
+    ActorRel(Equivalent, "d_a_tag_kf1"),
     ActorRel(Matching,    "d_a_tag_ret"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_tag_volcano"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_title"),


src/d/actor/d_a_tag_kf1.cpp
```diff
@@ -5,174 +5,413 @@
...
 /* 000010C0-000010E0       .text daTag_Kf1_Create__FP10fopAc_ac_c */
-static cPhs_State daTag_Kf1_Create(fopAc_ac_c* i_this) {
-    return ((daTag_Kf1_c*)i_this)->_create();
+static cPhs_State daTag_Kf1_Create(fopAc_ac_c* obj) {
+    (static_cast<daTag_Kf1_c*>(obj))->_create();
 }
 
 /* 000010E0-00001100       .text daTag_Kf1_Delete__FP11daTag_Kf1_c */
-static BOOL daTag_Kf1_Delete(daTag_Kf1_c* i_this) {
+static bool daTag_Kf1_Delete(daTag_Kf1_c* i_this) {
```
LagoLunatic: These static funcs should always return `BOOL` not `bool`. You can get the clrlwi to appear/disappear by changing the return types of the *inner* functions (e.g. `_delete`) instead.

src/d/actor/d_a_tag_kf1.cpp
```diff
@@ -5,174 +5,413 @@
...
 /* 000000EC-00000120       .text __ct__15daTag_Kf1_HIO_cFv */
-daTag_Kf1_HIO_c::daTag_Kf1_HIO_c() {
-    /* Nonmatching */
+daTag_Kf1_HIO_c::daTag_Kf1_HIO_c()
+{ 
+    mAttentionMaxEuclidDistance = a_prm_tbl[0];
+    mAttentionMaxYDistance = a_prm_tbl[1];
+    // this line should def load from a_prm_tbl according to asm,
+    // but unlike the others the type is not float so do this gross cast instead
+    f0x10 = *(u8*)(void*)&a_prm_tbl[2];
```
LagoLunatic: `a_prm_tbl` needs to be a struct, not an array of floats. You can define a custom struct here to match it.

Also, the name being `a_prm_tbl$4033` tells you to put the static var inside of the function - `$` means an in-function static.

src/d/actor/d_a_tag_kf1.cpp
```diff
@@ -5,174 +5,413 @@
...
 }
 
 /* 00000120-000001B0       .text searchActor_Kutani__FPvPv */
-void searchActor_Kutani(void*, void*) {
-    /* Nonmatching */
+void* searchActor_Kutani(void* arg1, void* _) {
+    fopAc_ac_c* act = (fopAc_ac_c*)arg1;
+    if (l_check_wrk < 100 && fopAc_IsActor(act) && act->base.base.mProcName == 0x1cb &&
+        daObj::PrmAbstract<daTsubo::Act_c::Prm_e>(act, daTsubo::Act_c::PRM_TYPE_W,
+                                                  daTsubo::Act_c::PRM_TYPE_S) == 0xe) {
```
LagoLunatic: Use daTsubo::Act_c's prm_chk_type_kutani inline

src/d/actor/d_a_tag_kf1.cpp
```diff
@@ -5,174 +5,413 @@
...
+    return TRUE;
 }
 
+static char* a_demo_name_tbl[] = {"BENSYO"};
+
 /* 00000220-00000234       .text setStt__11daTag_Kf1_cFSc */
-void daTag_Kf1_c::setStt(signed char) {
+char daTag_Kf1_c::setStt(signed char c) {
+    stt = c;
+    if (stt == 3) {
```
LagoLunatic: You can match the weird cmpwi here by doing a switch statement on `stt` that doesn't do anything.

For future reference, if you're having trouble with a function like this, you can ask for help in the discord by posting a scratch: https://github.com/zeldaret/tww/blob/main/docs/decompiling.md#asking-for-help-with-a-function-by-sharing-a-decompme-scratch

PR #700, opened 2025-03-08
d-a-obj-barrel

ross3102:


src/d/actor/d_a_obj_barrel.cpp
```diff
@@ -4,226 +4,758 @@
...
+        /* SrcGObjAt Se      */ 0,
+        /* SrcGObjAt HitMark */ 0,
+        /* SrcGObjAt Spl     */ 0,
+        /* SrcGObjAt Mtrl    */ 0,
+        /* SrcGObjAt SPrm    */ 0,
+        /* SrcGObjTg Se      */ 0,
+        /* SrcGObjTg HitMark */ 0,
+        /* SrcGObjTg Spl     */ 0,
+        /* SrcGObjTg Mtrl    */ 0,
+        /* SrcGObjTg SPrm    */ 1,
```
LagoLunatic, line 49:
```suggestion
        /* SrcGObjTg SPrm    */ dCcG_TgSPrm_Shield_e,
```

src/d/actor/d_a_obj_barrel.cpp
```diff
@@ -4,226 +4,758 @@
...
+
+#include "weak_data_1811.h" // IWYU pragma: keep
+
+const char daObjBarrel::Act_c::M_arcname[] = "Ktaru_01";
+const float daObjBarrel::Act_c::l_s_radius = 45.0f;
+const float daObjBarrel::Act_c::l_l_radius = 50.0f;
+const float daObjBarrel::Act_c::l_gnd_fric = 0.1f; // TODO use
+const short daObjBarrel::Act_c::l_gnd_deg = 0xf; // TODO ??
+const float daObjBarrel::Act_c::l_viscous_resist = 0.006f;
+const float daObjBarrel::Act_c::l_inert_resist = 0.001f; // TODO use more?
```
LagoLunatic, line 22: Old comments?

src/d/actor/d_a_obj_barrel.cpp
```diff
@@ -4,226 +4,758 @@
...
-    /* Nonmatching */
+u32 daObjBarrel::Act_c::get_se_map_hit() const {
+    return 0x0B;
 }
 
 /* 00001940-00001994       .text set_senv__Q211daObjBarrel5Act_cCFii */
-void daObjBarrel::Act_c::set_senv(int, int) const {
-    /* Nonmatching */
+void daObjBarrel::Act_c::set_senv(int p1, int p2) const {
+    // TODO:  create cXyz?
```
LagoLunatic, line 511: Old comment?

src/d/actor/d_a_obj_barrel.cpp
```diff
@@ -4,226 +4,758 @@
...
+        case MODE_CARRY:
+        case MODE_JUMP:
+        case MODE_WALK:
+            mDoMtx_stack_c::transM(0.0f, cM_scos(shape_angle.z) * 5.0f + l_s_radius, 0.0f);
+            mDoMtx_stack_c::ZXYrotM(0, shape_angle.y, shape_angle.z);
+            mDoMtx_stack_c::YrotM(m612);
+            mDoMtx_stack_c::XrotM(m610);
+            mDoMtx_stack_c::YrotM(-m612);
+            mDoMtx_stack_c::YrotM(m630);
+            mDoMtx_stack_c::transM(0.0f, -l_l_radius, 0.0f);
```
LagoLunatic, line 432:
```suggestion
            mDoMtx_stack_c::transM(0.0f, -l_l_radius, 0.0f);
            break;
```

src/d/actor/d_a_obj_barrel.cpp
ross3102: good catch, I'll remove these
I will call out that I'm not sure if I've over- or under-used these constants since the generated assembly doesn't change whether I use them or not. I've given it my best guess (although `l_gnd_deg` is still unused, maybe one of the inlines uses it)

src/d/actor/d_a_obj_barrel.cpp
ross3102: I noticed a call to `__ct__4cXyzFRC4cXyz` in the debug map for this function before the `fopAcM_GetID`, but not sure what it's for. I'll remove the comment though

LagoLunatic: It's because dKy_Sound_set takes a position passed by value and not reference, so when you pass `current.pos` the compiler automatically creates a new cXyz on the stack using the copy constructor `cXyz::cXyz(const cXyz&)`, and then passes a pointer to the temp on the stack to the function. This gets inlined but if you want to see what it would have looked like for reference, throw `#pragma dont_inline on` above this function and you should see the call.

PR #703, opened 2025-03-12
Added decompilation guides for new contributors

LagoLunatic:
I started writing some guides on the whole decompilation process that cover everything I could think of:

* Ghidra
* objdiff
* Reading the debug maps
* Naming styles used by the devs
* Regswaps
* Weak func order (unfinished)

View here: https://github.com/LagoLunatic/tww/blob/docs/docs/decompiling.md


magcius:
Great work! I have a couple more thoughts for tips (switch statements, cXyz), but this is fantastic already.

PR #704, opened 2025-03-12
d-a-obj-swheavy

ross3102:


include/d/actor/d_a_obj_swheavy.h
```diff
@@ -1,27 +1,53 @@
...
     public:
-        void attr() const {}
-        void is_switch() const {}
-        void off_switch() const {}
-        void on_switch() const {}
-        void prm_get_swSave() const {}
-        void prm_get_type() const {}
-        void rev_switch() const {}
+        struct Attr_c {
+            u8 m00[0x4];
```
ross3102, line 11: it seems like this isn't used anywhere?

include/d/actor/d_a_obj_swheavy.h
```diff
@@ -1,27 +1,53 @@
...
+            short mPushDelay;
+        };
+        const Attr_c& attr() const { return M_attr[mType]; }
+
+        enum Flag {
+            FLAG_OBEY_SAVE  = 0x1,
+            FLAG_STAY_PRESSED = 0x2,
+            FLAG_IS_TOGGLE  = 0x4,
+            FLAG_REQ_HEAVY  = 0x8,
+        };
```
ross3102, line 27: is it ok/good practice to make enums for stuff like this? are there naming conventions?

src/d/actor/d_a_obj_swheavy.cpp
```diff
@@ -4,133 +4,457 @@
...
+        4,
+        15,
+    },
+};
 
 /* 00000078-0000009C       .text solidHeapCB__Q212daObjSwheavy5Act_cFP10fopAc_ac_c */
-void daObjSwheavy::Act_c::solidHeapCB(fopAc_ac_c*) {
-    /* Nonmatching */
+BOOL daObjSwheavy::Act_c::solidHeapCB(fopAc_ac_c* i_this) {
+    return static_cast<daObjSwheavy::Act_c*>(i_this)->create_heap() & 0xFF; // Fakematch?
```
ross3102, line 60: I tried messing with return types to get this to match without the & 0xFF - found a similar thing in d_a_obj_akabe

PR #706, opened 2025-03-16
d_a_ship 95%

mattias-blum:
FYI to everyone that I'm working on this


src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4718 @@
...
+            if (fVar17 < 0.01f) {
+                currentRopeSegment->set(*(currentRopeSegment + 1) + l_rope_base_vec);
+            }
+            else {
+                currentRopeSegment->set(*(currentRopeSegment + 1) + (local_ec * 10.0f)/fVar17);
+            }
+
+            cStack_160 = (*currentRopeSegment - local_f8) * 0.05f; // This isn't meant to be storing this result in a local but couldn't figure out how to avoid it given the * 0.05f operation
+
+            PSVECAdd(ropeSegments, &cStack_160, ropeSegments);
```
LagoLunatic, line 3221:
```suggestion
            *ropeSegments += cStack_160;
```
If you ever see a `PSVEC` function being called, it's almost certainly an inline of the cXyz class

Applies in a lot of other places here

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4718 @@
...
+        }
+
+        mDoMtx_stack_c::transS(this->current.pos.x, this->current.pos.y, this->current.pos.z);
+        mDoMtx_ZXYrotM(mDoMtx_stack_c::get(), this->shape_angle.x, this->shape_angle.y, this->shape_angle.z);
+        
+        this->m0298->getModel()->setBaseTRMtx(mDoMtx_stack_c::get());
+        this->m0298->play(NULL, 0, 0);
+        this->m0298->calc();
+        
+        PSMTXMultVec(this->m0298->getModel()->getAnmMtx(10), &l_tiller_top_offset, &this->m0438);
```
LagoLunatic, line 4697:
```suggestion
        cMtx_multVec(this->m0298->getModel()->getAnmMtx(10), &l_tiller_top_offset, &this->m0438);
```
If you see a `PSMTX` function being called, it's usually an inline in m_Do_mtx.h

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4718 @@
...
+        this->m029C->calc();
+        
+        g_dComIfG_gameInfo.play.mpPlayerPtr[2] = this;
+        
+        this->mTrack.mMinVel = 3.0f;
+        this->mWaveL.mMaxParticleVelocity = 40.0f;
+        this->mWaveR.mMaxParticleVelocity = 40.0f;
+        this->m19C0.mRate = 0.0f;
+
+        if (dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == 7) {
```
LagoLunatic, line 4710:
```suggestion
        if (dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == dStageType_SEA_e) {
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4718 @@
...
+        this->mNextMessageID = 0xd67;
+    }
+    else if (daPy_getPlayerActorClass()->checkMasterSwordEquip()) {
+        this->mNextMessageID = 0xd66;
+    }
+    else if (dComIfGs_isEventBit(0x2D10)) {
+        this->mNextMessageID = 0xd65;
+    }
+    else if (dComIfGs_isEventBit(0x1E40)) {
+        if (dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == 1) {
```
LagoLunatic, line 439:
```suggestion
        if (dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == dStageType_DUNGEON_e) {
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4718 @@
...
-void daShip_c::checkOutRange() {
-    /* Nonmatching */
+BOOL daShip_c::checkOutRange() {
+    if (this->m034B == 0xFF) {
+        return FALSE;
+    }
+
+    int pathIndex = 0;
+    BOOL bVar4 = FALSE;
+    BOOL bVar5 = dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == 7;
```
LagoLunatic, line 947:
```suggestion
    BOOL bVar5 = dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == dStageType_SEA_e;
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4718 @@
...
+            }
+            else if (this->m034D == 16) {
+                procStartModeThrow_init();
+            }
+            else {
+                procWait_init();
+            }
+        }
+
+        if ((dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == 7) || !strcmp(dComIfGp_getStartStageName(), "Hyrule") || !strcmp(dComIfGp_getStartStageName(), "Ocean")) {
```
LagoLunatic, line 4625:
```suggestion
        if ((dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == dStageType_SEA_e) || !strcmp(dComIfGp_getStartStageName(), "Hyrule") || !strcmp(dComIfGp_getStartStageName(), "Ocean")) {
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4718 @@
...
+            this->m037A = 1;
+            if (pfVar2 != NULL) {
+                pfVar2->m31c = 1;
+            }
+            dKy_get_seacolor(&diff, &amb);
+            this->m03BC.x = 0;
+            this->m03BC.y = this->shape_angle.y + 0x8000;
+            this->m03BC.z = 0;
+            dComIfGp_particle_set(0x285, &this->current.pos, &this->shape_angle, NULL, 0xFF, &this->m1984, -1, &diff);
+            dComIfGp_particle_set(0x285, &this->current.pos, &this->m03BC, NULL, 0xFF, &this->m1998, -1, &diff);
```
LagoLunatic, line 2717:
```suggestion
            dComIfGp_particle_set(dPa_name::ID_COMMON_0285, &this->current.pos, &this->shape_angle, NULL, 0xFF, &this->m1984, -1, &diff);
            dComIfGp_particle_set(dPa_name::ID_COMMON_0285, &this->current.pos, &this->m03BC, NULL, 0xFF, &this->m1998, -1, &diff);
```

Particle IDs have an enum in d_particle_name.h

LagoLunatic (commented):
Partial review, didn't go over everything yet

src/d/actor/d_a_ship.cpp
LagoLunatic: That's going to be one of the abs inlines. abs2, abs, absXZ, abs2XZ, depending on what the code is doing

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+        
+        g_dComIfG_gameInfo.play.mpPlayerPtr[2] = this;
+        
+        mTrack.mMinVel = 3.0f;
+        mWaveL.mMaxParticleVelocity = 40.0f;
+        mWaveR.mMaxParticleVelocity = 40.0f;
+        m19C0.mRate = 0.0f;
+
+        if (dStage_stagInfo_GetSTType(dComIfGp_getStageStagInfo()) == dStageType_SEA_e) {
+            fopAcM_createChild(0x110, fopAcM_GetID(this), 0, &current.pos, -1, NULL);
```
LagoLunatic, line 4706:
```suggestion
            fopAcM_createChild(PROC_Coming2, fopAcM_GetID(this), 0, &current.pos, -1, NULL);
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+        }
+    }
+
+    mCurrentRopeSegmentIndex = targetRopeSegmentIndex;
+    return;
 }
 
 /* 0000882C-00009314       .text setRopePos__8daShip_cFv */
 void daShip_c::setRopePos() {
-    /* Nonmatching */
```
LagoLunatic, line 396: Functions that aren't 100% matched should have the nonmatching comment left in

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+            sVar2 = 1;
+        }
+        else if (sVar1 < 0) {
+            sVar2 = -1;
+        }
+    }
+    m036E += sVar2;
+    m036C += m036E;
+    cLib_addCalcAngleS(&m036E, 0, 0x14, 0x1000, 4);
+    return;
```
LagoLunatic, line 764: `return;` at the end of all these void functions isn't needed, only in the middle of a function for an early return

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+        if (tgHitObj && (tgHitObj->GetAtType() - AT_TYPE_WIND) == 0) {
+            if (dComIfGp_checkPlayerStatus0(0, daPyStts0_SHIP_RIDE_e)) {
+                onStateFlg((daSHIP_SFLG)(daSFLG_UNK1_e | daSFLG_UNK2000000_e));
+            }
+        }
+        else {
+            if (tgRVecP) {
+                short sVar2;
+                local_14.set(tgRVecP->x, 0.0f, tgRVecP->z);
+                if (local_14.abs2() < 0.1f) {
```
LagoLunatic, line 1241: I see several places where abs() or abs2() are called that have the x, 0, z pattern before them so the XZ versions of the inlines should work here I think:
```suggestion
                if (tgRVecP->abs2XZ() < 0.1f) {
```
btw this might help improve the stack offsets you mentioned before

include/d/actor/d_a_ship.h
```diff
@@ -38,7 +80,7 @@ class daShip_c : public fopAc_ac_c {
     void checkTornadoUp() const {}
     void getBeltSpeed() const {}
     void getBodyMtx() {}
-    void getCannonAngleX() const {}
+    s16 getCannonAngleX() const { return this->shape_angle.x + this->m0396 - 0x4000; }
```
LagoLunatic, line 83:
```suggestion
    s16 getCannonAngleX() const { return shape_angle.x + m0396 - 0x4000; }
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+        // cM3dGSphS
+        {
+            /* Center */ 0.0f, 0.0f, 0.0f,
+            /* Radius */ 75.0f,
+        },
+    };
+    
+    fopAcM_SetupActor(this, daShip_c);
+    
+    if (dComIfGs_isEventBit(0xF80) == 0) {
```
LagoLunatic, line 4517: A lot of these `func() == 0` and `func() != 0` in here can be written without the 0 if the return type is a boolean, e.g.
```suggestion
    if (!dComIfGs_isEventBit(0xF80)) {
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+                speedF = fVar3;
+                if (mTornadoActor) {
+                    fVar4 = m0404 * 40.0f + 30.0f;
+                    if (!dComIfGp_checkPlayerStatus0(0,0x1000) || !dComIfGp_checkCameraAttentionStatus(0,0x20)) {
+                        fVar4 *= 1.2f;
+                    }
+                }
+                else {
+                    fVar4 = m0404 * 30.0f + 10.0f;
+                    if (!dComIfGs_getBombNum() && fopAcM_GetRoomNo(this) == 0x2C) {
```
LagoLunatic, line 3789: And the opposite, this doesn't return a boolean
```suggestion
                    if (dComIfGs_getBombNum() == 0 && fopAcM_GetRoomNo(this) == 0x2C) {
```
Generally if a function has "is" in its name it probably returns a bool, if it has "num" it probably returns a count

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+        }
+    }
+    return;
 }
 
 /* 00001734-00001BD0       .text setNextMessage__8daShip_cFP9msg_class */
-void daShip_c::setNextMessage(msg_class*) {
-    /* Nonmatching */
+BOOL daShip_c::setNextMessage(msg_class* msg) {
+    uint currMessageID;
```
LagoLunatic, line 568:
```suggestion
    u32 currMessageNo;
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+         (((currMessageID != 0x5eb && (currMessageID != 0x5ee)) &&
+          ((currMessageID != 0x5ef && ((currMessageID != 0x5f5 && (currMessageID != 0x5f9)))))))) &&
+        ((currMessageID != 0x608 &&
+         ((((((currMessageID != 0x609 && (currMessageID != 0x60a)) && (currMessageID != 0x621)) &&
+            ((currMessageID != 0x623 && (currMessageID != 0x636)))) &&
+           ((currMessageID < 0xd5c || (currMessageID > 0xd78)))) &&
+          ((currMessageID != 0xd7b && (currMessageID != 0x164c)))))))) &&
+       ((currMessageID != 0x1650 &&
+        ((((currMessageID != 0x1651 && (currMessageID != 0x1684)) && (currMessageID != 0x1685)) &&
+         ((currMessageID != 0x1687 && (currMessageID != 0x1689)))))))) {
```
LagoLunatic, line 581: Most of these parentheses seem unnecessary, I think you can remove all of them except the pair around the OR condition

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
-    /* Nonmatching */
+BOOL daShip_c::procTalk() {
+    msg_class *msg;
+    u16 sVar4;
+    
+    firstDecrementShipSpeed(0.0f);
+    setMoveAngle(0);
+    if (m0430 == 0xFFFFFFFF) {
+        if (dComIfGp_checkCameraAttentionStatus(dComIfGp_getPlayerCameraID(0), 4)) {
+            m0430 = fopMsgM_messageSet(mNextMessageNo, &eyePos);
```
LagoLunatic, line 2471: fopMsgM_messageSet returns a fpc_ProcID, so m0430 can probably be changed to that type, and then also use the invalid ID enum:
```suggestion
    if (m0430 == fpcM_ERROR_PROCESS_ID_e) {
        if (dComIfGp_checkCameraAttentionStatus(dComIfGp_getPlayerCameraID(0), 4)) {
            m0430 = fopMsgM_messageSet(mNextMessageNo, &eyePos);
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+                int bombId;
+                if (this_00) {
+                    if (this_00) {
+                        bombId = ((base_process_class*)this_00)->mBsPcId;
+                    }
+                    else {
+                        bombId = 0xffffffff;
+                    }
+
+                    dCam_getBody()->ForceLockOn(bombId);
```
LagoLunatic, line 4034: Inline
```suggestion
                    dCam_getBody()->ForceLockOn(fopAcM_GetID(bomb));
```
Also, `this_00` is a bizarre Ghidra-ism, idk why it does that. Just name this bomb since we know its class already

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+        pModel = m029C->getModel();
+        pModelData = pModel->getModelData();
+        
+        pModel->setUserArea(reinterpret_cast<u32>(this));
+        
+        for (u16 jno = 0; jno < pModelData->getJointNum(); jno++) {
+            if ((jno == 8) || (jno == 10)) {
+                pModelData->getJointTree().getJointNodePointer(jno)->setCallBack(daShip_headJointCallBack0);
+            }
+            else if (!((jno != 2) && (jno != 3) && (jno != 4) && (jno != 5) && (jno != 6)) || (jno == 7)) {
```
LagoLunatic, line 4575: When there's a bunch of double negatives it probably means the whole condition can be inverted, I think it would be something like this
```suggestion
            else if (jno == 2 || jno == 3 || jno == 4 || jno == 5 || jno == 6 || jno == 7) {
```

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
+            }
+            
+            mpCannonModel->calc();
+            
+            if (m037A == 30) {
+                cMtx_multVec(mpCannonModel->getAnmMtx(2), &l_cannon_top, &cStack_bc);
+
+                local_184.set(getCannonAngleX(), shape_angle.y + m0394, shape_angle.z);
+
+                daBomb_c* this_00 = (daBomb_c *)fopAcM_fastCreate(0x128, daBomb_c::prm_make(daBomb_c::STATE_4, FALSE, TRUE), &cStack_bc, tevStr.mRoomNo, &local_184);
```
LagoLunatic, line 4023:
```suggestion
                daBomb_c* bomb = (daBomb_c *)fopAcM_fastCreate(PROC_BOMB, daBomb_c::prm_make(daBomb_c::STATE_4, FALSE, TRUE), &cStack_bc, tevStr.mRoomNo, &local_184);
```

src/d/actor/d_a_ship.cpp
mattias-blum: Yeah I'm aware of this inline but the asm expects the above code not the inline, it's odd but perhaps the dev didn't use it for whatever reason?

LagoLunatic: That seems unlikely in this particular case since there's a doubled up `if (this_00) {` check that a programmer wouldn't write out twice manually.

If it doesn't match leave a fakematch comment mentioning briefly what issue is caused by the inline

src/d/actor/d_a_ship.cpp
```diff
@@ -5,460 +5,4713 @@
...
-    /* Nonmatching */
+BOOL daShip_c::procPaddleMove() {  
+    if (checkNextMode(2)) {
+        return TRUE;
+    }
+    if (m0298->getFrame() >= 7.0f) {
+        mShipMode = Idle;
+        m03E8 = 1.0f;
+    }
+    if (dComIfGp_checkPlayerStatus0(0, 0x685000) || dComIfGp_checkPlayerStatus1(0, 8)) {
```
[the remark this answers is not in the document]
mattias-blum: Is there an enum for calls to `dComIfGp_checkCameraAttentionStatus`?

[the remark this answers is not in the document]
LagoLunatic: Not that I can see, probably because nobody's started work on the camera system yet

src/d/actor/d_a_ship.cpp
mattias-blum: I think the issue is because it's in that large execute function, the compiler is doing that thing where it "gives up" and instead of generating the asm it instead just puts the function name
Screenshot 2025-03-28 at 12 38 43 AM

src/d/actor/d_a_ship.cpp
mattias-blum: Yeah I tried inverting this condition but the asm generated is different. No idea why but I even noticed that when using certain inline functions in conditionals, using `&&` and `||` logical operators would change the asm that was generated. I'll leave a comment mentioning it.

src/d/actor/d_a_ship.cpp
LagoLunatic: Oh so it's in the same function as the dComIfGp_demo_getActor problem, but it's doing the opposite thing? That's really weird, the number of functions it inlines being wrong could be related to other issues in this function but I don't really get how it would be inlining both too many and too few at the same time, something's up with these inlines

src/d/actor/d_a_ship.cpp
LagoLunatic: The type of the variable can also impact how the compiler optimizes these chained comparisons btw, but the u16 you have would normally already be correct for a joint no, so not sure what the issue could be here

mattias-blum:
`checkNextMode` and `procZevDemo` now at 100% matching after adding `absXZ` inline

PR #707, opened 2025-03-18
d-a-obj-swpush

ross3102:


src/d/actor/d_a_obj_swpush.cpp
```diff
@@ -5,212 +5,802 @@
...
+        /* mFlags           */ static_cast<AttrFlag_e>(FLAG_REQ_HEAVY | FLAG_STAY_PRESSED | FLAG_OBEY_SAVE),
+        /* mScale           */ 1.5f,
+        /* mKbotaResName    */ daObjSwpush::Act_c::M_arcname_kbota,
+        /* mHhbotResName    */ daObjSwpush::Act_c::M_arcname_hhbot,
+        /* mBgArcName       */ daObjSwpush::Act_c::M_arcname_kbota,
+        /* mModelArcName    */ daObjSwpush::Act_c::M_arcname_hhbot,
+        /* mBtpArcName      */ NULL,
+        /* mBgResIndex      */ 7,
+        /* mModelResIndices */ { 4, 4 },
+        /* mBtpResIndex     */ 0xFFFF,
```
LagoLunatic, line 102:
```suggestion
        /* mBgResIndex      */ KBOTA_00_DZB_KBOTA_00,
        /* mModelResIndices */ { HHBOT_BDL_HHBOT1, HHBOT_BDL_HHBOT1 },
        /* mBtpResIndex     */ -1,
```
You can include a couple of the headers in the d/res folder to get enum names for all the file indexes in these structs.
Also, mBtpResIndex is signed, so -1 makes more sense than 0xFFFF here.

src/d/actor/d_a_obj_swpush.cpp
```diff
@@ -5,212 +5,802 @@
...
-    /* Nonmatching */
+    if (dComIfGp_evmng_endCheck(mEventID)) {
+        dComIfGp_event_reset();
+        demo_non_init();
+    }
 }
 
 /* 00001C84-00001F14       .text Mthd_Execute__Q211daObjSwpush5Act_cFv */
 BOOL daObjSwpush::Act_c::Mthd_Execute() {
     /* Nonmatching */
```
LagoLunatic, line 720: Old comment

src/d/actor/d_a_obj_swpush.cpp
```diff
@@ -5,212 +5,802 @@
...
 void daObjSwpush::Act_c::demo_non() {
-    /* Nonmatching */
 }
 
 /* 000019A0-000019F8       .text demo_reqPause_init__Q211daObjSwpush5Act_cFv */
 void daObjSwpush::Act_c::demo_reqPause_init() {
-    /* Nonmatching */
+    if (mDemoMode == DEMO_MODE_NON) {
+        mDemoMode = DEMO_MODE_REQ_PAUSE;
+        fopAcM_orderPotentialEvent(this, 2, 0, 0);
```
LagoLunatic, line 640:
```suggestion
        fopAcM_orderPotentialEvent(this, dEvtFlag_STAFF_ALL_e, 0, 0);
```

include/d/actor/d_a_obj_swpush.h
```diff
@@ -1,37 +1,105 @@
...
+        /* 0x2C */ float mSpeedDecay;
+        /* 0x30 */ float mPushSpeed0;
+        /* 0x34 */ float m34;
+        /* 0x38 */ short m38;
+        /* 0x3A */ short mMiniPushDelay1;
+        /* 0x3C */ short mPushDelay;
+        /* 0x3E */ short mMiniPushDelay2;
+        /* 0x40 */ float m40;
+        /* 0x44 */ float m44;
+        /* 0x48 */ short mPauseDuration;
```
LagoLunatic, line 43: We should probably use the `s16`/`f32`/etc dolphin typedefs instead of the full type names for primitives. Assertion strings throughout the game indicate the devs used those pretty consistently

ross3102: does `int` have an equivalent typedef like that? I tried changing an `int` parameter to `s32` (specifically the second argument to jnodeCB) and got an error:
`Cannot initialize a parameter of type 'J3DNodeCallBack' (aka 'int (*)(J3DNode *, int)') with an lvalue of type 'BOOL (J3DNode *, s32)' (aka 'int (J3DNode *, long)'): type mismatch at 2nd parameter ('int' vs 's32' (aka 'long'))`

LagoLunatic: No, don't convert `int` to `s32`, leave it as is. Even though `int` and `s32` are both signed 32 bit integers, the compiler treats them differently in a bunch of ways (not just errors, it sometimes produces different code that doesn't match if you use the wrong one).

PR #711, opened 2025-03-28
d_a_player_swim near complete

mattias-blum:
FYI to everyone that I'm working on this. I will also attempt to parametrize the swimming behaviour in a separate workflow for usage in speedrunning.


mattias-blum:
Mostly complete. Only 2 functions have non-matching code (others are just data offsets). Summary of outstanding issues:
- `changeSwimOutProc`: The call to `onNoResetFlg1` towards the end of the function is generating an `oris` instruction rather than `ori` but adding `>> 16` to the flag operand in the inline function causes other instances of that function call to no longer match. Not sure how to resolve that.
- `setSwimMoveAnime`: `currentFrame` needs to be loaded at the start of the function and used somewhere in the second param of the call to `setSingleMoveAnime`. When doing so it just ends up generating a bunch of extra asm that I don't want. Not sure how to handle that.

Both should be quick fixes though hopefully.

src/d/actor/d_a_player_swim.inc
```diff
@@ -11,85 +11,613 @@
...
+
+    gravity = 0.0f;
+    setSingleMoveAnime(ANM_SWIMWAIT, (float)(getSwimTimerRate() * l_HIO_swim_c0.m.field_0x70) + l_HIO_swim_c0.m.field_0x40, 0.0f, -1, l_HIO_swim_c0.m.field_0x44);
+    mFrameCtrl->setFrame(fVar2 * mFrameCtrl->getEnd());
+    mAnmRatioUnder[0].getAnmTransform()->setFrame(mFrameCtrl->getFrame());
+    dComIfGp_setPlayerStatus0(0, daPyStts0_SWIM_e);
+
+    if (checkNoResetFlg0(daPyFlg0_UNK100)) {
+        if (mEquipItem != 0x100) {
+            deleteEquipItem(1);
```
LagoLunatic, line 434:
```suggestion
            deleteEquipItem(TRUE);
```

src/d/actor/d_a_player_swim.inc
```diff
@@ -11,85 +11,613 @@
...
+                        m3570 = 1;
+                    }
+                }
+            }
+        }
+    }
+    else {
+        dComIfGp_setItemTimeCount(900);
+        dComIfGp_setItemTimeMax(1);
+        dComIfGp_setItemTimeMax(0); //Fakematch? Seems nonsensical to set it to 1 then immediately to 0, also could be `dComIfGp_stopItemSwimTimer` and `dComIfGp_startItemSwimTimer` inlines mentioned in the debug maps
```
LagoLunatic, line 325: Yeah something probably isn't right here, I suspect some of these inlines do more than one thing based on their byte sizes in the debug maps. It's fine like this for now though, as more code gets decompiled we can narrow down the possibilities

src/d/actor/d_a_player_swim.inc
```diff
@@ -11,85 +11,613 @@
...
+    float target = 0.0f;
+
+    if (mAcch.GetGroundH() <= m35D0 - 175.0f) {
+        dComIfGp_setItemTimeMax(1);
+
+        if (checkNoResetFlg0(daPyFlg0_UNK100)) {
+            float swimTimerRate = getSwimTimerRate();
+
+            if (swimTimerRate > 0.5f) {
+                target = cM_scos(cM_rad2s((1.0f - swimTimerRate) * 3.1415927f)) * -l_HIO_swim_c0.m.field_0x78;
```
LagoLunatic, line 308: Does the pi define work here or is the precision wrong?
```suggestion
                target = cM_scos(cM_rad2s((1.0f - swimTimerRate) * M_PI)) * -l_HIO_swim_c0.m.field_0x78;
```

src/d/actor/d_a_player_swim.inc
```diff
@@ -11,85 +11,613 @@
...
+    }
+    dComIfGp_clearItemTimeCount();
+    dComIfGp_setItemTimeMax(0);
+    resetPriTextureAnime();
 }
 
 /* 8013D314-8013D354       .text checkSwimFallCheck__9daPy_lk_cFv */
 BOOL daPy_lk_c::checkSwimFallCheck() {
-    /* Nonmatching */
+    if (!checkNoResetFlg0(daPyFlg0_UNK80) || current.pos.y > m35D0 + 30.1f && !(dPa_control_c::mStatus & 1)) { // Fakematch? (dPa_control_c::mStatus & 1) seems like an inline
```
LagoLunatic, line 220:
```suggestion
    if (!checkNoResetFlg0(daPyFlg0_UNK80) || current.pos.y > m35D0 + 30.1f && !dPa_control_c::isStatus(1)) {
```
You can "Find all references" (alt+shift+f12) to a field like mStatus in VSCode and it should show you any functions that use it, like this inline

include/d/actor/d_a_player_HIO.h
```diff
@@ -705,11 +705,12 @@ class daPy_HIO_fall_c0 {
...
-    /* 0x00 */ s16 field_0x0;
-    /* 0x02 */ s16 field_0x2;
+    /* 0x00 */ s16 field_0x0; // 0x18
+    /* 0x02 */ s16 field_0x2; // 0x0
     /* 0x04 */ s16 field_0x4;
-    /* 0x06 */ s16 field_0x6;
-    /* 0x08 */ s16 field_0x8;
+    /* 0x06 */ s16 field_0x6; 
+    /* 0x08 */ s16 field_0x8; 
+    /* 0x0A */ //s16 field_0xA; missing?
```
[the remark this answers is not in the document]
mattias-blum: Actually this a remnant from some earlier code I used where I was referencing `HIO_swim` when I was meant to be referencing `HIO_turn`. I'll remove this comment

src/d/actor/d_a_player_swim.inc
```diff
@@ -11,85 +11,613 @@
...
 }
 
 /* 8013CF2C-8013CFC0       .text checkNextModeSwim__9daPy_lk_cFv */
 BOOL daPy_lk_c::checkNextModeSwim() {
-    /* Nonmatching */
+    BOOL res;
+
+    if (mpAttnEntryA && mpAttnEntryA->mType == 7) { //Action_Ship? fopAc_Attn_ACTION_SHIP_e is 0x80 so not sure why Ghidra is saying this 
+        dComIfGp_setDoStatus(0x1C); // Show `Get In` on the A Button? Ghidra says `GetInShip` is the symbol
+        if (mItemTrigger & 1) {
```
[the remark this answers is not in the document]
mattias-blum: Would these inlines be in `d_a_player.h` or `d_a_player_main`. I already replaced `mItemTrigger & 1` with `doTrigger()` so are you suggesting that the three lines about should have their logic placed in an inline?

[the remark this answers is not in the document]
LagoLunatic: No, `doTrigger()` should be all that's needed here afaik

PR #715, opened 2025-03-30
d_a_player_hang.inc - All matching except floats.

123456789758:


src/d/actor/d_a_player_hang.inc
```diff
@@ -11,98 +11,613 @@
...
+    return true;
 }
 
 /* 80135094-80135140       .text procHangWait_init__9daPy_lk_cFv */
 BOOL daPy_lk_c::procHangWait_init() {
-    /* Nonmatching */
+    commonProcInit(daPyProc_HANG_WAIT_e);
+    //!@bug? passes a signed short (daPy_HIO_wallCatch_c0::m.field_0x2) as a
+    //!float.
+    setSingleMoveAnime(ANM_VJMPCHA, 0.0f, daPy_HIO_wallCatch_c0::m.field_0x2, -1,
```
LagoLunatic, line 416: If you pass an integer typed variable as a float argument to a function the compiler should automatically convert it to float type and put it in the right floating point register, this doesn't sound like a bug unless I'm misunderstanding the comment

PR #722, opened 2025-03-30
d_a_player_boomerang.inc - All fully matching.

123456789758:


src/d/actor/d_a_player_boomerang.inc
```diff
@@ -16,37 +16,126 @@
...
+          (mDemo.getDemoType() == 0)) &&
+         (((m_anm_heap_upper[UPPER_MOVE2_e].mIdx == 0xffff ||
+            (checkPlayerGuard())) ||
+           ((m_anm_heap_upper[UPPER_MOVE2_e].mIdx == LKANM_BCK_DAMDASH ||
+             (checkBoomerangAnime())))))) &&
+        ((!(checkModeFlg(ModeFlg_WHIDE | ModeFlg_ROPE | ModeFlg_IN_SHIP |
+                         ModeFlg_CLIMB | ModeFlg_SWIM | ModeFlg_00080000 |
+                         ModeFlg_GRAB | ModeFlg_PUSHPULL | ModeFlg_LADDER |
+                         ModeFlg_CROUCH | ModeFlg_CRAWL)) ||
+          (mCurProc == daPyProc_SHIP_BOOMERANG_e)))) {
```
LagoLunatic, line 66: This looks like Ghidra output, should probably be cleaned up somehow to make it a bit more readable. Maybe fix the indentation on the left to be more consistent or split this into multiple nested if statements instead of a single big one?

PR #724, opened 2025-03-30
d_a_player_hook.inc - All matching except floats.

123456789758:


src/d/actor/d_a_player_hook.inc
```diff
@@ -157,20 +255,133 @@ BOOL daPy_lk_c::procHookshotSubject() {
...
+          if (m355E == 0) {
+            if (!mpAttention->Lockon()) {
+              resetActAnimeUpper(UPPER_MOVE2_e, -1.0f);
+              goto LAB_8014b69c;
+            }
+          }
+          setSpeedAndAngleAtn();
+        }
+      }
+    LAB_8014b69c:
```
LagoLunatic, line 299: If I'm not mistaken this goto can be resolved as if/elseif/else if the nesting is reduced
```suggestion
        } else if (m355E == 0 && !mpAttention->Lockon()) {
            resetActAnimeUpper(UPPER_MOVE2_e, -1.0f);
        } else {
            setSpeedAndAngleAtn();
        }
```

PR #732, opened 2025-04-02
d_camera 28%

mattias-blum:
FYI to everyone that I'm working on this


mattias-blum:
I wanted to check in with the changes I've made so far before attempting the larger functions in this TU. There are some issues with how the class members are declared (mainly with the `BG` struct and the class array of size 2 at `0xA4`) and I believe that there might be something wrong some of the `dBgS` objects which is causing functions like `getWaterSurfaceHeight` to not match. Would appreciate some advice on cleanup/getting some of the closer functions.

src/d/d_stage.cpp
```diff
@@ -1315,10 +1315,10 @@ int dStage_cameraCreate(stage_camera2_data_class* i_cameraData, int i_cameraIdx,
...
 
     if (i_cameraData != NULL) {
-        i_cameraData->field_0x4 = 0.0f;
-        i_cameraData->field_0x8 = 0.0f;
-        i_cameraData->field_0x4 = 0.0f;
-        i_cameraData->field_0x0 = param_2;
+        i_cameraData->m00.mpTypeStr[1] = 0;
+        i_cameraData->m00.mpTypeStr[2] = 0;
+        i_cameraData->m00.mpTypeStr[1] = 0;
+        i_cameraData->m00.mpTypeStr[2] = param_2;
```
LagoLunatic, line 1321: This isn't right and caused the function's match% to regress:
`dStage_cameraCreate__FP24stage_camera2_data_classii | fuzzy_match  | 100.00% -> 81.48%`

I'm gonna need to clean these structs up later since their names were already wrong, but for now let's just revert the camera-related structs in d_stage.h to how they previously were (correct contents but wrong names) so that they at least don't get worse

include/f_ap/f_ap_game.h
```diff
@@ -14,6 +14,7 @@ class fapGm_HIO_c : public JORReflexible {
     virtual ~fapGm_HIO_c();
 
     void genMessage(JORMContext* ctx);
+    f32 getAspectRatio() { return mAspectRatio; }
```
LagoLunatic, line 17: `fapGm_HIO_c::getAspectRatio()` doesn't appear in any maps for either TWW or TP.

I think you meant to create `fapGmHIO_getAspectRatio()`
image

include/d/d_event_manager.h
```diff
@@ -68,6 +68,7 @@ class dEvent_manager_c {
     int setStartDemo(int eventInfoIdx) { return mException.setStartDemo(eventInfoIdx); }
     void cancelStartDemo() { /* TODO */ }
     void setCameraPlay(u32 v) { mCameraPlay = v; }
+    int getCameraPlay() { return mCameraPlay; }
```
LagoLunatic, line 71: Slightly wrong name, should be `cameraPlay()`
FYI, when adding inlines you can just copy paste all of the already-demangled names out of the list of all weak functions pinned in Discord so you don't have to manually try to demangle them one by one (or use objdiff):
```
dEvent_manager_c::cameraPlay()
dEvent_manager_c::cancelStartDemo()
dEvent_manager_c::dEvent_manager_c()
dEvent_manager_c::getEventHIO()
dEvent_manager_c::getFlags()
dEvent_manager_c::setCameraPlay(int)
dEvent_manager_c::setStartDemo(int)
dEvent_manager_c::~dEvent_manager_c()
```
Also, whoever added `setCameraPlay` gave it the wrong param type (u32 vs int)

include/d/d_bg_s_roof_chk.h
```diff
@@ -21,7 +21,7 @@ class dBgS_RoofChk : public cBgS_PolyInfo, public cBgS_Chk, public dBgS_Chk {
 
     virtual ~dBgS_RoofChk() {}
 
-    void SetPos(cXyz& pos) { m_pos = pos; }
+    void SetPos(cXyz const& pos) { m_pos = pos; }
```
LagoLunatic, line 24: Careful copy pasting inlines from TP. The devs changed tons of functions to be const after TWW's release for TP, so many of them that should be const in TP should not be for TWW. And these copy-pasted consts actually affect regalloc and cause other matching problems.

include/d/d_bg_s_gnd_chk.h
```diff
@@ -67,14 +67,17 @@ class dBgS_ObjGndChk_All : public dBgS_ObjGndChk {
...
 
 class dBgS_CamGndChk_Wtr : public dBgS_CamGndChk {
 public:
-    dBgS_CamGndChk_Wtr();
+    dBgS_CamGndChk_Wtr() {
+        OffNormalGrp();
+        OnWaterGrp();
+    };
 
     virtual ~dBgS_CamGndChk_Wtr();
```
LagoLunatic, line 82:
```suggestion
    virtual ~dBgS_CamGndChk_Wtr() {}
```
Is this what you meant about dBgS causing functions to not match?

include/d/d_camera.h
```diff
@@ -86,25 +113,33 @@ class dCamera_c {
...
+    /* 0x070 */ cXyz m070;
+    /* 0x07C */ u32 m07C;
     /* 0x080 */ u32 m080;
     /* 0x084 */ cXyz m084;
     /* 0x090 */ cXyz m090;
     /* 0x09C */ f32 m09C;
-    /* 0x0A0 */ f32 m0A0;
+    /* 0x0A0 */ cSAngle m0A0;
+    /* 0x0A2 */ u8 m0A2[0x0A4 - 0x0A2];
     // /* 0x0A4 */ @class$4017d_camera_cpp m0A4[2];
```
LagoLunatic, line 131: I think this is an anonymous struct? Take a look at some of the structs in d_vibration.h for an example. Dunno what goes in the struct though

LagoLunatic (commented):
Partial review, didn't look at d_camera.cpp

include/d/d_bg_s_gnd_chk.h
mattias-blum: In functions like [this](https://decomp.me/scratch/F7YHk) it seems like there's some missing vtable functions that get called as part of `dBgS_CamGndChk_Wtr`'s dtor. Also the extra `addi` instruction might be something wrong with the ctor? Not too sure if that's what's going on here

LagoLunatic: https://decomp.me/scratch/OvRLs
The vtable function's aren't being called, it's just the destructor. You haven't implemented the destructor so it can't be inlined, which is why all the vtable stuff is missing from that scratch

include/d/d_camera.h
mattias-blum: I added a struct which seemed to work but not sure how to resolve the name (if that even matters).

mattias-blum:
PR changes made. Something I forgot to mention is that the dtor of `dCamera_c` suggests that there might be a `d2DBSplinePath` object at `0x3F8`? I've attached an image of the objdiff output if you have any thoughts.
Screenshot 2025-04-06 at 10 40 07 PM

include/d/d_camera.h
LagoLunatic: By "anonymous" I mean that it doesn't have any name specified by the programmer, that's why it gets a compiler-generated name:
```suggestion
    /* 0x0A4 */ struct {
        /* 0x00 */ int m00;
        /* 0x04 */ int m04;
        /* 0x08 */ int m08;
        /* 0x0C */ int m0C;
        /* 0x10 */ int m10;
        /* 0x14 */ int m14;
        /* 0x18 */ int m18;
        /* 0x1C */ int m1C;
    } m0A4[2];
```
I'm not sure if it would get the exact same name ("d_camera_cpp" seems to imply it's defined in the cpp file...? that doesn't seem right) but I guess this would be close enough

LagoLunatic:
> PR changes made. Something I forgot to mention is that the dtor of dCamera_c suggests that there might be a d2DBSplinePath object at 0x3F8? I've attached an image of the objdiff output if you have any thoughts.

I think this might be implying that there's a class member starting at 0x3F8 that encompasses everything up to at least the mSpline2DPath at 0x4C8. Looking at TP decomp, they have a class called `dCamera_c::dCamera_event_data`that ends in a splinepath, and the same name exists in WW debug maps.

I also see two other class names in the debug maps that you might want to look into: `dCamera_c::dCamera_DMC_system` and `dCamera_c::dCamera_monitoring_things`

mattias-blum:
Added those classes and it fixed the dtor. I made my best guess on `dCamera_DMC_system` and `dCamera_monitoring_things` based on the usage of certain class members in the code but I'll run `configure.py` to see if the sizes match the debug maps.

include/d/d_camera.h
LagoLunatic: You can't give it an explicit ctor/dtor, but the compiler will create one automatically if the struct has at least one member with a non-trivial ctor of its own (like an array or a cXyz). It looks like there might actually be nested anonymous structs here (`dCamera_c::@class$4336d_camera_cpp::@class$4337d_camera_cpp::~@class$4337d_camera_cpp()`) meaning the inner anonymous struct has at least one member with a ctor, and then the outer struct might inherit the ctor from the inner one.

src/d/d_camera.cpp
```diff
@@ -9,356 +9,2342 @@
...
+    m064 = 1.0f;
+    m5F4 = 0.0f;
+    mTrimHeight = 0.0f;
+    mTrimSize = 0;
+    mTrimTypeForce = -1;
+
+    dStage_stageDt_c* stage_dt = &dComIfGp_getStage();
+    if (stage_dt != NULL) {
+        stage_stag_info_class*  stag_info = stage_dt->getStagInfo();
+        if (stag_info && stag_info->mCameraMapToolID != -1) { // Bug?
```
LagoLunatic, line 250: Could this comment go into a little bit of detail? Just from looking at the C++ code at a glance I don't what the see the bug is here, though maybe it's obvious in the asm?

src/d/d_camera.cpp
```diff
@@ -9,356 +9,2342 @@
...
+    mMonitoringThings.field_0x0C.x = 0.0f;
+    mMonitoringThings.field_0x10 = 0;
+    mMonitoringThings.field_0x14 = 0.0f;
 }
 
 /* 801627A4-801628DC       .text updateMonitor__9dCamera_cFv */
 void dCamera_c::updateMonitor() {
-    /* Nonmatching */
+    /* Nonmatching - Code 100%, mBG offset issue */
+    float playerMonitorHoritzontalDist;
```
LagoLunatic, line 609:
```suggestion
    float playerMonitorHorizontalDist;
```

src/d/d_camera.cpp
mattias-blum: Oh it's just that I get this warning in VSCode `Result of comparison of constant -1 with expression of type 'u8' (aka 'unsigned char') is always true` so thought that maybe it was a bug left in by the devs

LagoLunatic: Oh I see, yeah that's probably a bug and they meant to do `!= 0xFF`.

Could something like comment "Bug, comparing unsigned value with -1"

src/d/d_camera.cpp
```diff
@@ -9,356 +9,2342 @@
...
+    }
+
+    if (bVar4) {
+        dComIfGp_onCameraAttentionStatus(mCameraInfoIdx, 0x10);
+    }
+    else {
+        dComIfGp_offCameraAttentionStatus(mCameraInfoIdx, 0x10);
+    }
+
+    if (mEventFlags >> 0x12 & 1) {
```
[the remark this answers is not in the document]
mattias-blum: No it produces `extrwi. r0, r3, 1, 13` rather than `rlwinm. r0, r3, 0, 13, 13`

src/d/d_camera.cpp
```diff
@@ -9,356 +9,2342 @@
...
+    mEventData.field_0x14 = -1;
+    mEventData.field_0x18 = -1;
+    mEventData.field_0x04 = -1;
+    mEventData.field_0x0c = -1;
+    mRoomNo = -1;
+    m318 = -1e+09f;
+    m310 = -1e+09f;
+    mBG.m58 = -1e+09f;
+    mBG.m04.OffNormalGrp();
+    mBG.m04.OnWaterGrp();
```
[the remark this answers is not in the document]
mattias-blum: I already have those inlines in the ctor for `dBgS_CamGndChk_Wtr` and when I set `m04` to `dBgS_CamGndChk_Wtr` and delete those lines the asm no longer matches. Odd.

src/d/d_camera.cpp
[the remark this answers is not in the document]
LagoLunatic: That might indicate the return value of `chkFlag` should be u32 rather than bool. Does changing that cause any regressions?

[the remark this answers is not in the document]
mattias-blum: Didn't have any effect one way or another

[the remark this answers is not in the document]
LagoLunatic: Oh wait, you meant `extrwi. r0, r3, 1, 13` is the instruction in the target and `rlwinm. r0, r3, 0, 13, 13` is the one in the base? I read that backwards, I thought you needed `rlwinm. r0, r3, 0, 13, 13`.

In that case you might want to to try having chkFlag return `return (mEventFlags & i_flag) != 0;` or `return (mEventFlags & i_flag) ? true : false;`, one of those might produce the extra shift

src/d/d_camera.cpp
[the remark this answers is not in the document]
mattias-blum: Also the ctor for `mBG` explicitly wants the member as `dBgS_CamGndChk`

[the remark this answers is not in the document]
LagoLunatic: Interesting, guess the devs just forget the Wtr one existed

PR #733, opened 2025-04-03
D a w bird

Zeclos:
unsure what the remaining unnamed fields are. field_0x2A0 seems to be related to to yspeed but im unsure of what it is exactly


src/d/actor/d_a_wbird.cpp
```diff
@@ -5,70 +5,208 @@
...
+        dComIfGp_setOperateWindOn();
+        mDoAud_seStart(JA_SE_TAKT_WIND_DISP);
+        field_0x29E++;
+    } else if(sVar2 > 10){
+        mDoAud_seStart(JA_SE_SYS_WTAKT_WIND_AMB);
+        switch (dComIfGp_getOperateWind()) {
+            case 1: {
+                mDoAud_seStart(JA_SE_TAKT_WIND_DECIDE);
+                setStartPos();
+                mAction = 2;
```
LagoLunatic, line 141: This actor has an inline in the debug maps that could be implemented here
```suggestion
                setAction(2);
```

src/d/actor/d_a_wbird.cpp
```diff
@@ -5,70 +5,208 @@
...
+    case 1:
+        i_this->actionEnd();
+        break;
+    default:
+        i_this->actionWait();
+        break;
+    }
+    i_this->calcMtx();
+    return TRUE;
 }
```
LagoLunatic, line 193: The `execute` and `draw` methods of this actor are inlined, so for example this logic can be moved into `daWbird_c::execute` and the static would then just be:
```suggestion
static BOOL daWbird_Execute(daWbird_c* i_this) {
    return i_this->execute();
}
```

For some actors this matters for matching so it's good to be aware of. For this one it doesn't, but still might as well get the code closer to how the original devs wrote it.

src/d/actor/d_a_wbird.cpp
```diff
@@ -5,70 +5,208 @@
...
+    }
+    else {
+        field_0x29E++;
+    }
 }
-
 /* 00000850-00000858       .text daWbird_Draw__FP9daWbird_c */
 static BOOL daWbird_Draw(daWbird_c*) {
-    /* Nonmatching */
+    return TRUE;
```
LagoLunatic, line 172:
```suggestion
    return i_this->draw();
```

src/d/actor/d_a_wbird.cpp
```diff
@@ -5,70 +5,208 @@
...
+const int windpower = 1;
 /* 00000078-000000AC       .text calcMtx__9daWbird_cFv */
 void daWbird_c::calcMtx() {
-    /* Nonmatching */
+    eyePos.x = current.pos.x;
+    eyePos.y = current.pos.y;
+    eyePos.z = current.pos.z;
+    attention_info.position.x = current.pos.x;
+    attention_info.position.y = current.pos.y;
+    attention_info.position.z = current.pos.z;
```
LagoLunatic, line 23:
```suggestion
    eyePos.set(current.pos);
    attention_info.set(current.pos);
```

src/d/actor/d_a_wbird.cpp
```diff
@@ -5,70 +5,208 @@
...
+                player->setPlayerPosAndAngle(&player->current.pos, sVar2);
+                if(dComIfGp_checkPlayerStatus0(0, 0x00010000)){
+                    mEventIdx = dComIfGp_evmng_getEventIdx("TACT_WINDOW2_SHIP");
+                } else{
+                    mEventIdx = dComIfGp_evmng_getEventIdx("TACT_WINDOW2");
+                }
+                fopAcM_orderChangeEventId(this, mEventIdx, 0, 0xFFFF);
+                player = daPy_getPlayerActorClass();
+                player->cancelOriginalDemo();
+                dKyw_custom_windpower(windpower);
```
LagoLunatic, line 154: Was pulling `windpower` out as a constant instead of writing
```suggestion
                dKyw_custom_windpower(1.0f);
```
an intentional choice?

If you do want it as a constant, I think it should have its type changed, since the parameter to dKyw_custom_windpower is not an int:
```diff
-const int windpower = 1;
+const f32 windpower = 1.0f;
```

src/d/actor/d_a_wbird.cpp
Zeclos: in the debug map i d_a_wbird.map i cant find this inline? when i replace it with setAction(2); it also doesnt match anymore in objdiff

src/d/actor/d_a_wbird.cpp
Zeclos: i cant find daWbird_c::execute or draw in the files of vscode, do i need to create these functions or am i not searching correctly

src/d/actor/d_a_wbird.cpp
LagoLunatic: d_a_wbird.map is the non-debug map, it doesn't have any inlines in it. d_a_wbirdD.map is the debug map.

The inline in the header is empty and just a placeholder that I generated with a script, you need to implement it or else it won't do anything.

src/d/actor/d_a_wbird.cpp
LagoLunatic: Yeah you'll need to create them. My script put declarations of them in the header files so you don't need to add anything there, but I didn't create empty placeholders in the .cpp file like I did for the smaller inlines that go in the headers, such as `setAction`

include/d/actor/d_a_wbird.h
```diff
@@ -7,19 +7,28 @@ class daWbird_c : public fopAc_ac_c {
 public:
     inline BOOL draw();
     inline BOOL execute();
-    void setAction(unsigned char) {}
+    void setAction(int action) {
```
LagoLunatic, line 10: The debug map says this inline takes an unsigned char, not an int. Did it not match when you tried it with unsigned char?

Zeclos: oh no i thought that i only pass along ints so i just made it an int in the inline, it does match with an unsigned char so ill just change it back then

PR #734, opened 2025-04-03
implement daObjTable functions

onnileivo:
shows 100% on objdiff so I think its done


mattias-blum:
If you want to check if it matches, you can set `d_a_obj_table` in `configure.py` from `NonMatching` to `Matching` and then re-build the project (Ctrl-Shift-B in VSCode).

onnileivo:
forgot to mention the issue, #159

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
 }
 
 namespace daObjTable {
 namespace {
 /* 00000608-00000628       .text Mthd_Create__Q210daObjTable27@unnamed@d_a_obj_table_cpp@FPv */
-cPhs_State Mthd_Create(void*) {
-    /* Nonmatching */
+void Mthd_Create(void* i_this) {
+    ((Act_c*) i_this)->Mthd_Create();
+    return;
```
LagoLunatic, line 162: These five methods at the end of the the TU are all supposed to return values, not be changed to void.
```suggestion
cPhs_State Mthd_Create(void* i_this) {
    return ((Act_c*) i_this)->Mthd_Create();
```

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
 /* 00000078-000001B8       .text CreateHeap__Q210daObjTable5Act_cFv */
 BOOL daObjTable::Act_c::CreateHeap() {
-    /* Nonmatching */
+    u8 prm;
+    J3DModelData *model_data;
+    J3DModel *model;
+
+
+    prm = daObj::PrmAbstract(this, PRM_LINK_ID_W, PRM_LINK_ID_S);
+    if (!prm) {
```
LagoLunatic, line 23: If you check the debug map, you'll see that this actor has an inline:
```suggestion
    if (prm_get_mdl() == 0) {
```
There's an empty placeholder for it in the actor's header, but you need to implement its body.

And you can also change the enum name "LINK_ID" which were copy pasted from a different actor to instead say "MDL" now that we know the real name of this parameter.

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
+        model_data = (J3DModelData*) dRes_control_c::getRes(M_arcname, 4, g_dComIfG_gameInfo.mResControl.mObjectInfo, 0x40);
+        
+        JUT_ASSERT(0x51, model_data != 0);
```
LagoLunatic, line 26:
```suggestion
        JUT_ASSERT(0x51, model_data != NULL);
```
Assertion strings convert defines NULL into their values, so this will still match

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
 BOOL daObjTable::Act_c::CreateHeap() {
-    /* Nonmatching */
+    u8 prm;
+    J3DModelData *model_data;
+    J3DModel *model;
+
+
+    prm = daObj::PrmAbstract(this, PRM_LINK_ID_W, PRM_LINK_ID_S);
+    if (!prm) {
+        model_data = (J3DModelData*) dRes_control_c::getRes(M_arcname, 4, g_dComIfG_gameInfo.mResControl.mObjectInfo, 0x40);
```
LagoLunatic, line 24: Make sure you're checking the debug map for any inlines.
```suggestion
        model_data = (J3DModelData*) dComIfG_getObjectRes(M_arcname, TABLE_BDL_YTBLE);
```
Also, there's an enum in res/res_table.h you can include to make these file indexes more readable.

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
+    uint coll_id;
+    u8 release;
+
+    cullMtx = mpModel->mBaseTransformMtx;
+    init_mtx();
+
+    fopAcM_setCullSizeBox(this, -80.0f, -1.0f, -80.0f, 80.0f, 205.0f, 80.0f);
+    prm = daObj::PrmAbstract(this, PRM_LINK_ID_W, PRM_LINK_ID_S);
+
+    if (((uint)prm == '\x02') && (coll = mpBgW, coll != NULL)) {
```
LagoLunatic, line 54: The mdl param here doesn't make sense as an ascii character '\x02', that's a Ghidra-ism. You should write it as the integer 2.
```suggestion
    if (prm_get_mdl() == 2 && (coll = mpBgW, coll != NULL)) {
```

Also, we don't want assignments inside of if statements, that's another Ghidra-ism.
```suggestion
    if (prm_get_mdl() == 2) {
        coll = mpBgW;
        if (coll != NULL) {
```

Though I doubt this temp variable is necessary, so you could probably just write:
```suggestion
    if (prm_get_mdl() == 2 && mpBgW != NULL) {
```

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
+        coll_id = (coll->GetId());
+        if (((int)coll_id >= 0) && ((int)coll_id < 0x100)) {
+            release = true;
+        } else {
+            release = false;
+        }
+        
+        if (release) {
```
LagoLunatic, line 62:
```suggestion
        if (mpBgW->ChkUsed()) {
```

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
+            dComIfG_Bgsp()->Release(coll);
+        }
+    }
+    return 1;
```
LagoLunatic, line 66:
```suggestion
    return TRUE;
```

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
-    /* Nonmatching */
+    cPhs_State phase_state;
+    u8 prm;
+    fopAcM_SetupActor(this, Act_c);
+
+    phase_state = dComIfG_resLoad(&mPhs, M_arcname);
+    if (phase_state == cPhs_COMPLEATE_e) {
+        prm = daObj::PrmAbstract((fopAc_ac_c*) this, PRM_LINK_ID_W, PRM_LINK_ID_S);
+        if (prm == 0) {
+            phase_state = MoveBGCreate(M_arcname, 9, 0, 0xffffffff);
```
LagoLunatic, line 79:
```suggestion
            phase_state = MoveBGCreate(M_arcname, TABLE_DZB_YTBLE, NULL, -1);
```

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
+    return ret;
 }
 
 /* 00000408-00000488       .text set_mtx__Q210daObjTable5Act_cFv */
 void daObjTable::Act_c::set_mtx() {
-    /* Nonmatching */
+    PSMTXTrans(mDoMtx_stack_c::now, current.pos.x, current.pos.y, current.pos.z);
+    mDoMtx_ZXYrotM(mDoMtx_stack_c::now, shape_angle.x, shape_angle.y, shape_angle.z);
+    PSMTXCopy(mDoMtx_stack_c::now, mpModel->mBaseTransformMtx);
+    PSMTXCopy(mDoMtx_stack_c::now, M_tmp_mtx);
```
LagoLunatic, line 107: Any time you see a PSMTX function being called, it's usually an inline.

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
+    return;
```
LagoLunatic, line 108:
```suggestion
```
Explicit return not needed at end of void function

src/d/actor/d_a_obj_table.cpp
```diff
@@ -4,78 +4,185 @@
...
+BOOL daObjTable::Act_c::Execute(Mtx** matrix) {
+    set_mtx();
+    *matrix = &M_tmp_mtx;
+    return 1;
 }
 
 /* 00000500-00000608       .text Draw__Q210daObjTable5Act_cFv */
 BOOL daObjTable::Act_c::Draw() {
-    /* Nonmatching */
+    f32 angle; // not 100% sure it's an angle, best guess
```
LagoLunatic, line 132: Angles are pretty much always s16s, so I doubt this is right. It's better to leave the variable with a placeholder name so someone knows to fix it later rather than an incorrect name if you're not sure.

LagoLunatic (commented):
Missing a ton of inlines, probably more that I didn't mention here too. Make sure you're referencing the debug maps so you don't miss these.

onnileivo:
Thanks for the pointers! I'm going to probably fix all that tomorrow if I have time. I remember trying to fix the ugly ghidra code but it leading to non matching assembly, but I probably wrote something wrong then. I have also not yet been granted read access to the twilight princess server, so I don't have the debug map.
Thanks!

LagoLunatic:
The Wind Waker debug *maps* are pinned in the Discord server, you don't need Ghidra access for them.

The TP Ghidra server is just for the TP debug *build*, which is mostly useful for TP decomp + any engine code that is shared with Wind Waker. But for this actor, I doubt TP would help.

onnileivo:
Looks like it doesnt match the demo? Is this fixable or is it just that the demo has different functionality? If so, the configure py file be marked matcing for only the releases, right?

LagoLunatic:
> Looks like it doesnt match the demo? Is this fixable or is it just that the demo has different functionality? If so, the configure py file be marked matcing for only the releases, right?

Yes, just mark it as matching for retail only like most of the other TUs are

PR #740, opened 2025-04-14
D_a_obj_iceisland

Zeclos:


configure.py
```diff
@@ -1665,7 +1665,7 @@ def MatchingFor(*versions):
     ActorRel(Matching,    "d_a_obj_homensmoke", extra_cflags=['-pragma "nosyminline on"']),
     ActorRel(NonMatching, "d_a_obj_hsehi1"),
     ActorRel(NonMatching, "d_a_obj_htetu1"),
-    ActorRel(NonMatching, "d_a_obj_iceisland"),
+    ActorRel(Matching, "d_a_obj_iceisland", extra_cflags=['-pragma "nosyminline on"']),
```
LagoLunatic, line 1668: There's some kind of version difference, if you don't know what it is you can mark it matching for only JPN and PAL so the USA build doesn't fail:
```suggestion
    ActorRel(MatchingFor("GZLJ01", "GZLP01"), "d_a_obj_iceisland", extra_cflags=['-pragma "nosyminline on"']),
```

src/d/actor/d_a_obj_iceisland.cpp
```diff
@@ -5,95 +5,272 @@
...
+    cullSizeFar = 1.0f;
+    set_mtx();
+    dKy_tevstr_init(&mTevStr, fopAcM_GetHomeRoomNo(this), 0xFF);
+    g_env_light.settingTevStruct(TEV_TYPE_BG0, &current.pos, &mTevStr);
+    mEmitter1 = NULL;
+    mEmitter2 = NULL;
+    u8 switchNo = daObjIceisland_prm::getSwitchNo(this);
+    if(fopAcM_isSwitch(this, switchNo)){
+        mBrkAnm.setFrame(mBrkAnm.getEndFrame());
+        if(dComIfGs_getStartPoint() == 2 && current.roomNo == dComIfGs_getRestartRoomNo()){// some if
```
LagoLunatic, line 61: Looks like an old comment
```suggestion
        if(dComIfGs_getStartPoint() == 2 && current.roomNo == dComIfGs_getRestartRoomNo()){
```

PR #749, opened 2025-05-07
Player swimming documentation

mattias-blum:
Also included details on how the [SuperSwim](https://www.zeldaspeedruns.com/tww/techniques/superswim) glitch is possible


mattias-blum:
Still going to document the remainder of these functions but since this stuff is relevant for speed-runners I'd like to check these changes in first so they can be shared with them

magcius:
daNpc_Md_c is the class for Medli, and I don't think that's relevant for superswimming. Are you sure your documentation is correct?

mattias-blum:
> daNpc_Md_c is the class for Medli, and I don't think that's relevant for superswimming. Are you sure your documentation is correct?

Yeah I thought the same because it's odd that this function is declared both in `daPy_lk_c` and `daPy_md_c` but only implemented in `daPy_md_c`. Based purely on the code in that function it would make sense as a general purpose function for calculating the movement for both these actors as Medli's movement from a gameplay standpoint feels similar to Link's.

I'm trying to parametrize these functions so that I can run some Python sims to see if I can get the same results that the speed-running community got with their empirical investigations. I'm fairly confident that this is the code responsible for the super-swimming behaviour since some things are consistent with already existing documentation on the ZSR page such as losing 2 units of speed per frame which is a result of this function call:
```
setNormalSpeedF(
        calculatedSpeed,
        daPy_HIO_swim_c0::m.acceleration,
        daPy_HIO_swim_c0::m.speedStepMax,
        daPy_HIO_swim_c0::m.speedStepMin
);
```
adding daPy_HIO_swim_c0::m.acceleration (2.0f) to the negative speed. There's also the animation frame component that contributes to the speed I need to account for but I think I've also found the relevant code for that as well.

LagoLunatic:
> Yeah I thought the same because it's odd that this function is declared both in `daPy_lk_c` and `daPy_md_c` but only implemented in `daPy_md_c`. Based purely on the code in that function it would make sense as a general purpose function for calculating the movement for both these actors as Medli's movement from a gameplay standpoint feels similar to Link's.

It's not the same function. `daPy_lk_c::setNormalSpeedF(f32, f32, f32, f32)` is a member function in Link's class, while `daNpc_Md_c::setNormalSpeedF(f32, f32, f32, f32, f32)` is a member function in Medli's class. Link doesn't call Medi's functions, he calls his own functions. The reason they have the same name is probably because Link was coded first and then whoever coded Medli copy-pasted some of his functions to have as a base to start coding her, but there are differences (e.g. these two functions take a different number of arguments).

It's also not true that Link doesn't implement the function, he does, but it hasn't been decompiled so you can't document it yet. If you hold Ctrl in VSCode and click on a function call, it will take you to that function's body. So if you click the call in player_swim.inc it takes you to the function in player_main.cpp, not npc_md.cpp. This is more reliable than assuming that function names are unique.

Taking a quick glance at `daPy_lk_c::setNormalSpeedF` in Ghidra, I do think your guess about the superswim bug being caused by that function is reasonable, as it seems they still have that one line that checks if speed > max but no equivalent check for speed < -max, but some of the details are definitely different, so you're not gonna want to run your simulations on Medli's code with Link's parameters or the results will be wrong.

mattias-blum:
Okay thanks for the clarification, I'll just go ahead and decompile that function and move my comments to the appropriate location.

src/d/actor/d_a_npc_md.cpp
```diff
@@ -1629,41 +1629,64 @@ BOOL daNpc_Md_c::hitNpcAction(void* r29) {
...
-    f32 f31 = mMaxNormalSpeed * f1;
-    int r3 = wallHitCheck();
-    if (r3 >= 0) {
-        s16 angle = current.angle.y + 0x8000 - mAcchCir[r3].GetWallAngleY();
+void daNpc_Md_c::setNormalSpeedF(f32 i_speedScale, f32 i_acceleration, f32 i_scale, f32 i_maxStep, f32 i_minStep) {
+    // Calculate the desired speed based on a scaling factor and the NPC's maximum normal speed
+    f32 desiredSpeed = mMaxNormalSpeed * i_speedScale;
+
+    // Check if Medli is colliding with a wall
+    int wallIndex = wallHitCheck();
```
LagoLunatic, line 1637: This returns the index of the AcchCir that is colliding with a wall, not the wall itself. Should be named "cir index" or something like that.

src/d/actor/d_a_npc_md.cpp
```diff
@@ -1629,41 +1629,64 @@ BOOL daNpc_Md_c::hitNpcAction(void* r29) {
...
+
     f32 targetSpeed;
     f32 maxStep;
-    f32 f28;
-    if (f31 < m3108) {
-        f32 temp2 = m3108 - f31;
-        if (temp2 > i_maxStep) {
+
+    // If slowing down (desired speed is less than current speed)
+    if (desiredSpeed < mCurrentSpeed) {
```
LagoLunatic, line 1653: Some of these comments seem excessive. For example, does "If slowing down (desired speed is less than current speed)" give any more information than the line of code `if (desiredSpeed < mCurrentSpeed) {`?

It's better to comment either the purpose of the entire function above it, or of individual lines that do things that aren't immediately intuitive, rather than commenting every single line of code. Variable names should handle documenting most lines of code.

include/d/actor/d_a_player_HIO.h
```diff
@@ -712,36 +712,36 @@ class daPy_HIO_swim_c1 {
...
-    /* 0x0C */ f32 field_0xC;  // 18.0f
-    /* 0x10 */ f32 field_0x10; // 1.0f
-    /* 0x14 */ f32 field_0x14; // -50.0f
-    /* 0x18 */ f32 field_0x18; // 0.02f
-    /* 0x1C */ f32 field_0x1C; // 2.0f
-    /* 0x20 */ f32 field_0x20; // 0.5f
-    /* 0x24 */ f32 field_0x24; // 90.0f
-    /* 0x28 */ f32 field_0x28; // 15.0f
+    /* 0x04 */ s16 angleTurnRateMin;  // 0x1388
+    /* 0x06 */ s16 angleTurnRateMax;  // 0x4B0
```
LagoLunatic, line 716: These are swapped, min should be smaller than max.
`cLib_addCalcAngleS` takes max then min as arguments, not min then max

src/d/actor/d_a_player_main.cpp
```diff
@@ -2210,66 +2210,105 @@ int daPy_lk_c::getDirectionFromAngle(s16 angle) {
...
-        if (mVelocity > dVar10) {
-            mVelocity = dVar10;
-        }
-    } else {
-        cLib_addCalc(&mVelocity, dVar6, param_2, temp_f3, param_4);
-    }
+void daPy_lk_c::setNormalSpeedF(f32 i_acceleration, f32 i_scale, f32 i_maxStep, f32 i_minStep) {
+   f32 desiredSpeed;
+   // Calculate the desired speed based on the stick distance and Link's maximum normal speed
+   if (dComIfGp_event_getMode() != dEvtMode_NONE_e || checkHeavyStateOn() || field_0x2b0 < 0.0f) {
```
LagoLunatic, line 2225: Previous inline usage was more likely to be real (dComIfGp_event_runCheck, checkGrabWear)

src/d/actor/d_a_player_main.cpp
```diff
@@ -2210,66 +2210,105 @@ int daPy_lk_c::getDirectionFromAngle(s16 angle) {
...
+           Note: 
+           Negative speed is added additively as well
+       */
+       mVelocity += i_acceleration;
+        // Clamp to the desired speed if it would exceed it
+        /*
+           Note:
+           This is purely a clamp on the positive speed and the lack of one 
+           for negative speed is what gives rise to super-swimming
+        */
```
LagoLunatic, line 2297: Since the purpose of this comment is to document a bug, it would be helpful to include the word "bug" in here somewhere so someone searching the codebase for bugs can find all of them relatively easily.

src/d/actor/d_a_player_swim.inc
```diff
@@ -390,65 +616,107 @@ BOOL daPy_lk_c::procSwimUp() {
...
+ *   - param_1: If TRUE, adjust Link's velocity based on his progress in the swimming animation
+ * 
+ * Returns:
+ *   - TRUE when initialization of the swim wait process is successful.
+ */
 BOOL daPy_lk_c::procSwimWait_init(BOOL param_1) {
-    float fVar2;
+    float swim_animation_state;
     
     J3DFrameCtrl* mFrameCtrl = mFrameCtrlUnder;
```
LagoLunatic, line 635: The `m` prefix is for member variables. Local variables should be called something like `frame_ctrl`. https://github.com/zeldaret/tww/blob/main/docs/coding_guidelines.md#naming-style

src/d/actor/d_a_player_swim.inc
```diff
@@ -390,65 +616,107 @@ BOOL daPy_lk_c::procSwimUp() {
...
 
     if (param_1) {
-        fVar2 = mFrameCtrl->getFrame() / (float)mFrameCtrl->getEnd();
-        mVelocity = speedF * (1.0f - daPy_HIO_swim_c0::m.field_0x60) + speedF * std::fabsf(cM_fcos(fVar2 * M_PI)) * daPy_HIO_swim_c0::m.field_0x60;
+        // Adjust Link's velocity based on his forward speed and his current position within his swimming animation
+        /*
+            Given:
+            s: forward speed (`speedF`)
+            x: Link's current animation position (`mFrameCtrl->getFrame()`)
+            swim_animation_state: Represents (as a percentage) how far Link is through one cycle of his swimming animation
```
LagoLunatic, line 645: This local var ranges from 0.0-1.0? I feel like "state" isn't the clearest name for it, that sounds like a mode or state machine.

Maybe `anim_blend`/`swim_anim_blend`?

src/d/actor/d_a_player_swim.inc
```diff
@@ -314,6 +521,15 @@ void daPy_lk_c::setSwimTimerStartStop() {
 }
 
 /* 8013D814-8013DA18       .text procSwimUp_init__9daPy_lk_cFi */
+/*
+ * Starts or stops the air meter timer based on Link's vertical position.
+ *
+ * If Link is significantly below the surface, the timer begins. Also sets the
+ * drowning voice clip based on air meter level and current swim animation loop.
+ * 
+ * Adjusts `m3608` field using a cosine function to reflect Link's oxygen level.
```
LagoLunatic, line 530: This comment doesn't seem right, was this just copy-pasted from the above function?

LagoLunatic (commented):
Partial review

LagoLunatic:
Closing because there are like 50+ conflicts now, if you're still interested in fixing this it may be easier to start from scratch

PR #755, opened 2025-05-15
d_a_tag_kk1

WilliamArnett:
Equivalent. Weak func ordering.

### DIFF
--- a/src/d/actor/d_a_tag_kk1.cpp
+++ b/src/d/actor/d_a_tag_kk1.cpp
@@ -3,59 +3,50 @@
 // Translation Unit: d_a_tag_kk1.cpp
 //
 
+
 #include "d/actor/d_a_tag_kk1.h"
 #include "d/d_com_inf_game.h"
 #include "d/d_procname.h"
 
+
 static daTag_Kk1_HIO_c l_HIO;
-static f32 a_prm_tbl[] = {350.0f,30.0f,0.0f};
+
+static const u8 dummy5[] = { 0x00, 0xFF, 0x00, 0x80};
 
 daTag_Kk1_HIO_c::daTag_Kk1_HIO_c() {
-    /* Nonmatching */
-    // 99% matched
-    field_0x8 = a_prm_tbl[0];
-    field_0xC = a_prm_tbl[1];
-    field_0x10 = *(u8*)&a_prm_tbl[2];
-    mNo = -0x01;
-}
-void daTag_Kk1_HIO_c::foo() {
-
+    static f32 a_prm_tbl[] = {350.0f,30.0f,0.0f};
+    mHorizontalDistance = a_prm_tbl[0];
+    mVerticalDistance = a_prm_tbl[1];
+    mUnusedU8 = *reinterpret_cast<u8*>(&a_prm_tbl[2]); //TODO: This u8 appears to be inside the float table. May be a better way to refactor.
+    mNo = -1;
 }
 
 /* 00000120-00000128       .text createInit__11daTag_Kk1_cFv */
 BOOL daTag_Kk1_c::createInit() {
-    return 1;
+    return TRUE;
 }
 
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_tag_kk1.cpp
```diff
@@ -3,64 +3,120 @@
...
+static daTag_Kk1_HIO_c l_HIO;
+
+static const u8 dummy5[] = { 0x00, 0xFF, 0x00, 0x80};
+
 daTag_Kk1_HIO_c::daTag_Kk1_HIO_c() {
-    /* Nonmatching */
+    static f32 a_prm_tbl[] = {350.0f,30.0f,0.0f};
+    mHorizontalDistance = a_prm_tbl[0];
+    mVerticalDistance = a_prm_tbl[1];
+    mUnusedU8 = *reinterpret_cast<u8*>(&a_prm_tbl[2]); //TODO: This u8 appears to be inside the float table. May be a better way to refactor.
```
LagoLunatic, line 20: a_prm_tbl seems like it's not an array of floats, but a struct.

If you check the debug map, there's this symbol:
`__as__Q215daTag_Kk1_HIO_c9hio_prm_cFRCQ215daTag_Kk1_HIO_c9hio_prm_c`
Which objdiff's demangler turns into this (a compiler-generated copy assignment function):
`daTag_Kk1_HIO_c::hio_prm_c::operator=(const daTag_Kk1_HIO_c::hio_prm_c&)`

This tells us that inside of the `daTag_Kk1_HIO_c` class body in the header there's struct named `hio_prm_c`. Take a look at d_a_nh for an existing example. These three assignments don't need to be done manually, once you have the struct set up right it will just be:
```suggestion
    static hio_prm_c a_prm_tbl = {
        350.0f,
        30.0f,
        0,
    };
    prm = a_prm_tbl;
```

src/d/actor/d_a_tag_kk1.cpp
```diff
@@ -3,64 +3,120 @@
 // Translation Unit: d_a_tag_kk1.cpp
 //
 
+
 #include "d/actor/d_a_tag_kk1.h"
+#include "d/d_com_inf_game.h"
 #include "d/d_procname.h"
 
-/* 000000EC-00000120       .text __ct__15daTag_Kk1_HIO_cFv */
```
LagoLunatic, line 9: Let's keep this automatically generated comments in, the mangled symbol name is convenient for quick searches

src/d/actor/d_a_tag_kk1.cpp
```diff
@@ -3,64 +3,120 @@
...
-    /* Nonmatching */
+    return true;
 }
 
 /* 00000130-0000024C       .text _execute__11daTag_Kk1_cFv */
 bool daTag_Kk1_c::_execute() {
-    /* Nonmatching */
+
+    cXyz* player_pos = &dComIfGp_getPlayer(0)->current.pos; 
+    f32 distance = std::sqrtf(current.pos.abs2(*player_pos));
```
LagoLunatic, line 38: Inline
```suggestion
    f32 distance = current.pos.abs(dComIfGp_getPlayer(0)->current.pos);
```

src/d/actor/d_a_tag_kk1.cpp
```diff
@@ -3,64 +3,120 @@
...
+    f32 vert_distance = dComIfGp_getPlayer(0)->current.pos.y - this->current.pos.y;
```
LagoLunatic, line 39: Explicit `this->` not required
```suggestion
    f32 vert_distance = dComIfGp_getPlayer(0)->current.pos.y - current.pos.y;
```

src/d/actor/d_a_tag_kk1.cpp
WilliamArnett: Well spotted. I'll do my best to remember this.

PR #756, opened 2025-05-18
100% d_a_kui

ammaraskar:
All functions are matching, there is a block of code in `daKui_Execute` that needs fields in `dr2_class` and `btd_class`, I added those in order to allow the TU to match.


ammaraskar:
Closes https://github.com/zeldaret/tww/issues/111

src/d/actor/d_a_kui.cpp
```diff
@@ -4,68 +4,541 @@
...
+#include "f_op/f_op_actor_mng.h"
+#include "f_pc/f_pc_executor.h"
+#include "JSystem/JUtility/JUTReport.h"
 
 class J3DModelData;
 
 /* 00000078-000000C4       .text s_a_i_sub__FPvPv */
-void s_a_i_sub(void*, void*) {
-    /* Nonmatching */
+static void* s_a_i_sub(void* i_this, void*) {
```
LagoLunatic, line 24: The first param to search functions is the actor it's checking against to see if it matches, not the `this` actor. So let's name it something else like:
```suggestion
static void* s_a_i_sub(void* search, void*) {
```

src/d/actor/d_a_kui.cpp
```diff
@@ -4,68 +4,541 @@
...
-void search_dragontail(kui_class*) {
-    /* Nonmatching */
+static dr2_class* search_dragontail(kui_class* i_this) {
+    return (dr2_class*)fpcM_Search(&s_a_i_sub, i_this);
 }
 
 /* 000000F0-0000013C       .text b_a_i_sub__FPvPv */
-void b_a_i_sub(void*, void*) {
-    /* Nonmatching */
+static void* b_a_i_sub(void* i_this, void*) {
```
LagoLunatic, line 37:
```suggestion
static void* b_a_i_sub(void* search, void*) {
```

src/d/actor/d_a_kui.cpp
```diff
@@ -4,68 +4,541 @@
...
 }
 
 /* 00000540-00000920       .text demo_camera__FP9kui_class */
-void demo_camera(kui_class*) {
-    /* Nonmatching */
+static void demo_camera(kui_class* i_this) {
+    camera_class* camera = dComIfGp_getCamera(dComIfGp_getPlayerCameraID(0));
+    s8 bVar2 = true;
+
+    switch ((s8)i_this->field_0x2E8) {
```
LagoLunatic, line 137: Since you seem to have a `(s8)` cast on all reads of `field_0x2E8`, that might indicate its type should be changed from `u8` to `s8` and the casts can be removed. Unless you already tried that and it broke something.

src/d/actor/d_a_kui.cpp
```diff
@@ -4,68 +4,541 @@
...
+            camera->mCamera.Stop();
+            camera->mCamera.SetTrimSize(2);
+            i_this->field_0x2EA = 0;
+            // fallthrough
+        case 2:
+            i_this->current.pos.y = i_this->home.pos.y + i_this->field_0x30C;
+
+            s32 uVar3 = (s32) i_this->field_0x2EA;
+            if (uVar3 < 20) {
+                f32 sin_result = cM_ssin((uVar3 & 0x1F) << 11);
```
LagoLunatic, line 157: Probably intended to be a multiplication since it's an angle
```suggestion
                f32 sin_result = cM_ssin((uVar3 & 0x1F) * 0x800);
```

src/d/actor/d_a_kui.cpp
```diff
@@ -4,68 +4,541 @@
...
+            if (target_x_angle > unk) {
+                target_x_angle = unk;
+                unk_flag = 1;
+            } else if (target_x_angle < (s16)-unk) {
+                target_x_angle = -unk;
+                unk_flag = 2;
+            }
+
+            if (unk_flag != 0 && unk > 2000) {
+                s16* shorts_array = (&i_this->field_0x2DC);
```
LagoLunatic, line 256: Does changing field_0x2DC's type to `s16[3]` not match in some way?

src/d/actor/d_a_kui.cpp
```diff
@@ -4,68 +4,541 @@
...
+                i_this->field_0x308 = 1000;
+            }            
+        } else {
+            actor->current.pos.y = actor->home.pos.y - 70.0f;
+        }
+
+        if (i_this->field_0x308 != 0) {
+            i_this->field_0x308--;
+
+            s16 finished = REG8_S(3) + 0x3CA;
```
LagoLunatic, line 377:
```suggestion
            s16 finished = REG8_S(3) + 970;
```
This can be written in decimal since it's probably a timer value for something (like the above 1000 is also in decimal)

src/d/actor/d_a_kui.cpp
```diff
@@ -4,68 +4,541 @@
...
-static BOOL daKui_CreateHeap(fopAc_ac_c*) {
-    /* Nonmatching */
+static BOOL daKui_CreateHeap(fopAc_ac_c* a_this) {
+    kui_class* i_this = (kui_class*)a_this;
+    J3DModelData* modelData;
+
+    if (i_this->type == 3) {
+        // Bell body
+        modelData = (J3DModelData*)dComIfG_getObjectRes("Kui", KUI_BDL_HKANE1);
+        JUT_ASSERT(0x353, modelData != 0);
```
LagoLunatic, line 415: Defines in asserts get replaced with their values, so these were probably written more like this originally
```suggestion
        JUT_ASSERT(0x353, modelData != NULL);
```

src/d/actor/d_a_kui.cpp
ammaraskar: I thought I tried it earlier but I must have had some other error. It works with a `s16[3]` now and is much cleaner, thanks!

src/d/actor/d_a_kui.cpp
ammaraskar: The cast in the `switch` still seems necessary but the cast at the end of the function goes away changing it to a `s8`.

PR #758, opened 2025-05-18
d_a_hys 100% equivalent with weak func order

ammaraskar:
All sections are 100% in objdiff 
image

I tried both `-pragma "nosyminline on"` and `-sym off` but neither of them fully match. `nosyminline` gets close but it looks like the issue is related to the call order of the parent class? Not sure if there's anything else I should try there.

image

### DIFF
--- a/src/d/actor/d_a_hys.cpp
+++ b/src/d/actor/d_a_hys.cpp
@@ -4,11 +4,18 @@
 //
 
 #include "d/actor/d_a_hys.h"
+#include "d/d_com_inf_game.h"
+#include "m_Do/m_Do_mtx.h"
 #include "d/d_procname.h"
+
+const char* daHys_c::m_arcname[2] = {"Hys", "Hys"};
+const s16 daHys_c::m_dzbidx[2] = {0xB, 0xB};
+const u32 daHys_c::m_heapsize[2] = {0xA00, 0xA00};
 
 /* 00000078-000000B8       .text Delete__7daHys_cFv */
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_hys.cpp
```diff
@@ -4,86 +4,231 @@
...
 
+#include "weak_data_1811.h" // IWYU pragma: keep
+
+const char* daHys_c::m_arcname[2] = {"Hys", "Hys"};
+
+/* Model file indexes. */
+const s16 daHys_c::m_bdlidx[2] = {5, 5};
+/* Texture file  */
+const s16 daHys_c::m_btpidx[2] = {8, 8};
+const s16 daHys_c::m_dzbidx[2] = {11, 11};
```
LagoLunatic, line 19: HYS_RES_FILE_ID enum for these indexes

src/d/actor/d_a_hys.cpp
```diff
@@ -4,86 +4,231 @@
...
+    }
+
+    return res;
 }
 
 /* 00000250-00000368       .text Create__7daHys_cFv */
 BOOL daHys_c::Create() {
-    /* Nonmatching */
+    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -240.0, -240.0, -90.0, 240.0, 240.0, 90.0);
```
LagoLunatic, line 97: `f` suffix for single-precision float literals

PR #769, opened 2025-05-26
D_a_obj_pfall matching

Zeclos:
saw that  `#include "d/d_priority.h"` got added but havent really looked if i need to do anything with it


configure.py
```diff
@@ -1697,7 +1697,7 @@ def MatchingFor(*versions):
     ActorRel(NonMatching, "d_a_obj_ospbox"),
     ActorRel(Matching,    "d_a_obj_paper", extra_cflags=['-pragma "nosyminline on"']),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_obj_pbka"),
-    ActorRel(NonMatching, "d_a_obj_pfall"),
+    ActorRel(Matching, "d_a_obj_pfall", extra_cflags=['-pragma "nosyminline on"']),
```
LagoLunatic, line 1700:
```suggestion
    ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"), "d_a_obj_pfall", extra_cflags=['-pragma "nosyminline on"']),
```
It only matches for retail, so unmark it as matching for the demo version so the build doesn't fail

src/d/actor/d_a_obj_pfall.cpp
```diff
@@ -5,146 +5,473 @@
...
+    }
+    mpBgW2 = new dBgW();
+
+    if(mpBgW2 == NULL) {
+        return FALSE;
+    }
+    
+    pData = (cBgD_t*)dComIfG_getObjectRes("Pfall", PFALL_DZB_AOTOSI);
+    
+    return (mpBgW2->Set(pData, cBgW::MOVE_BG_e, &field_0x404) & 0xff) != 1 ? 1 : 0;
```
LagoLunatic, line 172: Does this still match?
```suggestion
    if (mpBgW2->Set(pData, cBgW::MOVE_BG_e, &field_0x404) == true) {
        return FALSE;
    }

    return TRUE;
```

src/d/actor/d_a_obj_pfall.cpp
```diff
@@ -5,146 +5,473 @@
...
 }
 
 /* 000010A4-00001130       .text mode_proc_call__13daObj_Pfall_cFv */
 void daObj_Pfall_c::mode_proc_call() {
-    /* Nonmatching */
+        typedef void(daObj_Pfall_c::*ProcFunc)(void);
+        static const ProcFunc mode_proc[] = {
+        &daObj_Pfall_c::mode_wait,
+        &daObj_Pfall_c::mode_event,
+    };
```
LagoLunatic, line 363: Indented
```suggestion
    typedef void(daObj_Pfall_c::*ProcFunc)(void);
    static const ProcFunc mode_proc[] = {
        &daObj_Pfall_c::mode_wait,
        &daObj_Pfall_c::mode_event,
    };
```

src/d/actor/d_a_obj_pfall.cpp
```diff
@@ -5,146 +5,473 @@
...
+            if (!(frame > 0.0f && frame <= 6.0f)) {
+                if (frame > 6.0f && frame <= 10.0f) {
+                    field_0x438 += 1.0f;
+                }
+                else {
+                    if (!(frame > 10.0f && frame <= 23.0f) && frame > 23.0f && frame <= 29.0f) {
+                        field_0x438 -= 3.0f;
+                    }
+                }
+            }
```
LagoLunatic, line 392: This is too indented too

PR #776, opened 2025-05-29
d_a_ff - 98% matching

123456789758:


src/d/actor/d_a_ff.cpp
```diff
@@ -1,55 +1,363 @@
...
+        i_this->mTimers[2] = cM_rndF(200.0f) + 100.0f;
+    }
+    if (i_this->mTimers[1] == 0) {
+        i_this->mScaleTarget = cM_ssin(i_this->mLiveTimer * 1000) * 0.15f * 0.25f + 0.225f;
+    } else {
+        i_this->mScaleTarget = 0.0f;
+    }
+    cLib_addCalc2(&i_this->mScale, i_this->mScaleTarget, 0.1f, 0.05f);
+    cVar1 = i_this->mState;
+    if (cVar1 == 2) {
```
ammaraskar, line 109: If you'd like to fix the matching issue here, the problem is that this is a switch-case on mState, which is why the conditions look so weird here. Take a look this part of the tutorial on identifying and dealing with switch-cases: https://github.com/zeldaret/tww/blob/main/docs/decompiling.md#recognizing-switch-statements

src/d/actor/d_a_ff.cpp
```diff
@@ -1,55 +1,363 @@
...
+            cLib_addCalc2(&i_this->current.pos.x, i_this->mHomePos.x, 0.1f,
+                          std::abs(i_this->speed.x));
+            cLib_addCalc2(&i_this->current.pos.y, i_this->mHomePos.y, 0.1f,
+                          std::abs(i_this->speed.y));
+            cLib_addCalc2(&i_this->current.pos.z, i_this->mHomePos.z, 0.1f,
+                          std::abs(i_this->speed.z));
+            fVar10 = player->current.pos.x - i_this->current.pos.x;
+            fVar2 = player->current.pos.y - i_this->current.pos.y;
+            fVar3 = player->current.pos.z - i_this->current.pos.z;
+            fVar10 = std::sqrtf(fVar3 * fVar3 + fVar10 * fVar10 + fVar2 * fVar2);
```
ammaraskar, line 131: There is a `cXyz.abs` function that performs this sort of distance calculation between two vectors, that might be what was used here.

LagoLunatic: C++-style classes usually use member methods like `cXyz::abs`, while C-style classes like this one usually don't and instead prefer to manually do the math.

src/d/actor/d_a_ff.cpp
```diff
@@ -1,55 +1,361 @@
...
-    /* Nonmatching */
+static void fire_fly_draw(ff_class* i_this) {
+    MtxTrans(i_this->current.pos.x, i_this->current.pos.y, i_this->current.pos.z, false);
+    MtxScale(i_this->mScale, i_this->mScale, i_this->mScale, true);
+    i_this->mpModel[0]->setBaseTRMtx(*calc_mtx);
+    if (i_this->mScale > 0.01f) {
+        i_this->mBrkAnm[0]->play();
+        i_this->mBrkAnm[0]->entry(i_this->mpModel[0]->getModelData());
+        g_dComIfG_gameInfo.drawlist.setOpaListMaskOff();
+        g_dComIfG_gameInfo.drawlist.setXluListMaskOff();
```
LagoLunatic, line 27: Inline from the debug map
```suggestion
        dComIfGd_setListMaskOff();
```

src/d/actor/d_a_ff.cpp
```diff
@@ -1,55 +1,361 @@
...
+        mDoExt_modelUpdate(i_this->mpModel[0]);
+        if (i_this->mGlowScale > 0.01f) {
+            mDoMtx_YrotM(*calc_mtx, i_this->mLiveTimer << 8);
```
LagoLunatic, line 30: If it still matches this makes more sense as multiplication
```suggestion
            mDoMtx_YrotM(*calc_mtx, i_this->mLiveTimer * 0x100);
```

src/d/actor/d_a_ff.cpp
```diff
@@ -1,55 +1,361 @@
...
+            MtxScale(i_this->mGlowScale, i_this->mGlowScale * i_this->mGlowScaleY,
+                     i_this->mGlowScale, true);
+            i_this->mpModel[1]->setBaseTRMtx(*calc_mtx);
+            i_this->mBrkAnm[1]->play();
+            i_this->mBrkAnm[1]->entry(i_this->mpModel[1]->getModelData());
+            mDoExt_modelUpdate(i_this->mpModel[1]);
+        }
+        g_dComIfG_gameInfo.drawlist.setOpaList();
+        g_dComIfG_gameInfo.drawlist.setXluList();
```
LagoLunatic, line 39:
```suggestion
        dComIfGd_setList();
```

src/d/actor/d_a_ff.cpp
```diff
@@ -1,55 +1,361 @@
...
+    }
+    return;
```
LagoLunatic, line 41:
```suggestion
```

src/d/actor/d_a_ff.cpp
```diff
@@ -1,55 +1,361 @@
...
+        i_this->mScaleTarget = cM_ssin(i_this->mLiveTimer * 1000) * 0.15f * 0.25f + 0.225f;
+    } else {
+        i_this->mScaleTarget = 0.0f;
+    }
+    cLib_addCalc2(&i_this->mScale, i_this->mScaleTarget, 0.1f, 0.05f);
+    cVar1 = i_this->mState;
+    switch (cVar1) {
+    case 0:
+        chk.GetPointP()->set(i_this->current.pos.x, i_this->current.pos.z + 250.0f,
+                             i_this->current.pos.y);
```
LagoLunatic, line 112: Debug map says SetPos was used, so try something closer to this pattern, might fix the regalloc too
```suggestion
        Vec pos;
        pos = i_this->current.pos;
        pos.y += 250.0f;
        chk.SetPos(&pos);
```

PR #778, opened 2025-05-31
Automatically generate dCcD_Src structs for most files

ammaraskar:
These `dCcD_Src*` structs are a pain to write out by hand so I scripted to generate most of them automatically.

I used a ghidra script that finds all the `dCcD_Cps::Set` calls etc and then finds xrefs to them. It then figures out what symbol was passed in.

A script then loads up the data from those `src` symbols and generates the C code to initialize the struct. If the symbol had a `$` it adds it to the top of the method, otherwise it adds it to the top of the file after the includes.

### Validation
I'm trying to validate that all these are correct. It looks like `ninja baseline` and `ninja changes` doesn't report a change in `matched_data`. Not sure if that corresponds to  something else. There also doesn't seem to be a symbol-level report that `objdiff` generates by default that I could use.
 
Obviously I spot checked a few of the symbols in objdiff to make sure they match 100% but that isn't fully conclusive. I'm open to any ideas if you have more familiarity with objdiff before I go digging into its internals.

### Scripts
Would it make sense to commit the ghidra and python templating script I used to generate these in the `tools/` folder even though it's just going to be a one-time run of it? (Source code for both available here in this gist for now:  https://gist.github.com/ammaraskar/bac2dd8c14b01ec276360052c7608a39)

### DIFF
--- a/include/d/actor/d_a_daiocta.h
+++ b/include/d/actor/d_a_daiocta.h
@@ -4,6 +4,8 @@
 #include "f_op/f_op_actor.h"
 
 class J3DNode;
+class dCcD_SrcSph;
+class dCcD_SrcCps;
 
 class daDaiocta_c : public fopAc_ac_c {
 public:
@@ -64,6 +66,9 @@ public:
     void createInit();
     cPhs_State _create();
     bool _delete();
+
+    static const dCcD_SrcSph m_sph_src;
+    static const dCcD_SrcCps m_cps_src;
 
 public:
     /* 0x290 */ u8 field_0x290[0x574 - 0x290];
--- a/include/d/actor/d_a_fallrock.h
+++ b/include/d/actor/d_a_fallrock.h
@@ -3,6 +3,8 @@
 
 #include "f_op/f_op_actor.h"
 
+class dCcD_SrcCyl;
+
 class daFallRock_c : public fopAc_ac_c {
 public:
     inline cPhs_State create();
@@ -12,6 +14,8 @@ public:
     void CreateHeap();
     void set_mtx();
     void setParticle(int, cXyz*);
+
+    static const dCcD_SrcCyl m_cyl_src;
 
 public:
     /* Place member variables here */
--- a/include/d/actor/d_a_fm.h
+++ b/include/d/actor/d_a_fm.h
@@ -116,6 +116,9 @@ public:
     cPhs_State _create();
     bool _delete();
 
+    static const dCcD_SrcSph m_sph_src;
+    static const dCcD_SrcCyl m_cyl_src;
+
 public:
     /* 0x2AC */ u8 m2AC[0x2C5 - 0x2AC];
     /* 0x2C5 */ s8 m2C5;
--- a/include/d/actor/d_a_npc_so.h
+++ b/include/d/actor/d_a_npc_so.h
@@ -4,6 +4,7 @@
 #include "f_op/f_op_actor.h"
 
 class J3DNode;
+class dCcD_SrcSph;
 
 class daNpc_So_c : public fopAc_ac_c {
 public:
@@ -125,6 +126,8 @@ public:
     void cutUnequipStart();
     void initCam();
     void moveCam();
+
+    static const dCcD_SrcSph m_sph_src;
 
 public:
     /* Place member variables here */
--- a/include/d/actor/d_a_obj_canon.h
+++ b/include/d/actor/d_a_obj_canon.h
@@ -4,6 +4,7 @@
 #include "f_op/f_op_actor.h"
 
 class J3DNode;
+class dCcD_SrcSph;
 
 class daObj_Canon_c : public fopAc_ac_c {
 public:
@@ -40,6 +41,8 @@ public:
     void getArg();
     cPhs_State _create();
     bool _delete();
+
+    static const dCcD_SrcSph m_sph_src;
 
 public:
     /* Place member variables here */
--- a/include/d/actor/d_a_obj_ikada.h
+++ b/include/d/actor/d_a_obj_ikada.h
@@ -80,6 +80,8 @@ public:
     cPhs_State _create();
     bool _delete();
 
+    const static dCcD_SrcSph m_sph_src;
+
 public:
     /* 0x0290 */ u32 mType;
     /* 0x0294 */ u32 m0294;
--- a/include/d/actor/d_a_obj_warpt.h
+++ b/include/d/actor/d_a_obj_warpt.h
@@ -3,6 +3,8 @@
 
 #include "f_op/f_op_actor.h"
 
+class dCcD_SrcCyl;
+
 class daObj_Warpt_c : public fopAc_ac_c {
 public:
     enum Proc_e {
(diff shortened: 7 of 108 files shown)


LagoLunatic:
> I used a ghidra script that finds all the `dCcD_Cps::Set` calls etc and then finds xrefs to them. It then figures out what symbol was passed in.

Our Ghidra server has an issue where a lot of the data is corrupted and offset by one or two bytes compared to the symbols. So wouldn't doing this via a Ghidra script result in bad data? It doesn't happen to all of the data though, so I'm not sure how to check if any of these are incorrect without looking through every single one by hand.

> I'm trying to validate that all these are correct. It looks like `ninja baseline` and `ninja changes` doesn't report a change in `matched_data`. Not sure if that corresponds to something else. There also doesn't seem to be a symbol-level report that `objdiff` generates by default that I could use.
> 
> Obviously I spot checked a few of the symbols in objdiff to make sure they match 100% but that isn't fully conclusive. I'm open to any ideas if you have more familiarity with objdiff before I go digging into its internals.

objdiff-cli handles data kind of weirdly. it only considers "matched data" to be entire sections. So if a single symbol is 100% matched, but there are any other symbols in the same section as it that aren't matched, it won't notice any change in matched data %. So I don't think objdiff-cli will be helpful here, unfortunately. Might be possible to make a fork of it to check somehow though.

ammaraskar:
> So wouldn't doing this via a Ghidra script result in bad data?

Yeah I noticed that, I only grabbed the src symbol names from ghidra and then pulled the actual data for each symbol from the `.s` files.

> Might be possible to make a fork of it to check somehow though.

Gotcha, I'll take a look at modifying objdiff for this.

ammaraskar:
> Haven't looked into it in detail, but those might be worth checking out.

Oh, these look like ones where the `dCcD_Src` structs are defined inside of a namespace, which I currently don't handle. For example, `d_a_daiocta` has `daDaiocta_c::m_sph_src` but I generated it as `m_sph_src` so objdiff doesn't pick up on the symbol name association.

Let me fix those by hand real quick.

ammaraskar:
Okay with those changes and your objdiff modification I do see 97 additional data matches. Let me see if I can hack objdiff to make sure those individual symbols are matching at 100% as well.

ammaraskar:
Oh yeah if you get rid of the continue in the same part of `report.rs`:

```diff
--- a/objdiff-cli/src/cmd/report.rs
+++ b/objdiff-cli/src/cmd/report.rs
@@ -234,10 +234,8 @@ fn report_object(
         match section.kind {
             SectionKind::Data | SectionKind::Bss => {
                 measures.total_data += section.size;
-                if section_match_percent == 100.0 {
-                    measures.matched_data += section.size;
-                }
-                continue;
+                measures.matched_data += 
+                    (section.size as f32 * section_match_percent / 100.0) as u64;
             }
             _ => {}
         }
```

then objdiff puts the data symbols in `report.json` with their `fuzzy_match_percent`:

```json
                {
                    "name": "m_sph_src__13daObj_Ikada_c",
                    "size": "64",
                    "fuzzy_match_percent": 100.0,
                    "metadata": {
                        "demangled_name": "daObj_Ikada_c::m_sph_src",
                        "virtual_address": "36"
                    },
                    "address": "36"
                },
```

Let me write a quick script to make sure all the objects we added match 100% with that.

ammaraskar:
Okay I used this regex `dCcD_Src[A-Z][a-z][a-z] (\S*)(?:;| )` to grab all the newly added src objects from https://patch-diff.githubusercontent.com/raw/zeldaret/tww/pull/778.diff

and then used this Python script to make sure all the symbols there exist in the report and are 100% fuzzy matching according to the report.json

<details>

<summary>Python verification script</summary>

```python
import json

json_path = r'report.json'

functions_to_check = {'head_co_sph_src', 'body_at_cyl_src', 'utiwa_sph_src', 'tama_at_co_sph_src', 'hand_cyl_src', 'tg_sph_src', 'kamen_sph_src', 'cyl_check_src', 'daFallRock_c::m_cyl_src', 'hahen_sph_src', 'sph_check_src', 'chest_sph_src', 'l_at_cps_src', 'body_tg_sph_src', 'bero_sph_src', 'defence_sph_src', 'body_cyl_src', 'cyl_src', 'eye_sph_src', 'body_co_cyl_src', 'wepon_sph_src', 'wing_co_cyl_src', 'canon_cyl_src', 'tg_cyl_src', 'fire_sph_src', 'daOship_c::m_cyl_src', 'wind_at_sph_src', 'daObj_Canon_c::m_sph_src', 'daObj_Ikada_c::m_sph_src', 'core_cyl_src', 'daNpc_So_c::m_sph_src', 'tg_hit_puropera_co_cyl', 'wind_co_cyl_src', 'sph2_src', 'core_sph_src', 'body_at_sph_src', 'bm_sph_src', 'cc_sph_src', 'eff_sph_src', 'wepon2_sph_src', 'daPz_c::m_cyl_src', 'cc_cyl_src', 'daDaiocta_c::m_cps_src', 'head_tg_sph_src', 'daSaku_c::m_cyl_src', 'hs_sph_src', 'weapon_co_sph_src', 'ball_co_sph_src', 'bero_co_sph_src', 'body_co_cyl', 'beam_sph_src', 'daFm_c::m_sph_src', 'himo_cyl_src', 'co_sph_src', 'body_co_sph_src', 'daObj_Warpt_c::m_cyl_huta_src', 'daDaiocta_c::m_sph_src', 'kantera_co_sph_src', 'l_cyl_src', 'canon_cyl2_src', 'tama_tg_co_sph_src', 'finger_sph_src', 'l_sph_head_src', 'body_sph_src', 'wind_cc_sph_src', 'at_hit_puropera_co_cyl', 'damage_ball_co_sph_src', 'l_wind_cyl_src', 'cps_src', 'at_sph_src', 'l_cps_src_body', 'foot_cc_sph_src', 'daFm_c::m_cyl_src', 'co_cyl_src', 'tosaka_tg_sph_src', 'mesh_cc_sph_src', 'head_sph_src', 'l_cps_src', 'cps2_src', 'body_cc_sph_src', 'sph_src', 'p_co_cyl_src', 'foot_co_sph_src', 'head_at_sph_src', 'sibuki_sph_src', 'l_cps_src_huta', 'l_sph_src', 'daObj_Warpt_c::m_cyl_body_src'}

def load_json(path):
    with open(path, 'r', encoding='utf-8') as f:
        return json.load(f)

def check_functions(json_data, functions):
    missing = []
    not_100 = []
    for unit in json_data.get("units", []):
        for func in unit.get("functions", []):
            # Try both mangled and demangled names
            name = func.get("name", "")
            name = name.split("$")[0]
            demangled = func.get("metadata", {}).get("demangled_name", "")
            if name in functions or demangled in functions:
                if func.get("fuzzy_match_percent", 0) != 100.0:
                    not_100.append(name or demangled)
                functions = [f for f in functions if f != name and f != demangled]
    missing = functions
    return not_100, missing

def main():
    data = load_json(json_path)
    not_100, missing = check_functions(data, functions_to_check)
    if not_100:
        print("Functions with fuzzy_match_percent != 100.0:", not_100)
    if missing:
        print("Functions not found in the JSON:", missing)
    if not not_100 and not missing:
        print("All functions found and have fuzzy_match_percent == 100.0")

if __name__ == "__main__":
    main()
```

</details>

LagoLunatic:
Nice!

Issue #181, opened 2023-09-28
d_a_tag_attention


magcius:
AttTag, AttTagB

Skelp:
Working on it.

PR #784, opened 2025-06-03
#181: d_a_tag_attention

Skelp:
Most things matched, still working on the rest, hence a draft-only PR. Since this is my first contribution to the project, any advice on matching the remaining bits are highly appreciated!

/edit:
After all the extremely helpful comments, the TU is now finally 100% matched. It is set to `Equivalent` due to weak function ordering issues but otherwise functionally identical.
As I don't have a copy of D44J01 (yet), I can't check out the mismatch seen in decomp-dev's checks.


src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +40,120 @@ static dCcD_SrcSph sph_check_src = {
...
 }
 
 /* 00000340-00000480       .text _execute__Q214daTagAttention5Act_cFv */
 bool daTagAttention::Act_c::_execute() {
     /* Nonmatching */
+    mSph.SetC(current.pos);
+    mSph.SetR(scale.x * 100.0f);
+    dComIfG_Ccsp()->Set(&mSph);
+    m_b0x290 = true;
+    int iVar1 = daObj::PrmAbstract(this, PRM_1_W, PRM_1_S);
```
ammaraskar: In the debug map, this `PrmAbstract` call is under `prm_get_Type__Q214daTagAttention5Act_cCFv` which demangles to `daTagAttention::Act_c::prm_get_Type() const`

Those are the empty functions in the header, they will get inlined here so put this `PrmAbstract` call in there and use that `prm_get_Type` function instead.

See [the tutorial on debug maps](https://github.com/zeldaret/tww/blob/main/docs/decompiling.md#reading-rel-debug-maps) and [this example of `PrmAbstract`](https://github.com/zeldaret/tww/blob/4bf91cbfea65258b917cfc5dd60f909b2c9b715d/include/d/actor/d_a_obj_correct.h#L19-L38) for details

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -19,13 +22,13 @@ static dCcD_SrcSph sph_check_src = {
         /* SrcObjTg  SPrm    */ 0,
         /* SrcObjCo  SPrm    */ cCcD_CoSPrm_Set_e | cCcD_CoSPrm_IsOther_e | cCcD_CoSPrm_NoCrr_e,
         /* SrcGObjAt Se      */ 0,
-        /* SrcGObjAt HitMark */ dCcG_AtHitMark_None_e,
-        /* SrcGObjAt Spl     */ dCcG_At_Spl_UNK0,
+        /* SrcGObjAt HitMark */ 0,
```
ammaraskar: Is there a reason you changed this? Those enum values here before are equivalent to 0

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +40,120 @@ static dCcD_SrcSph sph_check_src = {
...
+bool daTagAttention::Act_c::chk_inside(cXyz* pos) const {
+    if (!m_b0x290){
+        return false;
+    }
+
+    cXyz plyrToObjVec = dComIfGp_getPlayer(0)->current.pos - current.pos;
+    if (subtype == 0){
+        // spherical collision check
+        f32 distance = plyrToObjVec.getSquareMag();
+        distance = std::sqrtf(distance);
```
ammaraskar: Debug map suggests this is a call to `cXyz.abs()`, which internally does the `getSquareMag` and `sqrt` part for you.

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +40,120 @@ static dCcD_SrcSph sph_check_src = {
...
+        
+        if (distance > scale.x * 100.0f) {
+            return false;
+        }
+    }else{
+        // box collision check
+        s16 yRotAngle = current.angle.y;
+        if (yRotAngle != 0){
+            u32 sinCosTableIndex = yRotAngle >> (jmaSinShift & 0xFFFF);
```
ammaraskar: This is `cM_ssin`

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +40,120 @@ static dCcD_SrcSph sph_check_src = {
...
+        }
+    }else{
+        // box collision check
+        s16 yRotAngle = current.angle.y;
+        if (yRotAngle != 0){
+            u32 sinCosTableIndex = yRotAngle >> (jmaSinShift & 0xFFFF);  
+            plyrToObjVec.z = 
+                plyrToObjVec.x * jmaSinTable[sinCosTableIndex] +
+                plyrToObjVec.z * jmaCosTable[sinCosTableIndex];
+            plyrToObjVec.x =
```
ammaraskar: `cM_scos`

ammaraskar (commented):
Using the correct inline functions might get stuff to match up a little better, I'd encourage you to take a look at the tutorial section on them and get a feel for reading the debug maps.

src/d/actor/d_a_tag_attention.cpp
Skelp: I made my changes long before the auto-generated structs came in two days ago - seems like I merged a bit too aggressively, I'll revert to the auto-generated values.

src/d/actor/d_a_tag_attention.cpp
Skelp: I've struggled with this one for a while until I dug out PrmAbstract and how it was used. I'm surprised I got it as right as I did without debug maps, I'll definitely shift my focus on them for now, making your suggested changes as I go. Thank you!

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +39,118 @@ static dCcD_SrcSph sph_check_src = {
...
+        // box collision check
+        s16 yRotAngle = current.angle.y;
+        if (yRotAngle != 0){
+            plyrToObjVec.z = 
+                plyrToObjVec.x * cM_ssin(yRotAngle) +
+                plyrToObjVec.z * cM_scos(yRotAngle);
+            plyrToObjVec.x = 
+                plyrToObjVec.x * cM_scos(yRotAngle) -
+                plyrToObjVec.z * cM_ssin(yRotAngle);
+        }
```
WilliamArnett: There is a regalloc error unless you do the check with `current.angle.y`.
Inside the conditional, you need to assign your first statement to a temporary variable. That way, the second statement doesn't read an incorrect value for `plyrToObjVec.z`

WilliamArnett:
```suggestion
        if (current.angle.y){
            s16 yRotAngle = current.angle.y;
            f32 temp = 
                plyrToObjVec.x * cM_ssin(yRotAngle) +
                plyrToObjVec.z * cM_scos(yRotAngle);
            plyrToObjVec.x = 
                plyrToObjVec.x * cM_scos(yRotAngle) -
                plyrToObjVec.z * cM_ssin(yRotAngle);
            plyrToObjVec.z = temp;
        }
```

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +39,118 @@ static dCcD_SrcSph sph_check_src = {
     },
 };
 
+bool daTagAttention::Act_c::chk_inside(cXyz* pos) const {
+    if (!m_b0x290){
+        return false;
+    }
+
+    cXyz plyrToObjVec = dComIfGp_getPlayer(0)->current.pos - current.pos;
```
WilliamArnett:
```suggestion
            fopAc_ac_c* player = dComIfGp_getPlayer(0);
            cXyz plyrToObjVec = player->current.pos-current.pos;
```
This needs to be split up into two different statements in order to match the assembly.

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +39,118 @@ static dCcD_SrcSph sph_check_src = {
     },
 };
 
+bool daTagAttention::Act_c::chk_inside(cXyz* pos) const {
```
WilliamArnett: This entire function actually needs to be put into the header instead of the cpp file. This will make your .rodata align, and also improve your weak function ordering.

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +39,118 @@ static dCcD_SrcSph sph_check_src = {
...
+            return false;
+        }
+        curScale = scale.y;
+        if ((plyrToObjVec.y < -curScale * 100.0f) || (curScale * 100.0f > plyrToObjVec.y)){
+            return false;
+        }
+        curScale = scale.z;
+        if ((plyrToObjVec.z < -curScale * 100.0f) || (curScale * 100.0f > plyrToObjVec.z)){
+            return false;
+        }
```
WilliamArnett:
```suggestion
            plyrToObjVec.z = temp;
        }

                if ((plyrToObjVec.x < -scale.x * 100.0f) || (plyrToObjVec.x > scale.x * 100.0f)){
                    return false;
                }
                if ((plyrToObjVec.y < -scale.y * 100.0f) || (plyrToObjVec.y > scale.y * 100.0f)){
                    return false;
                }
                if ((plyrToObjVec.z < -scale.z * 100.0f) || (plyrToObjVec.z > scale.z * 100.0f)){
                    return false;
                }
```
This matches if you don't use the `curScale` variable and just pass the values directly. You also need to change the order of operands in each second comparison.

src/d/actor/d_a_tag_attention.cpp
LagoLunatic: It is a weak function, but since it's so big it would be better kept in the cpp file instead of the header. You can mark it as weak by putting `inline bool chk_inside(cXyz*) const;` in the class in the header and then keeping the actual function here in the cpp. (The `inline` keyword won't actually inline it because it's so big, but it still may have an effect on rodata and weak function ordering.)

LagoLunatic: Actually nevermind, this function seems to have its own text split so I guess this one actually *does* go in the header. Usually big weak functions are in the same text split as the rest of the cpp functions, but this must be an exception for some reason.

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +39,118 @@ static dCcD_SrcSph sph_check_src = {
...
+            if (bVar2 == FALSE){
+                m_b0x290 = false;
+            }
+        }
+    }
+
+    // TODO for matching: implement chk_inside and dComIfGp_att_Look2RequestF
+    bool wasHit = chk_inside(&current.pos);
+    if (wasHit){
+        dComIfGp_att_Look2RequestF(this, 0x6000, 1);
```
WilliamArnett, line 68: The function this wraps is in `d_attention.h`. on line 263. I believe it should be
```
    void Look2RequestF(fopAc_ac_c* param_1, s16 param_2, int param_3) {mLook[1].requestF(param_1,param_2,param_3);}

```

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +39,118 @@ static dCcD_SrcSph sph_check_src = {
...
+    }else{
+        iVar1 = prm_get_Type();
+        if (iVar1 == 2){
+            iVar1 = prm_get_swSave();
+            BOOL bVar2 = dComIfGs_isSwitch(iVar1, home.roomNo);
+            if (bVar2 == FALSE){
+                m_b0x290 = false;
+            }
+        }
+    }
```
WilliamArnett:
```suggestion
    if (prm_get_Type() == 1){
        if (fopAcM_isSwitch(this, prm_get_swSave()) == FALSE){
            m_b0x290 = false;
        }
    }else if (prm_get_Type() == 2){
        if (fopAcM_isSwitch(this, prm_get_swSave()) != FALSE){
            m_b0x290 = false;
        }
    }
```

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -37,42 +39,118 @@ static dCcD_SrcSph sph_check_src = {
...
+
+    // TODO for matching: implement chk_inside and dComIfGp_att_Look2RequestF
+    bool wasHit = chk_inside(&current.pos);
+    if (wasHit){
```
WilliamArnett:
```suggestion
    cXyz unused;
    if (chk_inside(&unused)){
```
For this to match, you need to pass an uninitialized `cXyz` into the function

WilliamArnett (changes requested):

include/d/actor/d_a_tag_attention.h
```diff
@@ -1,20 +1,85 @@
...
+                        s16 yRotAngle = current.angle.y;
+                        f32 temp = 
+                            plyrToObjVec.x * cM_ssin(yRotAngle) +
+                            plyrToObjVec.z * cM_scos(yRotAngle);
+                        plyrToObjVec.x = 
+                            plyrToObjVec.x * cM_scos(yRotAngle) -
+                            plyrToObjVec.z * cM_ssin(yRotAngle);
+                        plyrToObjVec.z = temp;
+                    }
+                }
```
WilliamArnett:
```suggestion
                if (current.angle.y){
                    s16 yRotAngle = current.angle.y;
                    f32 temp = 
                        plyrToObjVec.x * cM_ssin(yRotAngle) +
                        plyrToObjVec.z * cM_scos(yRotAngle);
                    plyrToObjVec.x = 
                        plyrToObjVec.x * cM_scos(yRotAngle) -
                        plyrToObjVec.z * cM_ssin(yRotAngle);
                    plyrToObjVec.z = temp;
                }
```
I accidentally selected the wrong lines for my suggestion. It should look like this 👍

WilliamArnett (changes requested):

include/d/actor/d_a_tag_attention.h
```diff
@@ -1,20 +1,82 @@
...
     
         cPhs_State _create();
         bool _execute();
-    
+
+        enum Prm_e {
+            PRM_1_W = 0x02,
+            PRM_1_S = 0x08,
+            PRM_2_W = 0x08,
+            PRM_2_S = 0x00,
```
LagoLunatic: We know the official names from the inlines above, so we can replace these placeholders
```suggestion
            PRM_TYPE_W = 0x02,
            PRM_TYPE_S = 0x08,
            PRM_SWSAVE_W = 0x08,
            PRM_SWSAVE_S = 0x00,
```

include/d/actor/d_a_tag_attention.h
```diff
@@ -1,20 +1,82 @@
...
+        };
+
     public:
         /* Place member variables here */
```
LagoLunatic: Can remove the placeholder line
```suggestion
```

src/d/actor/d_a_tag_attention.cpp
```diff
@@ -6,8 +6,10 @@
 #include "d/actor/d_a_tag_attention.h"
 #include "d/d_procname.h"
 #include "d/d_priority.h"
-#include "d/d_cc_d.h"
+#include "f_op/f_op_actor_mng.h"
+#include "d/d_com_inf_game.h"
 
+/* ram:8049162C-ram:804917AF  */
```
LagoLunatic: RELs don't get loaded at a fixed RAM address. These addresses are just for our Ghidra project and aren't useful in the game itself
```suggestion
```

PR #788, opened 2025-06-05
98% matching d_a_dk

ammaraskar:
Weirdly enough there seems to be an unused `d_a_dk::@4026` float with a `-1.0` value. At least it seems to be dead based on ghidra xrefs and the assembly.

And this TU needs a string data ordering fix to make `"Dk"` appear before `"Bdk"`


ammaraskar:
Taking a look at the demo build, and actually it might be helpful for matching up the remaining functions as well.

include/d/actor/d_a_dk.h
```diff
@@ -1,23 +1,65 @@
...
+#include "d/d_bg_s_acch.h"
+#include "d/d_particle.h"
 #include "f_op/f_op_actor.h"
+#include "m_Do/m_Do_hostIO.h"
 
-struct tail_s {
-    
-};
+class tail_s {
+    public:
```
LagoLunatic: I think names ending in `_s` and `_struct` are intended to be structs, while ones ending in `_c` and `_class` are intended to be classes (though it doesn't make a practical difference besides default visibility)
```suggestion
struct tail_s {
```

include/d/actor/d_a_dk.h
```diff
@@ -1,23 +1,65 @@
...
+    /* 0x000 */ J3DModel* models[9];
+    /* 0x024 */ cXyz field_0x024[10];
+    /* 0x09C */ csXyz field_0x09C[10];
+    /* 0x0D8 */ cXyz field_0x0D8[10];
+    /* 0x150 */ cXyz field_0x150[2];
+    /* 0x168 */ csXyz field_0x168;
+    /* 0x16E */ u8 field_0x16E[0x170 - 0x16E];
```
LagoLunatic: Can remove this, alignment padding
```suggestion
```

include/d/actor/d_a_dk.h
```diff
@@ -1,23 +1,65 @@
...
 
-class daDk_HIO_c {
+    dk_class() : field_0xACC(1, 1, 0, 0) {
+
+    }
+
+    /* 0x290 */ u8 field_0x290[0x2AC - 0x290];
+    /* 0x2AC */ request_of_phase_process_class mPhs;
+    /* 0x2B4 */ u8 field_0x2B4;
+    /* 0x2B5 */ u8 field_0x2B5[0x2B8 - 0x2B5];
```
LagoLunatic:
```suggestion
```
Same with the rest of these padding fields in this header, except for field_0xAC4 and field_0xAEC

include/d/actor/d_a_dk.h
```diff
@@ -1,23 +1,65 @@
...
+
+class daDk_HIO_c : public JORReflexible {
 public:
     daDk_HIO_c();
+    virtual ~daDk_HIO_c() {}
 
-public:
-    /* Place member variables here */
-};
+    /* 0x04 */ s8 field_0x04;
```
LagoLunatic: This field in all HIO subclasses is called "No" (known from TP debug)
```suggestion
    /* 0x04 */ s8 mNo;
```

include/d/actor/d_a_dk.h
ammaraskar: Aah that makes sense, changed.

include/d/actor/d_a_dk.h
ammaraskar: Done, I always forget to do this after exporting the classes from ghidra.

PR #791, opened 2025-06-08
Add a few more matching functions in f_op_msg_mng

ammaraskar:
Just fixed a few easy non-matching functions in f_op_msg_mng

### DIFF
--- a/src/f_op/f_op_msg_mng.cpp
+++ b/src/f_op/f_op_msg_mng.cpp
@@ -405,7 +405,7 @@ u32 fopMsgM_searchMessageNumber(u32 msgNo) {
     msgGet.mMsgIdx = 0;
     msgGet.mGroupID = 0;
     msgGet.mMsgID = 0;
-    msgGet.mResMsgIdx = 0;  
+    msgGet.mResMsgIdx = 0;
 
     for(u32 i = msgNo & 0xFFFF; i < 0xFFFF; i++) {
         mesg_header* header = msgGet.getMesgHeader(i);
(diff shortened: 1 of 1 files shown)


src/f_op/f_op_msg_mng.cpp
```diff
@@ -922,7 +935,7 @@ void fopMsgM_outFontSet(J2DPicture* param_1, J2DPicture* param_2, s16* param_3,
     param_2->show();
     fopMsgM_blendDraw(param_1, fopMsgM_buttonTex[param_5]);
     fopMsgM_blendDraw(param_2, fopMsgM_buttonTex[param_5]);
-    if((0xA <= param_5 && param_5 <= 0xE) || param_5 == 0x15 || param_5 == 0x17) {
+    if((u8)(param_5 - 10) <= 3 || param_5 == 0x15 || param_5 == 0x17) {
```
LagoLunatic: Usually when you see `(unsigned cast)(value - base) <= range` it's a compiler optimization for a chain of OR statements checking contiguous values against a single variable (especially since this parameter had a high chance of being an enum originally). Does this still match?
```suggestion
    if(param_5 == 0xA || param_5 == 0xB || param_5 == 0xC || param_5 == 0xD || param_5 == 0x15 || param_5 == 0x17) {
```

src/f_op/f_op_msg_mng.cpp
```diff
@@ -962,7 +975,7 @@ void fopMsgM_outFontSet(J2DPicture* param_1, s16* param_2, u32 param_3, u8 param
 
     param_1->show();
     fopMsgM_blendDraw(param_1, fopMsgM_buttonTex[param_4]);
-    if((0xA <= param_4 && param_4 <= 0xE) || param_4 == 0x15 || param_4 == 0x17) {
+    if((u8)(param_4 - 10) <= 3 || param_4 == 0x15 || param_4 == 0x17) {
```
LagoLunatic: Same as above

src/f_op/f_op_msg_mng.cpp
ammaraskar: Aah yeah it still matches with that, thanks. Good to know.

PR #794, opened 2025-06-13
d_a_tag_ba1

DamienIsPoggers:
All functions matching except for <code>daTag_Ba1_c::_execute</code> which is mostly matching but has some weird gameinfo loads.


include/d/actor/d_a_tag_ba1.h
```diff
@@ -2,27 +2,40 @@
...
+    bool createInit();
     BOOL _draw();
     BOOL _execute();
     BOOL _delete();
     cPhs_State _create();
 
 public:
-    /* Place member variables here */
-};
+    /* 0x290 */ s16 field_0x290;
```
ammaraskar, line 18: Given the way this gets accessed, this should probably be typed as an array of a single short: `s16 field_0x290[1]`

src/d/actor/d_a_tag_ba1.cpp
```diff
@@ -7,54 +7,102 @@
...
-    /* Nonmatching */
+    return TRUE;
 }
 
 /* 00000290-00000340       .text _execute__11daTag_Ba1_cFv */
 BOOL daTag_Ba1_c::_execute() {
     /* Nonmatching */
+    int r3 = -1;
+    if (dComIfGp_event_runCheck() != FALSE && !eventInfo.checkCommandTalk())
+        r3 = dComIfGp_evmng_getMyStaffId("TagBa1", NULL, 0);
```
ammaraskar, line 75: You can omit the `NULL` and `0`, those are the default values if those arguments aren't passed in.

Issue #429, opened 2023-09-28
d_a_obj_nest


magcius:
MtoriSU

Skelp:
I'll take a crack at this, if there's no objections.

PR #798, opened 2025-06-20
#429: d_a_obj_nest

Skelp:
100% matching and linked
Needs some polish, and I believe some parts could be more verbose than needed. (i.e. a function that performs multiple lines of code that I did not know of and could find on my own)

/edit: well I guess D44J01 isn't agreeing...

### DIFF
--- a/src/d/actor/d_a_obj_nest.cpp
+++ b/src/d/actor/d_a_obj_nest.cpp
@@ -4,11 +4,11 @@
  */
 
 #include "d/actor/d_a_obj_nest.h"
+#include "d/res/res_mtorisu.h"
 #include "d/d_com_inf_game.h"
 #include "d/d_procname.h"
 #include "d/d_priority.h"
 #include "d/d_bg_w.h"
-
 namespace daObjNest{
     namespace{
         struct Attr_c {
@@ -49,8 +49,6 @@ namespace daObjNest{
 
 Mtx daObjNest::Act_c::M_tmp_mtx;
 const char daObjNest::Act_c::M_arcname[] = "MtoriSU";
-const int MTORISU_BDL_MTORISU = 0x4;
-const int MTORISU_DZB_MTORISU = 0x7;
 
 /* 00000078-0000012C       .text CreateHeap__Q29daObjNest5Act_cFv */
 BOOL daObjNest::Act_c::CreateHeap() {


src/d/actor/d_a_obj_nest.cpp
```diff
@@ -4,67 +4,166 @@
...
+    }
+}
+
+Mtx daObjNest::Act_c::M_tmp_mtx;
+const char daObjNest::Act_c::M_arcname[] = "MtoriSU";
 
 /* 00000078-0000012C       .text CreateHeap__Q29daObjNest5Act_cFv */
 BOOL daObjNest::Act_c::CreateHeap() {
-    /* Nonmatching */
+    J3DModelData* model_data = (J3DModelData*)dComIfG_getObjectRes(M_arcname, 0x4);
```
LagoLunatic: Res enum
```suggestion
    J3DModelData* model_data = (J3DModelData*)dComIfG_getObjectRes(M_arcname, MTORISU_BDL_MTORISU);
```

src/d/actor/d_a_obj_nest.cpp
```diff
@@ -4,67 +4,166 @@
...
 }
 
 /* 000001F0-000002E8       .text Mthd_Create__Q29daObjNest5Act_cFv */
 cPhs_State daObjNest::Act_c::Mthd_Create() {
-    /* Nonmatching */
+    fopAcM_SetupActor(this, daObjNest::Act_c);
+    
+    cPhs_State phase_state = dComIfG_resLoad(&mPhs, M_arcname);
+    if (phase_state == cPhs_COMPLEATE_e) {
+        phase_state = MoveBGCreate(M_arcname, 0x7, NULL, 0xAA0);
```
LagoLunatic: Res enum
```suggestion
        phase_state = MoveBGCreate(M_arcname, MTORISU_DZB_MTORISU, NULL, 0xAA0);
```

configure.py
```diff
@@ -1694,7 +1694,7 @@ def MatchingFor(*versions):
     ActorRel(NonMatching, "d_a_obj_msdan_sub"),
     ActorRel(NonMatching, "d_a_obj_msdan_sub2"),
     ActorRel(Matching,    "d_a_obj_mtest", extra_cflags=['-pragma "nosyminline on"']),
-    ActorRel(NonMatching, "d_a_obj_nest"),
+    ActorRel(Matching, "d_a_obj_nest"),
```
LagoLunatic: You can mark it as as matching for retail only so the build doesn't fail
```suggestion
    ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"), "d_a_obj_nest"),
```

src/d/actor/d_a_obj_nest.cpp
```diff
@@ -4,67 +4,168 @@
...
+        };
+
+        inline static const Attr_c& attr() { return L_attr; }
+    }
+}
+
+Mtx daObjNest::Act_c::M_tmp_mtx;
+const char daObjNest::Act_c::M_arcname[] = "MtoriSU";
+const int MTORISU_BDL_MTORISU = 0x4;
+const int MTORISU_DZB_MTORISU = 0x7;
```
LagoLunatic: There's no need to define these manually, I already generated the enums for all resource archives in the game. You just need to include the header `d/res/res_mtorisu.h` so that you have access to them, if you don't include the header then it won't compile.

Skelp: Right, your messages above the code suggestions were good hints for that, had I seen them and not accidentally skipped past them. :) Will fix these.

LagoLunatic: Yeah, maybe I should add a section about those enums to the tutorial.

Looks all good now!

LagoLunatic:
> @LagoLunatic, how did that happen? Didn't even touch that part of the codebase.

This is a compiler bug discovered recently where it will randomly fail to convert a pointer to a boolean a small percentage of the time, which breaks the match in that one function. It's not related to anything you did, just random chance.

PR #807, opened 2025-07-14
d_a_acorn_leaf

Kelebek1:


configure.py
```diff
@@ -1530,7 +1530,7 @@ def MatchingFor(*versions):
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_tornado", extra_cflags=['-pragma "nosyminline on"']),
     ActorRel(NonMatching, "d_a_warpf"),
     ActorRel(NonMatching, "d_a_wind_tag"),
-    ActorRel(NonMatching, "d_a_acorn_leaf"),
+    ActorRel(MatchingFor("D44J01", "GZLE01"), "d_a_acorn_leaf", extra_cflags=["-sym off"]),
```
LagoLunatic: The bot says it matches for JPN and PAL so you can probably just put `Matching` here for all versions.

Also, did you check if `'-pragma "nosyminline on"'` works? The debug map has multiple .text sections so sym off doesn't seem right, but either way it's not a big deal.

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -38,50 +45,193 @@ static dCcD_SrcCyl l_cyl_src = {
...
-    /* Nonmatching */
+    daPy_py_c* player = (daPy_py_c*)g_dComIfG_gameInfo.play.getPlayer(0);
+
+    fpc_ProcID sp14 = unk_424;
+    fopAc_ac_c* actor = fopAcM_Search(fpcSch_JudgeByID, &sp14);
+    
+    bool var_r29 = false;
+
+    if (actor != NULL) {
+        var_r29 = daObj::PrmAbstract<daTsubo::Act_c::Prm_e>(actor, daTsubo::Act_c::PRM_TYPE_1, daTsubo::Act_c::PRM_TYPE_1F);
```
LagoLunatic: This should use the param getter inline mentioned in the debug map:
```suggestion
        var_r29 = daTsubo::Act_c::prm_get_stick();
```
And the enums should be renamed `PRM_TYPE_1` -> `PRM_STICK_W` and `PRM_TYPE_1F` -> `PRM_STICK_S`, as this isn't related to the "type" param

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -90,8 +240,9 @@ static cPhs_State daAleaf_Create(void* i_this) {
 }
 
 /* 00000CF4-00000D24       .text daAleaf_Delete__FPv */
-static BOOL daAleaf_Delete(void*) {
-    /* Nonmatching */
+static BOOL daAleaf_Delete(void* i_this) {
+    dComIfG_resDeleteDemo(&((daAleaf_c*)i_this)->unk_290, daAleaf_c::m_arcname);
+    return TRUE;
```
LagoLunatic: The contents of this go in the `daAleaf_c::_delete` inline, which can be defined in the .cpp file right above this static func
```suggestion
    return ((daAleaf_c*)i_this)->_delete();
```

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -38,50 +45,193 @@ static dCcD_SrcCyl l_cyl_src = {
...
 /* 0000010C-00000258       .text CreateHeap__9daAleaf_cFv */
-void daAleaf_c::CreateHeap() {
-    /* Nonmatching */
+BOOL daAleaf_c::CreateHeap() {
+    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes(daAleaf_c::m_arcname, 7);
+#if VERSION == VERSION_DEMO
+    JUT_ASSERT(0x102, modelData != NULL);
+#else
+    JUT_ASSERT(0x106, modelData != NULL);
+#endif
```
LagoLunatic: You can condense this assert by using the VERSION_SELECT macro to change the line numbers only, same for the one below
```suggestion
    JUT_ASSERT(VERSION_SELECT(258, 262, 262, 262), modelData != NULL);
```

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -38,50 +45,193 @@ static dCcD_SrcCyl l_cyl_src = {
...
-    /* Nonmatching */
+static BOOL CheckCreateHeap(fopAc_ac_c* i_actor) {
+    return ((daAleaf_c*)i_actor)->CreateHeap();
 }
 
 /* 0000010C-00000258       .text CreateHeap__9daAleaf_cFv */
-void daAleaf_c::CreateHeap() {
-    /* Nonmatching */
+BOOL daAleaf_c::CreateHeap() {
+    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes(daAleaf_c::m_arcname, 7);
```
LagoLunatic: Autogenerated resource enum, same for the other one below
```suggestion
    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes(daAleaf_c::m_arcname, VIGAH_BDL_VIGAH);
```

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -38,50 +45,193 @@ static dCcD_SrcCyl l_cyl_src = {
...
+    unk_298->setBaseTRMtx(mDoMtx_stack_c::get());
 }
 
 /* 000008D0-00000C44       .text _execute__9daAleaf_cFv */
 bool daAleaf_c::_execute() {
-    /* Nonmatching */
+    daPy_py_c* player = (daPy_py_c*)g_dComIfG_gameInfo.play.getPlayer(0);
+
+    fpc_ProcID sp14 = unk_424;
+    fopAc_ac_c* actor = fopAcM_Search(fpcSch_JudgeByID, &sp14);
```
LagoLunatic, line 155: Simplifies with inline, same for the other one below
```suggestion
    fopAc_ac_c* actor = fopAcM_SearchByID(unk_424)
```

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -38,50 +45,193 @@ static dCcD_SrcCyl l_cyl_src = {
...
+    unk_298->setBaseScale(scale);
+    mDoMtx_trans(mDoMtx_stack_c::get(), current.pos.x, current.pos.y, current.pos.z);
+    mDoMtx_stack_c::ZXYrotM(current.angle.x, current.angle.y, current.angle.z);
+    unk_298->setBaseTRMtx(mDoMtx_stack_c::get());
 }
 
 /* 000008D0-00000C44       .text _execute__9daAleaf_cFv */
 bool daAleaf_c::_execute() {
-    /* Nonmatching */
+    daPy_py_c* player = (daPy_py_c*)g_dComIfG_gameInfo.play.getPlayer(0);
```
LagoLunatic:
```suggestion
    daPy_py_c* player = (daPy_py_c*)dComIfGp_getPlayer(0);
```

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -38,50 +45,193 @@ static dCcD_SrcCyl l_cyl_src = {
...
+        CreateInit();
+    }
+    return ret;
 }
 
 /* 00000848-000008D0       .text set_mtx__9daAleaf_cFv */
 void daAleaf_c::set_mtx() {
-    /* Nonmatching */
+    unk_298->setBaseScale(scale);
+    mDoMtx_trans(mDoMtx_stack_c::get(), current.pos.x, current.pos.y, current.pos.z);
```
LagoLunatic:
```suggestion
    mDoMtx_stack_c::trans(current.pos);
```

src/d/actor/d_a_acorn_leaf.cpp
```diff
@@ -38,50 +45,193 @@ static dCcD_SrcCyl l_cyl_src = {
...
 void daAleaf_c::CreateInit() {
-    /* Nonmatching */
+    fopAcM_SetMtx(this, unk_298->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -50.0f, 0.0f, -50.0f, 50.0f, 100.0f, 50.0f);
+    unk_29C.Init(255, 255, this);
+    unk_2D8.Set(l_cyl_src);
+    unk_2D8.SetStts(&unk_29C);
+    current.angle = home.angle;
+    unk_428 = -1;
+    unk_424 = -1;
```
LagoLunatic:
```suggestion
    unk_428 = fpcM_ERROR_PROCESS_ID_e;
    unk_424 = fpcM_ERROR_PROCESS_ID_e;
```

configure.py
Kelebek1: The pragma works too, but sym off seems cleaner and nicer. Didn't know the pragma is preferred.

LagoLunatic: Depends on the actor, C-style actors like almost all enemies usually have just a single .text section meaning sym off, but C++-style actors like this one usually have multiple .text sections meaning sym on (acorn_leaf has 5 in the debug map, 3 in the non-debug map)

src/d/actor/d_a_acorn_leaf.cpp
Kelebek1: I can't match the stack with this, so I left it.

LagoLunatic: It looks like the reason for the stack issues was that you were using the `getMagXZ` inline instead of the `absXZ` inline. (I don't know why we had `getMagXZ` in the decomp since it doesn't appear in any maps, I'm getting rid of it to prevent confusion.)

PR #812, opened 2025-07-22
d_a_coming2

Kelebek1:


src/d/actor/d_a_coming2.cpp
```diff
@@ -6,180 +6,607 @@
...
+    }
 }
 
 /* 00000164-00000200       .text get_speed__9daComing2FP10fopAc_ac_c */
-void daComing2::get_speed(fopAc_ac_c*) {
-    /* Nonmatching */
+f32 daComing2::get_speed(fopAc_ac_c* actor) {
+    const static u8 L_attr[] = {
+        0, 0, 0, 0xC, 0x2, 0x58, 0x00, 0x87, 0x05, 0x46,
+    };
```
LagoLunatic: I believe the symbol name `L_attr__Q29daComing225@unnamed@d_a_coming2_cpp@` suggests that this variable goes inside an unnamed namespace within the daComing2 namespace, not inside of this get_speed function.
So it would be like
```cpp
namespace daComing2 {
namespace {
const static u8 L_attr[] = {
...
```
probably at the top of the file with the other data.

include/d/actor/d_a_coming2.h
```diff
@@ -3,56 +3,113 @@
...
+    class _barrel_info_ {
+    public:
+        _barrel_info_() {}
+        ~_barrel_info_() {}
+        
+        /* 0x00 */ fpc_ProcID unk_00;
+        /* 0x04 */ s16 unk_04;
+        /* 0x06 */ u8 unk_06;
+        /* 0x08 */ cXyz unk_08;
+    }; // size = 0x14
```
LagoLunatic: Is explicitly specifying the ctor and dtor needed to match? If it's not, then I would assume these were intended to be plain old data structs by the original programmers (like `_start_stat_` is), and the ctor/dtor are compiler-generated:
```suggestion
    struct _barrel_info_ {
        /* 0x00 */ fpc_ProcID unk_00;
        /* 0x04 */ s16 unk_04;
        /* 0x06 */ u8 unk_06;
        /* 0x08 */ cXyz unk_08;
    }; // size = 0x14
```
(Same for `_start_buoy_` below)

src/d/actor/d_a_coming2.cpp
```diff
@@ -6,180 +6,607 @@
...
+    mDoMtx_stack_c::YrotM(shape_angle.y);
+    MTXCopy(mDoMtx_stack_c::get(), sp4C);
+
+    for (s32 i = 0; i < 2; i++) {
+        if (mStartBuoy[i].unk_00 == fpcM_ERROR_PROCESS_ID_e) {
+            cMtx_multVec(sp4C, &sp34[i], &sp28);
+            if (get_water_height(&sp28.y, &sp28) && check_in_large_sea(&sp28) && 
+                position_is_none_obj(&sp28) && no_ship_obstacle(&sp28)) {
+#if VERSION == VERSION_DEMO
+                fpc_ProcID barrel = daObjBarrel2::Act_c::make_coming(&sp28, 1, daObjBarrel2::Type_01_e, 0x3F, true, mStartStat.unk_00, daObjBuoyflag::Texture_e(1));
```
LagoLunatic: Instead of casting an int literal to an enum, it's better to create a placeholder name inside the enum itself so that when someone renames it later in a documentation pass intellisense will rename all instances of it. Placeholder name can be something like
```suggestion
                fpc_ProcID barrel = daObjBarrel2::Act_c::make_coming(&sp28, 1, daObjBarrel2::Type_01_e, 0x3F, true, mStartStat.unk_00, daObjBuoyflag::TEXTURE_UNK_01);
```
Also applies to other casts of Texture_e in this file, and Type_e too

include/d/actor/d_a_coming2.h
Kelebek1: They were required when I defined them, but maybe after adding cXyz they're auto-generated, I'll try remove them and see.

PR #813, opened 2025-07-23
d_a_coming3

Kelebek1:
Unfortunately un-matches d_a_coming2 due to the barrel param inlines.


include/SSystem/SComponent/c_xyz.h
```diff
@@ -10,7 +10,7 @@ struct cXy {
 };
 
 struct cXyz : Vec {
-    static const cXyz Zero;
+    static cXyz Zero;
```
LagoLunatic: This is supposed to be const since it's a constant zero.

src/d/actor/d_a_coming3.cpp
```diff
@@ -4,125 +4,471 @@
...
+    unk_440->calc();
 }
 
 /* 0000167C-000016F8       .text collision_init__Q29daComing35Act_cFv */
 void daComing3::Act_c::collision_init() {
-    /* Nonmatching */
+    mStts.Init(0xFF, 0xFF, this);
+    mCyl.Set(M_cyl_src);
+    mCyl.SetStts(&mStts);
+    mCyl.SetTgVec(cXyz::Zero);
```
LagoLunatic: You can get this to compile by casting away the const instead.
```suggestion
    mCyl.SetTgVec((cXyz&)cXyz::Zero);
```

tools/dump_cylinder.py
```diff
@@ -0,0 +1,556 @@
+"""
+Automatically writes source structs for colliders.
```
LagoLunatic: Let's remove this file from this PR and figure it out later

src/d/actor/d_a_coming3.cpp
```diff
@@ -4,125 +4,471 @@
...
+                current.pos.x,
+                waterHeight,
+                current.pos.z
+            );
+
+            if (unk_400 == fpcM_ERROR_PROCESS_ID_e) {
+                static s32 make_item_table[] = {
+                    dItem_BLUE_RUPEE_e, dItem_YELLOW_RUPEE_e,
+                };
+                unk_400 = daObjBarrel2::Act_c::make_coming(&sp18, fopAcM_GetRoomNo(this), daObjBarrel2::Type_e(1), make_item_table[get_challenge_id()], true, shape_angle.y, daObjBuoyflag::Texture_e(1));
```
LagoLunatic: These enums should use the enum entries instead of casting from int

PR #816, opened 2025-07-24
d_a_door10

Kelebek1:
Everything matching, but some weak functions out of order it seems.


src/d/actor/d_a_door10.cpp
```diff
@@ -6,170 +6,835 @@
...
+    u8 cVar1 = getType();
+    s32 ret;
+    s32 v = 2;
+
+    if (cVar1 == 4 || cVar1 == 5) {
+        ret = 1;
+    } else {
+        ret = cVar1 == 1 ? 2 : 0;
+    }
+    return ret;
```
LagoLunatic: I think Roeming's suggestion is probably how the original devs wrote this, not a ternary https://decomp.me/scratch/QstUs
```suggestion
    if (cVar1 == 4 || cVar1 == 5) {
        return 1;
    }
    if (cVar1 == 1) {
        return 2;
    }
    return 0;
```

src/d/actor/d_a_door10.cpp
```diff
@@ -6,170 +6,835 @@
...
 
 /* 00000788-000007CC       .text getSize2X__10daDoor10_cFv */
-void daDoor10_c::getSize2X() {
-    /* Nonmatching */
+f32 daDoor10_c::getSize2X() {
+    switch (getType()) {
+        case 3:
+            return 225.0f;
+    }
+    return 12100.0f;
```
LagoLunatic: These are square distances so this is more readable
```suggestion
            return 15.0f * 15.0f;
    }
    return 110.0f * 110.0f;
```

src/d/actor/d_a_door10.cpp
```diff
@@ -6,170 +6,835 @@
...
+
+            default:
+                if (dComIfGs_getKeyNum() == 0) {
+                    return;
+                }
+                break;
+        }
+    }
+
+    if (checkArea(getSize2X(), 12100.0f, 62500.0f)) {
```
LagoLunatic:
```suggestion
    if (checkArea(getSize2X(), 110.0f * 110.0f, 250.0f * 250.0f)) {
```

PR #822, opened 2025-07-28
daNpc_Cb1_c Almost OK

SuperDude88:
Almost matching (though will definitely have weak function ordering issues), still needs a bunch of cleanup with inlines/enums/etc


src/d/actor/d_a_npc_cb1.cpp
```diff
@@ -72,532 +88,2993 @@ static dCcD_SrcCyl l_wind_cyl_src = {
...
+
+    return TRUE;
 }
 
 /* 00000D9C-00001458       .text createHeap__11daNpc_Cb1_cFv */
-void daNpc_Cb1_c::createHeap() {
-    /* Nonmatching */
+BOOL daNpc_Cb1_c::createHeap() {
+    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes("Cb", CB_BDL_CB);
+    JUT_ASSERT(0x3A8, modelData != 0);
```
LagoLunatic:
```suggestion
    JUT_ASSERT(0x3A8, modelData != NULL);
```
These asserts can use defines in them and they get converted to their value in the string itself. Also applies to the other `modelData != 0` asserts in this TU.

src/d/actor/d_a_npc_cb1.cpp
```diff
@@ -72,532 +88,2993 @@ static dCcD_SrcCyl l_wind_cyl_src = {
...
+                        setNpcAction(&daNpc_Cb1_c::waitNpcAction, this);
+
+                        return TRUE;
+                    }
+
+                    temp2 = 0.0f;
+                    temp = current.angle.y;
+                }
+
+                if(dist2 < 160000.0f && cLib_distanceAngleS(shape_angle.y, temp) < 0x2000 && std::fabsf(fopAcM_searchPlayerDistanceY(this)) < 100.0f) {
```
LagoLunatic:
```suggestion
                if(dist2 < (400.0f * 400.0f) && cLib_distanceAngleS(shape_angle.y, temp) < 0x2000 && std::fabsf(fopAcM_searchPlayerDistanceY(this)) < 100.0f) {
```
Because this is a square dist it can be a bit more clear if the literal is written unsquared

src/d/actor/d_a_npc_cb1.cpp
```diff
@@ -72,532 +88,2993 @@ static dCcD_SrcCyl l_wind_cyl_src = {
...
+    else {
+        if(chkWallHit()) {
+            f32 temp = checkWallJump(*param_2);
+            if(temp >= 0.0f) {
+                setNpcAction(&daNpc_Cb1_c::jumpNpcAction, &temp);
+
+                return TRUE;
+            }
+
+            if(param_1 > 360000.0f) {
```
LagoLunatic: Here too (param_1 is dist2)

PR #824, opened 2025-07-29
d_a_majuu_flag

Kelebek1:


src/d/actor/d_a_majuu_flag.cpp
```diff
@@ -6,55 +6,794 @@
...
+
+#include "assets/l_flag02TEX.h"
+
+static Vec l_majuu_flag_pos[] = {
+    { 0.0f, -255.0f, 1300.0f },
+    {
+    0.0f,
+    -306.0f,
+    1040.0f
+    },
```
LagoLunatic: Maybe run "Format Selection" on these long arrays (l_majuu_flag_pos l_texCoord rel_pos_idx_tbl rel_pos_idx_tbl) to fix the weird whitespace

src/d/actor/d_a_majuu_flag.cpp
```diff
@@ -6,55 +6,794 @@
...
+    dKy_setLight_mine(mpTevStr);
+    GXClearVtxDesc();
+    GXSetVtxDesc(GX_VA_POS,GX_INDEX8);
+    GXSetVtxDesc(GX_VA_NRM,GX_INDEX8);
+    GXSetVtxDesc(GX_VA_TEX0,GX_INDEX8);
+    GXSetVtxAttrFmt(GX_VTXFMT0,GX_VA_POS, GX_CLR_RGBA,GX_F32, 0);
+    GXSetVtxAttrFmt(GX_VTXFMT0,GX_VA_NRM, GX_CLR_RGB,GX_F32, 0);
+    GXSetVtxAttrFmt(GX_VTXFMT0,GX_VA_TEX0, GX_CLR_RGBA,GX_F32, 0);
+    GXSetArray(GX_VA_POS, getPos(), sizeof(cXyz) * 1);
+    GXSetArray(GX_VA_NRM, getNrm(), sizeof(cXyz) * 1);
```
LagoLunatic, line 496: Is `* 1` required to match?

src/d/actor/d_a_majuu_flag.cpp
```diff
@@ -6,55 +6,794 @@
...
+    GXLoadPosMtxImm(*getMtx(), 0);
+    GXLoadNrmMtxImm(*getMtx(), 0);
+    GXSetCullMode(GX_CULL_BACK);
+    GXCallDisplayList(l_majuu_flagDL, 0x80);
+    GXSetCullMode(GX_CULL_FRONT);
+#if VERSION == VERSION_DEMO
+    GXSetArray(GX_VA_NRM, mpNrmArrBack[mCurArr], sizeof(cXyz) * 1);
+#else
+    GXSetArray(GX_VA_NRM, getNrmBack(), sizeof(cXyz) * 1);
+#endif
```
LagoLunatic: Where did `getNrmBack` come from? This inline doesn't seem to be in the debug map, but this also doesn't look like a genuine version difference between demo and retail since it's functionally equivalent, so I don't think there should be an ifdef here.

Kelebek1: There's regalloc differences without it.

src/d/actor/d_a_majuu_flag.cpp
Kelebek1: No, but since it's grabbing an array I wanted to show it's only using 1 element.

src/d/actor/d_a_majuu_flag.cpp
Kelebek1: All format opotions are disabled by the VSCode C++ engine being disabled by the repo, there is no way to format, at least for me.

LagoLunatic: That's because this repo uses the vscode-clangd extension instead of C++ tools.

src/d/actor/d_a_majuu_flag.cpp
LagoLunatic: I think it's a fakematch. Seems better to leave the regalloc and Nonmatching comment so that someone can try to figure out the real solution later.

Kelebek1: Why is it fake? Why can't it be something they added for retail?

LagoLunatic: Yeah, I guess it could be, but it seems unlikely since it doesn't change the functional equivalence at all, I'd expect that inline to exist for demo too since the other inlines like `getNrm` exist, and `setBackNrm` also exists.
I remember there was some weirdness with this same field in other classes like d_a_sail so I'll have to compare those later and see what's going on.

For now let's just rename it to `getNrmBack` -> `getBackNrm` to match `setBackNrm` and add a comment that this inline doesn't appear in the debug maps.

.clang-format
```diff
@@ -0,0 +1,74 @@
+---
```
LagoLunatic: Looks like you added this file on accident

src/d/actor/d_a_majuu_flag.cpp
```diff
@@ -4,72 +4,1092 @@
...
+#include "d/res/res_xhcf.h"
+#include "dolphin/gx/GXTexture.h"
+#include "f_op/f_op_actor_mng.h"
+#include "m_Do/m_Do_graphic.h"
+#include "m_Do/m_Do_hostIO.h"
+
+class daMajuu_Flag_HIO_c : public mDoHIO_entry_c {
+public:
+  daMajuu_Flag_HIO_c() {
+    mbUseToonTex = true;
```
LagoLunatic: The indentation in this whole file is wrong, we use 4 spaces not 2. I recommend reverting to before you ran that autoformatter.

.clang-format
Kelebek1: It's not by accident, I wanted some consistent formatting config in the repo. VSCode formatting is all wrong without something like this, like formatting to 2 spaces and stuff.

PR #829, opened 2025-08-02
d_a_obj_ice

Kelebek1:


src/d/actor/d_a_obj_ice.cpp
```diff
@@ -4,84 +4,545 @@
...
+        m458 = 0x5a;
+        JPABaseEmitter* emitter = dComIfGp_particle_set(
+            dPa_name::ID_COMMON_0464, &current.pos, NULL, &scale, 0xFF, NULL, -1, &tevStr.mColorK0);
+        if (emitter != NULL) {
+#if VERSION == VERSION_DEMO
+            JGeometry::TVec3<f32> s(scale.x, scale.y, scale.z);
+#else
+            JGeometry::TVec3<f32> s;
+            s.set(scale);
+#endif
```
LagoLunatic: The demo debug map suggests the two-line version was used there as well.
```suggestion
            JGeometry::TVec3<f32> s;
            s.set(scale);
```

src/d/actor/d_a_obj_ice.cpp
```diff
@@ -4,84 +4,545 @@
...
 void daObjIce_c::init_mtx() {
-    /* Nonmatching */
+    mModel->setBaseScale(scale);
+    set_mtx();
 }
 
 /* 000001C8-000001EC       .text solidHeapCB__10daObjIce_cFP10fopAc_ac_c */
-void daObjIce_c::solidHeapCB(fopAc_ac_c*) {
-    /* Nonmatching */
+int daObjIce_c::solidHeapCB(fopAc_ac_c* a_this) {
```
LagoLunatic:
```suggestion
BOOL daObjIce_c::solidHeapCB(fopAc_ac_c* a_this) {
```

include/d/actor/d_a_obj_ice.h
```diff
@@ -3,19 +3,30 @@
...
     inline void setEffectMtx();
 
-    void chk_appear();
+    u8 chk_appear();
     void set_mtx();
     void init_mtx();
-    void solidHeapCB(fopAc_ac_c*);
-    void create_heap();
-    void tg_hitCallback(fopAc_ac_c*, dCcD_GObjInf*, fopAc_ac_c*, dCcD_GObjInf*);
+    static int solidHeapCB(fopAc_ac_c*);
```
LagoLunatic:
```suggestion
    static BOOL solidHeapCB(fopAc_ac_c*);
```

src/d/actor/d_a_obj_ice.cpp
Kelebek1: It doesn't match.

LagoLunatic: No need to fakematch demo, it's better to leave it unmatched and maybe someone can find the real fix later.

Kelebek1: Why do you think it's fake? One wants to call the 3 float version, the other wants to call the TVec3 copy.

LagoLunatic: > Why do you think it's fake?

Because the demo debug maps say the following:
```
        7] __ct__Q29JGeometry8TVec3<f>Fv (func,weak) found in d_a_obj_ice.o 
        7] set<f>__Q29JGeometry8TVec3<f>Ffff (func,local) found in d_a_obj_ice.o 
        7] setGlobalDynamicsScale__14JPABaseEmitterFRCQ29JGeometry8TVec3<f> (func,weak) found in d_a_obj_ice.o 
         8] set<f>__Q29JGeometry8TVec3<f>FRCQ29JGeometry8TVec3<f> (func,local) found in d_a_obj_ice.o 
        7] setGlobalParticleScale__14JPABaseEmitterFRCQ29JGeometry8TVec3<f> (func,weak) found in d_a_obj_ice.o 
```
The demo debug map can't be wrong for the demo version. So it definitely initializes a TVec3 with no arguments, then calls set with 3 floats. It doesn't initialize a TVec3 with 3 floats.

src/d/actor/d_a_obj_ice.cpp
```diff
@@ -341,8 +341,7 @@ void daObjIce_c::wait_act_proc() {
 #if VERSION == VERSION_DEMO
             JGeometry::TVec3<f32> s(scale.x, scale.y, scale.z);
 #else
-            JGeometry::TVec3<f32> s;
-            s.set(scale);
+            JGeometry::TVec3<f32> s(scale);
```
LagoLunatic: Now I'm confused. What's the reasoning behind this change?

Kelebek1: Because it's more symmetrical and they both call set anyway.

PR #833, opened 2025-08-05
d_a_obj_mshokki

Kelebek1:


src/d/actor/d_a_obj_mshokki.cpp
```diff
@@ -4,62 +4,386 @@
...
+#if VERSION == VERSION_DEMO
+class daObjMshokki_HIO_c : public JORReflexible {
+public:
+    daObjMshokki_HIO_c();
+    virtual ~daObjMshokki_HIO_c() {}
+
+    void genMessage(JORMContext*);
+
+public:
+    /* 0x04 */ s8 m04;
```
LagoLunatic:
```suggestion
    /* 0x04 */ s8 mNo;
```
The s8 at the start of most HIO classes is called "No", based off a TP debug assert.

src/d/actor/d_a_obj_mshokki.cpp
```diff
@@ -4,62 +4,386 @@
...
+        /* SrcGObjCo SPrm    */ 0,
+    },
+    // cM3dGCylS
+    {
+        /* Center */ 0.0f,
+        0.0f,
+        0.0f,
+        /* Radius */ 35.0f,
+        /* Height */ 50.0f,
+    },
```
LagoLunatic: You can help your autoformatter here by adding extra braces
```suggestion
    {{
        /* Center */ {0.0f, 0.0f, 0.0f},
        /* Radius */ 35.0f,
        /* Height */ 50.0f,
    }},
```

Kelebek1: I'm dumping these with the script, I changed Sphere to do this but didn't do the others. I'll fix that.

LagoLunatic:
FYI, the bot says that two funcs regressed for demo with that last commit. Is it related to the (s32) casts that were added?

PR #834, opened 2025-08-05
d_a_obj_ohatch

Kelebek1:


src/d/actor/d_a_obj_ohatch.cpp
```diff
@@ -4,72 +4,324 @@
...
+#if VERSION == VERSION_DEMO
+class daObjOhatch_HIO_c : public JORReflexible {
+public:
+    daObjOhatch_HIO_c();
+    virtual ~daObjOhatch_HIO_c() {}
+
+    void genMessage(JORMContext*);
+
+public:
+    /* 0x04 */ s8 m04;
```
LagoLunatic:
```suggestion
    /* 0x04 */ s8 mNo;
```

src/d/actor/d_a_obj_ohatch.cpp
```diff
@@ -4,72 +4,324 @@
...
+    m24 = 131072.0f;
+    m28 = 6.0f;
+    m2C = -64.0f;
+    m30 = 1024.0f;
+    m34 = 8192.0f;
+    m38 = 0;
+}
+#else
+namespace L_HIO {
+static const f32 m04 = -1;
```
LagoLunatic:
```suggestion
```
"No" doesn't need to exist as a constant since it's for tracking the ID of the HIO and is unused in the retail version of this actor.

src/d/actor/d_a_obj_ohatch.cpp
```diff
@@ -4,72 +4,324 @@
...
+static const f32 m14 = -25.0f;
+static const f32 m18 = 190.0f;
+static const f32 m1C = 1.0f;
+static const f32 m20 = 8192.0f;
+static const f32 m24 = 131072.0f;
+static const f32 m28 = 6.0f;
+static const f32 m2C = -64.0f;
+static const f32 m30 = 1024.0f;
+static const f32 m34 = 8192.0f;
+static const f32 m38 = 0;
```
LagoLunatic:
```suggestion
static const u8 m38 = 0;
```
To match the type of the non-const equivalent version

src/d/actor/d_a_obj_ohatch.cpp
```diff
@@ -4,72 +4,324 @@
...
+                    }
+
+                    dComIfGp_getVibration().StartShock(3, 1, cXyz(0.0f, 1.0f, 0.0f));
+                    mActionFunc = &daObjOhatch_c::tremor_act_proc;
+                }
+            }
+        }
+    }
+#if VERSION == VERSION_DEMO
+    else if (l_HIO.m38 == 1)
```
LagoLunatic: Does this still match if you remove the ifdef? I'd assume so since checking if a const 0 equals 1 should optimize out the entire block in retail.
```suggestion
    else if (HIO(m38) == 1)
```

src/d/actor/d_a_obj_ohatch.cpp
```diff
@@ -4,72 +4,324 @@
...
+    } else {
+        m2F2 += (s32)HIO(m34);
+    }
 }
 
 /* 00000A4C-00000A50       .text open_wait_act_proc__13daObjOhatch_cFv */
 void daObjOhatch_c::open_wait_act_proc() {
-    /* Nonmatching */
+#if VERSION == VERSION_DEMO
+    if ((m29C == 0xff || fopAcM_isSwitch(this, m29C) != 1) && HIO(m38) == 0) {
```
LagoLunatic:
```suggestion
    if ((m29C == 0xff || fopAcM_isSwitch(this, m29C) != TRUE) && HIO(m38) == 0) {
```

PR #843, opened 2025-08-09
d_a_sk

### LINKED ISSUE #172: d_a_sk
Turu

Kelebek1:
Closes #172.

### DIFF
--- a/src/d/actor/d_a_sk.cpp
+++ b/src/d/actor/d_a_sk.cpp
@@ -4,34 +4,192 @@
 //
 
 #include "d/actor/d_a_sk.h"
+#include "d/res/res_sk.h"
 #include "m_Do/m_Do_ext.h"
 #include "d/d_procname.h"
 #include "d/d_priority.h"
+#include "d/d_jnt_hit.h"
 #include "d/d_cc_d.h"
+#include "d/d_com_inf_game.h"
+#include "f_op/f_op_actor_mng.h"
 
 /* 00000078-00000240       .text nodeCallBack__FP7J3DNodei */
-static BOOL nodeCallBack(J3DNode*, int) {
-    /* Nonmatching */
+static BOOL nodeCallBack(J3DNode* node, int calcTiming) {
+    if (calcTiming == 0) {
+        J3DJoint* joint = (J3DJoint*)node;
+        s32 uVar4 = joint->getJntNo();
+        J3DModel* model = j3dSys.getModel();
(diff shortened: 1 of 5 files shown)


src/d/actor/d_a_sk.cpp
```diff
@@ -40,17 +198,48 @@ static BOOL daSk_IsDelete(sk_class*) {
...
 }
 
 /* 0000090C-00000A6C       .text useHeapInit__FP10fopAc_ac_c */
-static BOOL useHeapInit(fopAc_ac_c*) {
-    /* Nonmatching */
+static BOOL useHeapInit(fopAc_ac_c* a_this) {
+    sk_class* i_this = (sk_class*)a_this;
+    J3DModelData* pModelData = (J3DModelData*)dComIfG_getObjectRes("Sk", SK_BDL_TURU_00);
+
+    i_this->mpMorf = new mDoExt_McaMorf(pModelData, NULL, NULL, NULL, ~J3DFrameCtrl::EMode_NONE, 1.0f, 0, -1, 1, NULL, 0x80000, 0x11000022);
```
LagoLunatic:
```suggestion
    i_this->mpMorf = new mDoExt_McaMorf(pModelData, NULL, NULL, NULL, -1, 1.0f, 0, -1, 1, NULL, 0x80000, 0x11000022);
```
-1 means something different from 0 so using J3DFrameCtrl::EMode_NONE here would be misleading.

I think it's supposed to mean the loop mode is read from the BCK, but the BCK is NULL here so I'm not really sure what the programmer's intent was. Maybe we should add a new `J3DFrameCtrl::EMode_NULL = -1,` enum or something for cases like this?

src/d/actor/d_a_sk.cpp
```diff
@@ -40,17 +198,48 @@ static BOOL daSk_IsDelete(sk_class*) {
...
+
+#if VERSION > VERSION_DEMO
+    static Vec sph_offset = {0.0f, 0.0f, 0.0f};
+    static __jnt_hit_data_c search_data[] = {
+        {8, 1, 140.0f, &sph_offset},
+        {8, 2, 90.0f, &sph_offset},
+        {8, 3, 50.0f, &sph_offset},
+    };
+
+    i_this->m924 = JntHit_create(i_this->mpMorf->getModel(), search_data, 3);
```
LagoLunatic:
```suggestion
    i_this->m924 = JntHit_create(i_this->mpMorf->getModel(), search_data, ARRAY_SIZE(search_data));
```

PR #856, opened 2025-08-16
d_a_tsubo

Kelebek1:


include/d/actor/d_a_tsubo.h
```diff
@@ -13,28 +13,166 @@
...
+        #if VERSION == VERSION_DEMO
+        bool is_switch() const { 
+            return fopAcM_isSwitch(const_cast<Act_c*>(this), prmZ_get_swSave());
+        }
+        #else
+        BOOL is_switch() const { 
+            int sw = prmZ_get_swSave();
+            return fopAcM_isSwitch(const_cast<Act_c*>(this), sw);
+        }
+        #endif
```
LagoLunatic: These two are equivalent so it's unlikely it changed in between versions. Let's avoid using version ifdefs for fakematches as it just obscures the real match (e.g. we might still need to change the compiler flags for demo to fix certain things, the current ones are just my guess and are very likely incomplete). Someone else can come back later and find the real fix that matches both versions at once, but for now it's fine for it to only match on retail.
```suggestion
        BOOL is_switch() const { 
            int sw = prmZ_get_swSave();
            return fopAcM_isSwitch(const_cast<Act_c*>(this), sw);
        }
```

src/d/actor/d_a_tsubo.cpp
```diff
@@ -4,622 +4,3686 @@
...
+    }
+
+#if VERSION == VERSION_DEMO
+    bool bVar2 = true;
+    if (!player->getGrabUpEnd() && player->getRightFootOnGround() == 0) {
+        bVar2 = false;
+    }
+#else
+    bool bVar2 = player->getGrabUpEnd() || player->getRightFootOnGround();
+#endif
```
LagoLunatic: Same as above, avoid fakematching just for demo
```suggestion
    bool bVar2 = player->getGrabUpEnd() || player->getRightFootOnGround();
```

src/d/actor/d_a_tsubo.cpp
```diff
@@ -4,622 +4,3686 @@
...
+                }
+            }
+        }
+    } else {
+#if VERSION == VERSION_DEMO
+        bool hitWall = mAcch.ChkWallHit();
+        if (hitWall)
+#else
+        if (mAcch.ChkWallHit())
+#endif
```
LagoLunatic:
```suggestion
        if (mAcch.ChkWallHit())
```

src/d/actor/d_a_tsubo.cpp
```diff
@@ -4,622 +4,3686 @@
...
+                    fVar1 = 0.0f;
+                }
+#else
+                s16 iVar5 = cM_atan2s(sp3C.x, sp3C.z);
+                if (cM_scos(pfVar4->shape_angle.y - iVar5) > 0.866f) {
+                    fVar1 = cM_scos(pfVar4->shape_angle.y - iVar5) * 2.0f + 1.0f;
+                } else {
+                    fVar1 = 0.0f;
+                }
+#endif
```
LagoLunatic:
```suggestion
                s16 iVar5 = cM_atan2s(sp3C.x, sp3C.z);
                if (cM_scos(pfVar4->shape_angle.y - iVar5) > 0.866f) {
                    fVar1 = cM_scos(pfVar4->shape_angle.y - iVar5) * 2.0f + 1.0f;
                } else {
                    fVar1 = 0.0f;
                }
```

include/d/actor/d_a_tsubo.h
Kelebek1: The instructions are very different as there's more int -> bool conversions here, so I'm sure the types did change, but ok.

PR #861, opened 2025-08-19
d_a_ykgr

Kelebek1:


src/d/actor/d_a_ykgr.cpp
```diff
@@ -4,22 +4,119 @@
...
 
 /* 00000234-00000408       .text getPosRate__8daYkgr_cFv */
-void daYkgr_c::getPosRate() {
-    /* Nonmatching */
+f32 daYkgr_c::getPosRate() {
+    if (m_path == NULL) {
+        return 0.0f;
+    }
+
+    f32 fVar4 = 3.4028235e+38;
```
LagoLunatic: I believe this should work
```suggestion
    f32 fVar4 = FLOAT_MAX;
```

include/d/actor/d_a_ykgr.h
```diff
@@ -35,21 +39,59 @@ class daYkgr_c : public fopAc_ac_c {
     inline bool _execute();
     void hide() {}
     void setAimRate(float) {}
-    void setAlpha(unsigned char) {}
-    void set_mtx() {}
+    void setAlpha(unsigned char alpha) { m_alpha = alpha; }
+    void set_mtx() {
```
LagoLunatic: I think usually these setmtx inlines would be defined in the cpp file because they're larger than most inlines. You can add the inline keyword to the declaration to make it still inline.
```suggestion
    inline void set_mtx();
```

PR #877, opened 2025-08-30
d_a_hmlif

### LINKED ISSUE #216: d_a_hmlif
Hmlif, Hyuf1, Hyuf2

Kelebek1:
Closes #216.


src/d/actor/d_a_hmlif.cpp
```diff
@@ -5,9 +5,45 @@
...
+const s16 daHmlif_c::m_bmdidx[] = {HMLIF_BDL_HMLIF, HYUF1_BDL_HYUF1, HYUF2_BDL_HYUF2};
+const s16 daHmlif_c::m_dzbidx[] = {HMLIF_DZB_HMLIF, HYUF1_DZB_HYUF1, HYUF2_DZB_HYUF2};
+const s16 daHmlif_c::m_btpidx[] = {-1, HYUF1_BTP_HYUF1, HYUF2_BTP_HYUF2};
+const s16 daHmlif_c::m_brkidx[] = {HMLIF_BRK_HMLIF, HYUF1_BRK_HYUF1, -1};
+
+const f32 daHmlif_c::m_cull_box[][6] = {
+    {-200.0f, -100.0f, -200.0f, 200.0f, 50.0f, 200.0f},
+    {-200.0f, -100.0f, -200.0f, 200.0f, 450.0f, 200.0f},
+    {-200.0f, -100.0f, -200.0f, 200.0f, 450.0f, 200.0f},
+};
```
LagoLunatic, line 45: Does this still match when changed to an array of fopAc_cullSizeBox? Like in f_op_actor_mng.cpp

src/d/actor/d_a_hmlif.cpp
```diff
@@ -38,110 +74,462 @@ static dCcD_SrcSph l_sph_src = {
...
-    /* Nonmatching */
+BOOL daHmlif_c::Execute(Mtx** o_mtx) {
+    move();
+    set_se();
+    anim_play();
+    check_col();
+    *o_mtx = &mBgMtx;
+    set_mtx();
+    return TRUE;
+    ;
```
LagoLunatic:
```suggestion
```

src/d/actor/d_a_hmlif.cpp
```diff
@@ -38,110 +74,462 @@ static dCcD_SrcSph l_sph_src = {
...
+    mDoExt_modelUpdateDL(mpModel);
+    dComIfGd_setList();
+    return TRUE;
 }
 
 /* 00001610-00001630       .text daHmlif_Create__FPv */
-static cPhs_State daHmlif_Create(void* i_this) {
-    return ((daHmlif_c*)i_this)->daHmlifCreate();
+static cPhs_State daHmlif_Create(void* v_this) {
+    return ((daHmlif_c*)v_this)->daHmlifCreate();
```
LagoLunatic, line 516: Does v_this mean something? It doesn't appear in the assertion strings for either TWW or TP.

Kelebek1: It's the void version of this, if we have a_this for the actor version, and i_this for the input one, I just decided to use v_this for the void version.

src/d/actor/d_a_hmlif.cpp
Kelebek1: Nope, that makes initialisation runtime.

src/d/actor/d_a_hmlif.cpp
LagoLunatic: It's still an input parameter so `i_this` seems right here. The devs tend to name the actor input parameter `i_this` regardless of whether its class is `fopAc_ac_c*` or a subclass so they probably do for `void*` as well.

src/d/actor/d_a_hmlif.cpp
LagoLunatic: Oh I see, in that case it should be `Vec[][2]` like `d_a_windmill`.

PR #888, opened 2025-09-06
d_a_bdk - 98%

maimai0254:
(WIP) The weak functions and the regswap are not finished yet.

### DIFF
--- a/src/d/actor/d_a_bdk.cpp
+++ b/src/d/actor/d_a_bdk.cpp
@@ -1,317 +1,3747 @@
 /**
  * d_a_bdk.cpp
- * Boss - Helmaroc King
+ * Boss - Helmaroc King / 戦闘用大怪鳥 (Giant monstrous bird for battle)
  */
 
 #include "d/dolzel_rel.h" // IWYU pragma: keep
+#include "d/res/res_bdk.h"
 #include "d/actor/d_a_bdk.h"
-#include "m_Do/m_Do_ext.h"
-#include "d/d_procname.h"
-#include "d/d_priority.h"
+#include "d/actor/d_a_player.h"
+#include "d/d_bg_s_movebg_actor.h"
+#include "d/d_bg_s_lin_chk.h"
 #include "d/d_cc_d.h"
+#include "d/d_cc_uty.h"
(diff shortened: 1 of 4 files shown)


src/d/actor/d_a_bdk.cpp
```diff
@@ -1,317 +1,3747 @@
...
+    i_this->bva = new mDoExt_bvaAnm();
+    JUT_ASSERT(0x1864, i_this->bva);
+
+    s32 bva = i_this->bva->init(i_this->mp8F0, (J3DAnmVisibilityFull *)dComIfG_getObjectRes("Bdk", BDK_BVA_HIBIWARE1), 1, 0, 1.0f, 0, -1, FALSE, 0);
+    if (bva == 0) {
+        return FALSE;
+    }
+
+    modelData = (J3DModelData*)dComIfG_getObjectRes("Bdk", BDK_BDL_DK_TAIL);
+    JUT_ASSERT(0x1878, modelData != 0);
```
LagoLunatic: `modelData` is a pointer, so this should use NULL.
```suggestion
    JUT_ASSERT(0x1878, modelData != NULL);
```
The compiler will internally convert nested defines into their value, so "modelData != NULL" will appear as "modelData != 0" in the strings and still match.

This also applies to the other "modelData != 0" asserts in this file.

src/d/actor/d_a_bdk.cpp
```diff
@@ -1,317 +1,3747 @@
...
+    }
+    modelData = (J3DModelData*)dComIfG_getObjectRes("Bdk", BDK_BDL_S_TTOGE);
+    JUT_ASSERT(0x18C4, modelData != 0);
+    i_this->mp62D8 = mDoExt_J3DModel__create(modelData, 0x80000, 0x11000022);
+    if (i_this->mp62D8 == NULL) {
+        return FALSE;
+    }
+
+    i_this->pm_bgw = new dBgW();
+    JUT_ASSERT(0x18CE, i_this->pm_bgw != 0);
```
LagoLunatic: And here too:
```suggestion
    JUT_ASSERT(0x18CE, i_this->pm_bgw != NULL);
```

PR #895, opened 2025-09-22
d_a_gnd - 99% matching

123456789758:


src/d/actor/d_a_gnd.cpp
```diff
@@ -1,180 +1,2768 @@
...
+        200.0f,
+        actor->current.pos.y,
+        i_this->mAcch.GetGroundH(),
+        i_this->mAcch.m_gnd,
+        &actor->tevStr,
+        0,
+        1.0f,
+        dDlst_shadowControl_c::getSimpleTex()
+    );
+    //! @bug(?): Same ID as Jalhalla?
```
SuperDude88, line 497: Happened to be looking through figurine stuff again, the bug here *might* actually be on Jalhalla's side. He calls `dSnap_RegistFig` twice (first time with `DSNAP_TYPE_UNKCE`, second with `DSNAP_TYPE_UNKCB`), and it turns out WWHD removed the first call with `UNKCE` (dunno how the double-register behaves in-game/if it has any observable effect).

PR #896, opened 2025-09-25
d_a_npc_md - 99% matching

123456789758:


include/JSystem/JParticle/JPAEmitter.h
```diff
@@ -221,6 +221,7 @@ class JPABaseEmitter {
     void setGlobalRotation(const JGeometry::TVec3<s16>& rot) {
         JPAGetXYZRotateMtx(rot.x, rot.y, rot.z, mGlobalRotation);
     }
+    void setGlobalRotation(s16 x, s16 y, s16 z) { JPAGetXYZRotateMtx(x, y, z, mGlobalRotation); }
```
LagoLunatic: This is a fake inline, the real inline called in this TU is `setGlobalRotation(const JGeometry::TVec3<s16>& rot)`.

Didn't these functions already match 100% when using the correct inline? What's the reasoning for a fake one?

123456789758: This change was made to match the demo version. The fake inline has been removed, and now the functions match without it.

PR #922, opened 2026-01-14
d_a_npc_mk_static matching

Lord-McSweeney:


src/d/d_a_npc_mk_static.cpp
```diff
@@ -8,66 +8,338 @@
...
+    daPy_lk_c* link_actor = daPy_getPlayerLinkActorClass();
+    dNpc_PathRun_c dStack_74;
+    dStack_74.setInfDrct(param_2->mPath);
+
+    if (mPointIndex2 == mPointIndex1) {
+        return m0;
+    } else {
+        cXyz local_60 = dStack_74.getPoint(mPointIndex2);
+        cXyz local_6c = dStack_74.getPoint(mPointIndex1);
+        s16 sVar4 = cLib_targetAngleY(&param_1->current.pos, &link_actor->current.pos);
```
Lord-McSweeney: I think this is missing an inline, but I'm not sure which one it is

src/d/d_a_npc_mk_static.cpp
```diff
@@ -8,66 +8,338 @@
...
+    s16 local_48;
+    dNpc_calc_DisXZ_AngY(param_1->current.pos, local_2c, NULL, &local_48);
+
+    if (param_3 & 2) {
+        cLib_addCalcAngleS(&param_1->current.angle.y, local_48, 2, 0x2000, 0x400);
+    } else {
+        cLib_addCalcAngleS(&param_1->current.angle.y, local_48, 8, 0x800, 0x200);
+    }
+
+    f32 cosine = JMASCos(param_1->current.angle.y - local_48);
```
Lord-McSweeney: I'm not sure whether this and the other `JMASCos` call should be changed to `cM_scos`

src/d/d_a_npc_mk_static.cpp
```diff
@@ -8,66 +8,338 @@
...
+    if (param_2->mPath != NULL) {
+        param_1->current.pos = param_2->getPoint((int) cM_rndF((int) param_2->maxPoint()));
+        local_90.mFlag &= 0xfffffffd;
+
+        f32 z = fopAcM_GetPosition_p(param_1)->z;
+        f32 y = fopAcM_GetPosition_p(param_1)->y + 200.0f;
+        f32 x = fopAcM_GetPosition_p(param_1)->x;
+        cXyz pos(x, y, z);
+
+        local_90.m_pos = pos;
```
LagoLunatic: Is there a reason this is so verbose? Does something like this not match?
```suggestion
        local_90.SetPos(&param_1->current.pos);
```

src/d/d_a_npc_mk_static.cpp
```diff
@@ -8,66 +8,338 @@
...
 }
 
 /* 800CC400-800CC664       .text setRndPathPos__17daNpc_Mk_Static_cFP10fopAc_ac_cP14dNpc_PathRun_c */
-void daNpc_Mk_Static_c::setRndPathPos(fopAc_ac_c*, dNpc_PathRun_c*) {
-    /* Nonmatching */
+void daNpc_Mk_Static_c::setRndPathPos(fopAc_ac_c* param_1, dNpc_PathRun_c* param_2) {
+    dBgS_GndChk local_90;
+    if (param_2->mPath != NULL) {
+        param_1->current.pos = param_2->getPoint((int) cM_rndF((int) param_2->maxPoint()));
+        local_90.mFlag &= 0xfffffffd;
```
LagoLunatic: OffWall

src/d/d_a_npc_mk_static.cpp
LagoLunatic: Probably cM_scos, JMASCos is usually only called directly within JSystem code, not game code.

src/d/d_a_npc_mk_static.cpp
```diff
@@ -8,66 +8,338 @@
...
-void daNpc_Mk_Static_c::goFarLink_3(fopAc_ac_c*, dNpc_PathRun_c*) {
-    /* Nonmatching */
+u8 daNpc_Mk_Static_c::goFarLink_3(fopAc_ac_c* param_1, dNpc_PathRun_c* param_2) {
+    daPy_lk_c* link_actor = daPy_getPlayerLinkActorClass();
+    dNpc_PathRun_c dStack_74;
+    dStack_74.setInfDrct(param_2->mPath);
+
+    if (mPointIndex2 == mPointIndex1) {
+        return m0;
+    } else {
```
LagoLunatic: No need for else after return, you can reduce the indentation depth later in this function by getting rid of that here

src/d/d_a_npc_mk_static.cpp
LagoLunatic: I don't think there are any inlines that call cLib_targetAngleY.

src/d/d_a_npc_mk_static.cpp
```diff
@@ -8,66 +8,338 @@
...
+        cXyz local_60 = dStack_74.getPoint(mPointIndex2);
+        cXyz local_6c = dStack_74.getPoint(mPointIndex1);
+        s16 sVar4 = cLib_targetAngleY(&param_1->current.pos, &link_actor->current.pos);
+        s16 sVar5 = cLib_targetAngleY(&local_60, &local_6c);
+
+        cM3dGLin unusedLine;
+
+        if (m0 == 1 || m0 == 2) {
+            if (abs(sVar5 - sVar4) < 0x1800) {
```
LagoLunatic: Can further reduce the indentation depth by combining these two if statements
```suggestion
        if ((m0 == 1 || m0 == 2) && abs(sVar5 - sVar4) < 0x1800) {
```

src/d/d_a_npc_mk_static.cpp
```diff
@@ -8,66 +8,338 @@
...
+            if (m8 + 3 < ma) {
+                m8 += 3;
+            }
+            break;
+
+        case 1:
+        case 2:
+            f32 fVar9 = (link_actor->current.pos - param_1->current.pos).absXZ();
+
+            if (walkPath(param_1, param_2, (m0 == 1) ? (u8)2 : (u8)3)) {
```
LagoLunatic: Does this match? The casts there might indicate it's a separate variable instead
```suggestion
            u8 temp = (m0 == 1) ? 2 : 3;
            if (walkPath(param_1, param_2, temp)) {
```

src/d/d_a_npc_mk_static.cpp
LagoLunatic: Maybe like this one
https://github.com/zeldaret/tww/blob/9dedd4456723081946aac0e0afc651cc46c647f6/src/d/actor/d_a_ep.cpp#L479-L482

src/d/d_a_npc_mk_static.cpp
Lord-McSweeney: That doesn't seem to match unless the casts are also kept, like
```
            u8 options = (m0 == 1) ? (u8)2 : (u8)3;
            if (walkPath(param_1, param_2, options)) {
```
Should I do it that way?

src/d/d_a_npc_mk_static.cpp
Lord-McSweeney: I've tried
```
        cXyz pos = param_1->current.pos;
        pos.y += 200.0f;

        local_90.m_pos = pos;
```
```
        cXyz pos = param_1->current.pos;
        pos.y += 200.0f;

        local_90.SetPos(&pos);
```
```
        cXyz* pos = fopAcM_GetPosition_p(param_1);
        pos->y += 200.0f;

        local_90.SetPos(pos);
```
and
```
        cXyz pos = *fopAcM_GetPosition_p(param_1);
        pos.y += 200.0f;

        local_90.SetPos(&pos);
```
but none of them match.

LagoLunatic: What about like this one?
https://github.com/zeldaret/tww/blob/9dedd4456723081946aac0e0afc651cc46c647f6/src/d/actor/d_a_coming3.cpp#L321-L326

(FYI `fopAcM_GetPosition_p` is never used except to pass a pointer to `mAcch.Set`, that's why I don't believe the `fopAcM_GetPosition_p(param_1)->x` stuff is right)

src/d/d_a_npc_mk_static.cpp
LagoLunatic: Ah, in that case nevermind.

PR #933, opened 2026-02-23
rename fopAcM_SetupActor to fopAcM_ct

Luzagar:
based on assert from tp  ` fopAcM_assert(1113, actor, fopAcM_CheckCondition(actor, fopAcCnd_INIT_e), "fopAcM_ct No Call !!");`


LagoLunatic:
I made a separate PR (#978) for this change to do it in a way that avoids conflicts, so I'll close this PR now.

PR #946, opened 2026-04-09
D boss magma 100% matching

noskap:


include/d/d_magma.h
```diff
@@ -11,19 +11,24 @@ class dMagma_ball_c {
     virtual ~dMagma_ball_c() {}
     virtual void calc(f32, u8, int) = 0;
     virtual void update() = 0;
-    virtual void setup(f32, u8, int) = 0;
 
+    virtual void setup(f32, u8, int);
```
LagoLunatic: This is wrong and causes a regression in the vtable.

include/d/d_magma.h
```diff
@@ -11,19 +11,24 @@ class dMagma_ball_c {
...
     /* 0x14 */ f32 mBaseY;
     /* 0x18 */ s16 mWave;
+    /* 0x1A */ u8 mTimer;
     /* 0x1C */ Mtx mPosMtx;
     /* 0x4C */ Mtx mTexProjMtx;
-    /* 0x7C */ u8 mWaveTimer;
+    union {
+        s16 mWaveTimer;
+        u8 mPathTimer;
+    };
```
LagoLunatic: Union seems unlikely. It's more likely that the base class ends with `mTexProjMtx` and then the two child classes just have two completely separate member fields of their own, rather than it being in the parent they inherit from.

src/d/d_boss_magma.cpp
```diff
@@ -6,25 +6,87 @@
...
+#include "d/d_procname.h"
+
+static btd_class* btd = NULL;
 
 /* 80076B00-80076CDC       .text calc__17dMagma_ballBoss_cFfUci */
-void dMagma_ballBoss_c::calc(f32, u8, int) {
-    /* Nonmatching */
+void dMagma_ballBoss_c::calc(f32 param_1, u8 param_2, int param_3) {
+    if (btd != NULL && btd->m6E84 <= -150.0f) {
+        this->mPos.y = -200.0f;
```
LagoLunatic: Typing `this->` is unnecessary. Applies everywhere in this file.

src/d/d_boss_magma.cpp
```diff
@@ -6,25 +6,87 @@
...
+        return;
+    }
+
+    if (btd != NULL) {
+        f32 diff = std::fabsf(btd->m6E7C - std::sqrtf((this->mPos.x * this->mPos.x) + (this->mPos.z * this->mPos.z)));
```
LagoLunatic: Looks like absxz inline, check if it matches

src/d/d_boss_magma.cpp
```diff
@@ -6,25 +6,87 @@
...
 }
 
 /* 80076CDC-80076D50       .text update__17dMagma_ballBoss_cFv */
 void dMagma_ballBoss_c::update() {
-    /* Nonmatching */
+    mDoMtx_stack_c::transS(this->mPos.x, this->mPos.y, this->mPos.z);
+    mDoMtx_stack_c::scaleM(this->mScale, 1.0f, this->mScale);
+
+    PSMTXCopy(mDoMtx_stack_c::get(), this->mTexProjMtx);
+    PSMTXConcat(j3dSys.getViewMtx(), this->mTexProjMtx, this->mPosMtx);
```
LagoLunatic: `PS` functions generally shouldn't be called directly. Try the `mDoMtx_copy`/`mDoMtx_concat` inlines

src/d/d_boss_magma.cpp
```diff
@@ -6,25 +6,87 @@
...
 }
 
 /* 80076D9C-80076FEC       .text setup__17dMagma_ballBoss_cFfUci */
-void dMagma_ballBoss_c::setup(f32, u8, int) {
-    /* Nonmatching */
+void dMagma_ballBoss_c::setup(f32 param_1, u8 param_2, int param_3) {
+    this->mPos.x = cM_rndFX(1300.0f);
+    this->mPos.y = 0.0f;
+
+    f32 max_z = std::sqrtf((1300.0f * 1300.0f) - (this->mPos.x * this->mPos.x));
```
LagoLunatic:
```suggestion
    f32 max_z = std::sqrtf(SQUARE(1300.0f) - SQUARE(this->mPos.x));
```

src/d/d_boss_magma.cpp
```diff
@@ -6,25 +6,87 @@
...
+
+    this->mScale = REG0_F(11) + 0.5f + cM_rndF(REG0_F(12) + 0.5f);
+
+    this->mBaseY = REG0_F(6) + -150.0f - 20.0f + param_1 + 90.0f;
+
+    btd = (btd_class*)fpcEx_Search(b_a_sub, NULL);
+
+    if (btd != NULL) {
+        if (btd->m6E88 == 1) {
+            f32 dist = std::sqrtf((this->mPos.x * this->mPos.x) + (this->mPos.z * this->mPos.z));
```
LagoLunatic: absxz

src/d/d_magma.cpp
```diff
@@ -101,11 +101,11 @@ BOOL dMagma_ball_c::rangeCheck(cXyz& pos, f32* dst) {
...
     if (mWave < 0) {
         setup(offsY, pathNo, roomNo);
-        mWaveTimer = 0;
+        mPathTimer = 0;
         mWave = 0;
     }
 
-    if (cLib_calcTimer(&mWaveTimer) == 0) {
+    if (cLib_calcTimer(&mPathTimer) == 0) {
         mWave += 200;
```
LagoLunatic: Are you sure the name change to "mPathTimer" is correct? It seems to be modifying a variable called "mWave"

src/d/d_boss_magma.cpp
noskap: I gave this a try ` f32 diff = std::fabsf(btd->m6E7C - mPos.absXZ());` but unfortunately no match

image

src/d/d_boss_magma.cpp
```diff
@@ -5,26 +5,87 @@
...
-void dMagma_ballBoss_c::calc(f32, u8, int) {
-    /* Nonmatching */
+void dMagma_ballBoss_c::calc(f32 param_1, u8 param_2, int param_3) {
+    if (btd != NULL && btd->m6E84 <= -150.0f) {
+        mPos.y = -200.0f;
+        return;
+    }
+
+    if (btd != NULL) {
+        f32 diff = std::fabsf(btd->m6E7C - std::sqrtf((mPos.x * mPos.x) + (mPos.z * mPos.z)));
```
LagoLunatic:
```suggestion
        f32 diff = std::fabsf(btd->m6E7C - std::sqrtf(SQUARE(mPos.x) + SQUARE(mPos.z)));
```

src/d/d_boss_magma.cpp
```diff
@@ -5,26 +5,87 @@
...
+
+    mScale = REG0_F(11) + 0.5f + cM_rndF(REG0_F(12) + 0.5f);
+
+    mBaseY = REG0_F(6) + -150.0f - 20.0f + param_1 + 90.0f;
+
+    btd = (btd_class*)fpcEx_Search(b_a_sub, NULL);
+
+    if (btd != NULL) {
+        if (btd->m6E88 == 1) {
+            f32 dist = std::sqrtf((mPos.x * mPos.x) + (mPos.z * mPos.z));
```
LagoLunatic:
```suggestion
            f32 dist = std::sqrtf(SQUARE(mPos.x) + SQUARE(mPos.z));
```

src/d/d_boss_magma.cpp
```diff
@@ -5,26 +5,87 @@
...
+    if (btd != NULL) {
+        f32 diff = std::fabsf(btd->m6E7C - std::sqrtf((mPos.x * mPos.x) + (mPos.z * mPos.z)));
+
+        if (diff < 300.0f) {
+            f32 targetY = (mBaseY + (50.0f + REG0_F(5)) + 30.0f) - (diff * 0.1f);
+            cLib_addCalc2(&mPos.y, targetY, 0.5f, 20.0f);
+            cLib_addCalc2(&mScale, 1.5f, 0.2f, 0.2f);
+
+            mWave = 0x4000;
+            mWaveTimer = 1000;
```
LagoLunatic: It looks like you copied the name `mWaveTimer` from the other class, but it doesn't look like this field is being used as a timer in this case, right? So leaving it unnamed would be less misleading.

src/d/d_magma.cpp
LagoLunatic: Was there a reason for the rename from `mWaveTimer` -> `mPathTimer` in this class?

src/d/d_boss_magma.cpp
noskap: I got stuck on this one. I couldn't seem to change it without causing regression in d_magma. 

-dMagma_ballPath_c wants a u8
-but dMagma_ballBoss_c wants a s16

for now I've added those fields in the dMagma_ballPath_c and dMagma_ballBoss_c classes respectively, and that cleared it up

PR #948, opened 2026-04-12
d_a_nzg ok

ThePlayerRolo:
Rat group progress


include/d/actor/d_a_nzg.h
```diff
@@ -2,10 +2,30 @@
...
 #include "f_op/f_op_actor.h"
+#include "d/d_cc_d.h"
 
-class nzg_class : public fopAc_ac_c {
+struct dPath;
+
+class nzg_class : public fopEn_enemy_c {
 public:
-    /* Place member variables here */
+    /* 0x2AC */ request_of_phase_process_class mPhs;
```
LagoLunatic: C-style classes (the ones with names ending in `_class` instead of `_c`) should use composition instead of inheritance:
```suggestion
struct nzg_class {
    /* 0x000 */ fopEn_enemy_c actor;
    /* 0x2AC */ request_of_phase_process_class mPhs;
```

src/d/actor/d_a_nzg.cpp
```diff
@@ -1,32 +1,113 @@
...
+    return TRUE;
 }
 
 /* 000000E4-00000348       .text nzg_00_move__FP9nzg_class */
-void nzg_00_move(nzg_class*) {
-    /* Nonmatching */
+void nzg_00_move(nzg_class* i_this) {
+    /* Nonmatching for Demo */
+    //fixes regswap issue in demo for the most part but introduces a misplaced instruction error
+    nzg_class* actor = i_this;
```
LagoLunatic: Adding on to the above, every function in a C-style class should start with a line like this:
```suggestion
    fopAc_ac_c* actor = &i_this->actor;
```
When the base class's members are accessed you should use `actor`, and when the specific actor's members are accessed you use `i_this`. Doing this the same way the original devs did can sometimes fix regalloc and misplaced instruction errors, so if you adjust this it might fix the nonmatching for demo you mention in the comment.

src/d/actor/d_a_nzg.cpp
```diff
@@ -35,18 +116,39 @@ static BOOL daNZG_IsDelete(nzg_class*) {
...
 
 /* 000004D8-00000508       .text daNZG_Delete__FP9nzg_class */
-static BOOL daNZG_Delete(nzg_class*) {
-    /* Nonmatching */
+static BOOL daNZG_Delete(nzg_class* i_this) {
+#if VERSION == VERSION_DEMO
+    g_dComIfG_gameInfo.mResControl.deleteObjectRes("NZG");
+#else
+    dComIfG_resDelete(&i_this->mPhs, "NZG");
+#endif
```
LagoLunatic: There's a macro to simplify this
```suggestion
    dComIfG_resDeleteDemo(&i_this->mPhs, "NZG");
```

ThePlayerRolo:
updated to match review

now everything matches besides rodata which just seems to be due to data used in unused functions presisting

LagoLunatic:
> now everything matches besides rodata which just seems to be due to data used in unused functions presisting

You can try something like
```cpp
static f32 dummy() {
    return 300.0f;
}
```
To simulate the unused functions.

ThePlayerRolo:
everything is linked now

PR #862, opened 2025-08-20
d_a_obj_gnndemotakis

Xxj2211:
Closes #400

PR #953, opened 2026-04-15
d_a_obj_gnndemotakis

jaysc:
Issue: #400
 
I started this without realising there was an old PR here: #862 
Huge thanks to @Xxj2211 as It helped a lot.

Hopefully next time I can do one entirely myself.

---
Todo:
## create_heap
Some reason it's not matching fully. Trying to figure out why.
image


LagoLunatic:
> Some reason it's not matching fully. Trying to figure out why.

That looks like an objdiff bug, it seems to already match. Which version of objdiff are you using? If it's an old one update to the latest version.

Also, try linking it to see if it actually matches or not, regardless of what objdiff says.

jaysc:
> > Some reason it's not matching fully. Trying to figure out why.
> 
> That looks like an objdiff bug, it seems to already match. Which version of objdiff are you using? If it's an old one update to the latest version.
> 
> Also, try linking it to see if it actually matches or not, regardless of what objdiff says.

I'm using 3.7.1.

The TU is failing unfortunately. Not entirely sure where to go next with this.

src/d/actor/d_a_obj_gnndemotakis.cpp
```diff
@@ -4,43 +4,99 @@
...
 void daObjGnntakis_c::init_mtx() {
-    /* Nonmatching */
+    mpModel->setBaseScale(scale);
 }
 
 /* 000003FC-00000514       .text _execute__15daObjGnntakis_cFv */
 bool daObjGnntakis_c::_execute() {
-    /* Nonmatching */
+    if (strcmp(dComIfGp_getStartStageName(), "GTower") == 0 &&
+        g_dComIfG_gameInfo.play.mEvtCtrl.runCheck() &&
```
jaysc: I'm aware of `dComIfGp_event_runCheck`, but that seems to reduce the match rate. Do we still want to use `dComIfGp_event_runCheck`?

LagoLunatic: Yes, because the debug map says that `dComIfGp_event_runCheck` was used.

If you're having trouble matching a function, you can follow the instructions in the tutorial to post a scratch and ask for help, and the fix may be something simple. For example, maybe you have to do `dComIfGp_event_runCheck() != FALSE` rather than just `dComIfGp_event_runCheck()`.

jaysc: Oh wow I'm surprised that was the difference. 
Yeah I forgot it makes sense to follow the map. I'll be more careful going forward.

jaysc:
This is matching, and TU shows matching too other than `D44J01`.

PR #956, opened 2026-04-17
fix: d_a_movie_player

swgillespie:
This file hasn't been touched much so I took a shot at some of the more obvious functions.

I can't get the string literal `OSReport("Error happen");` - this is the right string in the target, but because we're missing a bunch of string literals in this file the linker isn't putting it in the right place in the decomp.

### DIFF
--- a/src/d/actor/d_a_movie_player.cpp
+++ b/src/d/actor/d_a_movie_player.cpp
@@ -20,6 +20,8 @@
 #include "m_Do/m_Do_graphic.h"
 #include "dolphin/os/OSMessage.h"
 #include "dolphin/base/PPCArch.h"
+#include "dolphin/vi/vi.h"
+#include "JSystem/JAudio/JASAiCtrl.h"
 
 static u8 THPStatistics[0x460]; // TODO
 
@@ -108,7 +110,7 @@ THPVideoInfo daMP_videoInfo;
 THPAudioInfo daMP_audioInfo;
 u32 daMP_DrawPosX;
 u32 daMP_DrawPosY;
-void** daMP_buffer;
+void* daMP_buffer;
 
 BOOL daMP_Fail_alloc;
 u16 daMP_backup_FrameRate;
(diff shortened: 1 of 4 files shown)


src/d/d_map.cpp
```diff
@@ -1675,7 +1675,10 @@ void dMap_c::getFmapChkPntPrm(int, s8*, s8*, s16*, s16*, u8*) {
 
 /* 8004ACD8-8004AD00       .text setFmapChkDtPrm__6dMap_cFv */
 void dMap_c::setFmapChkDtPrm() {
-    /* Nonmatching */
+    int* p = (int*)g_dComIfG_gameInfo.play.mpFmapData;
```
LagoLunatic, line 1678: Missing inline

src/d/d_map.cpp
```diff
@@ -1675,7 +1675,10 @@ void dMap_c::getFmapChkPntPrm(int, s8*, s8*, s16*, s16*, u8*) {
...
+    if (p == NULL) return;
+    mFmapChkPntValue = *p;
+    mFmapChkPntData_p = (int)(p + 1);
```
LagoLunatic, line 1681: This shouldn't be a simple int pointer, it's going to be a pointer to a struct of some kind. More of the functions in this TU need to be decompiled to figure out what that struct is and what goes inside of it, so decompiling just this one function on its own isn't a good idea. It's better to do a whole TU or at least a big chunk of it at once.

src/d/actor/d_a_movie_player.cpp
```diff
@@ -2955,23 +3005,34 @@ void daMP_THPPlayerDrawCurrentFrame(const GXRenderModeObj*, u32, u32, u32, u32)
...
+    }
+    return FALSE;
 }
 
 /* 00005DF4-00005E3C       .text daMP_THPPlayerGetAudioInfo__FP12THPAudioInfo */
-void daMP_THPPlayerGetAudioInfo(THPAudioInfo*) {
-    /* Nonmatching */
+BOOL daMP_THPPlayerGetAudioInfo(THPAudioInfo* info) {
+    if (daMP_ActivePlayer.mA0 != 0) {
+        memcpy(info, &daMP_ActivePlayer.mAudioInfo, 0x10);
```
LagoLunatic, line 3019: Missing sizeof

src/d/actor/d_a_movie_player.cpp
```diff
@@ -2955,23 +3005,34 @@ void daMP_THPPlayerDrawCurrentFrame(const GXRenderModeObj*, u32, u32, u32, u32)
 }
 
 /* 00005DAC-00005DF4       .text daMP_THPPlayerGetVideoInfo__FP12THPVideoInfo */
-void daMP_THPPlayerGetVideoInfo(THPVideoInfo*) {
-    /* Nonmatching */
+BOOL daMP_THPPlayerGetVideoInfo(THPVideoInfo* info) {
+    if (daMP_ActivePlayer.mA0 != 0) {
+        memcpy(info, &daMP_ActivePlayer.mVideoInfo, 0xC);
```
LagoLunatic, line 3010: sizeof

src/d/actor/d_a_movie_player.cpp
```diff
@@ -2847,22 +2860,28 @@ void daMP_audioCallbackWithMSound(s32) {
 
 /* 00004A04-00004A30       .text daMP_audioInitWithMSound__Fv */
 void daMP_audioInitWithMSound() {
-    /* Nonmatching */
+    JASystem::Kernel::registerMixCallback((s16*(*)(s32))daMP_audioCallbackWithMSound, 3);
```
LagoLunatic, line 2863: Cast should not be necessary here

src/d/actor/d_a_movie_player.cpp
```diff
@@ -2986,17 +3047,32 @@ BOOL daMP_ActivePlayer_Init(const char*) {
...
+        OSReport("Error happen");
+    }
 }
 
 /* 000061DC-00006230       .text daMP_ActivePlayer_Draw__Fv */
 void daMP_ActivePlayer_Draw() {
-    /* Nonmatching */
+    const GXRenderModeObj* renderMode = JUTVideo::getManager()->getRenderMode();
+    daMP_THPPlayerDrawCurrentFrame(renderMode, daMP_DrawPosX, daMP_DrawPosY,
+        ((u32*)&daMP_videoInfo)[0], ((u32*)&daMP_videoInfo)[1]);
```
LagoLunatic, line 3074: This is another struct, so similar to above you should not be accessing its fields via casts and indexes. Struct fields should be defined at the same time as the functions that access their members are decompiled.

LagoLunatic (commented):
Are you new to decompiling? If so, I don't recommend you work on d_a_movie_player as it's probably going to be one of the most complex actors in the game. It's better to pick some small actor to get to 100% to learn the basics before trying something complicated like this. There's also a [decompiling tutorial here](https://github.com/zeldaret/tww/blob/main/docs/decompiling.md) that goes over all the steps for decompiling a single actor.

PR #959, opened 2026-04-19
d_a_tag_ba1

swgillespie:
This is my attempt at picking up https://github.com/zeldaret/tww/issues/182 which was claimed almost a year ago but hasn't had any activity. This is my first substantial decomp. I welcome feedback on it!

`d_a_tag_ba1` is the tag actor for healing Grandma with a fairy. This TU matches but does have some weak symbol issues stemming from a `std::sqrtf` brought in via HIO. From here I think we can intuit that event bit `0x2A20` is probably the "Healed Grandma" tag, but I didn't rename it in this PR since I haven't conclusively proven that - happy to do that though.

`daTag_Ba1_c::_execute` doesn't use the inline `dComIfGs_*` helpers for the most part because the codegen requires that `play` be a local, but I tried to use the existing inline functions wherever I could.

### DIFF
--- a/include/d/actor/d_a_tag_ba1.h
+++ b/include/d/actor/d_a_tag_ba1.h
@@ -5,24 +5,17 @@
 
 class daTag_Ba1_c : public fopAc_ac_c {
 public:
-    void XyCheck_cB(int);
-    void XyEvent_cB(int);
-    void createInit();
+    s16 XyCheck_cB(int);
+    s16 XyEvent_cB(int);
+    bool createInit();
     BOOL _draw();
     BOOL _execute();
     BOOL _delete();
     cPhs_State _create();
 
 public:
-    /* Place member variables here */
-};
-
-class daTag_Ba1_HIO_c {
-public:
-    daTag_Ba1_HIO_c();
-
-public:
-    /* Place member variables here */
+    /* 0x290 */ s16 eventIds[1];
+    /* 0x292 */ s16 eventIdx;
 };
 
 #endif /* D_A_TAG_BA1_H */
--- a/include/d/actor/d_a_tag_ba1.h
+++ b/include/d/actor/d_a_tag_ba1.h
@@ -14,8 +14,26 @@ public:
     cPhs_State _create();
 
 public:
-    /* 0x290 */ s16 eventIds[1];
-    /* 0x292 */ s16 eventIdx;
+    /* 0x290 */ s16 mEventIds[1];
+    /* 0x292 */ s16 mEventIdx;
 };
+
+class daTag_Ba1_HIO_c : public JORReflexible {
+public:
+    struct hio_prm_c {
+        /* 0x0 */ u8 field_0x0;
+    };
+
+    daTag_Ba1_HIO_c();
+    virtual ~daTag_Ba1_HIO_c() {}
+
+    void genMessage(JORMContext* ctx) {}
+
+public:
+    /* 0x04 */ s8 mNo;
+    /* 0x08 */ s32 mRefCount;
+    /* 0x0C */ hio_prm_c mPrm;
+};
+
 
 #endif /* D_A_TAG_BA1_H */
--- a/src/d/actor/d_a_tag_ba1.cpp
+++ b/src/d/actor/d_a_tag_ba1.cpp
@@ -10,23 +10,6 @@
 #include "d/d_priority.h"
 #include "m_Do/m_Do_hostIO.h"
 
-class daTag_Ba1_HIO_c : public JORReflexible {
-public:
-    struct hio_prm_c {
-        /* 0x0 */ u8 field_0x0;
-    };
-
-    daTag_Ba1_HIO_c();
-    virtual ~daTag_Ba1_HIO_c() {}
-
-    void genMessage(JORMContext* ctx) {}
-
-public:
-    /* 0x04 */ s8 mNo;
-    /* 0x08 */ s32 mRefCount;
-    /* 0x0C */ hio_prm_c prm;
-};
-
 static const char* l_evn_tbl[1] = {"Use_Fairy"};
 
 static daTag_Ba1_HIO_c l_HIO;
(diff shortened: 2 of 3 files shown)


src/d/actor/d_a_tag_ba1.cpp
```diff
@@ -5,57 +5,124 @@
 
 #include "d/dolzel_rel.h" // IWYU pragma: keep
 #include "d/actor/d_a_tag_ba1.h"
+#include "d/d_item_data.h"
 #include "d/d_procname.h"
 #include "d/d_priority.h"
+#include "m_Do/m_Do_hostIO.h"
+
+class daTag_Ba1_HIO_c : public JORReflexible {
```
LagoLunatic: You said there are weak ordering issues from the HIO. Have you already tried keeping the HIO class in the header to see if that fixes them and allows the TU to be linked? Or did you move it to the cpp because that improves the order?

src/d/actor/d_a_tag_ba1.cpp
```diff
@@ -5,57 +5,124 @@
...
-    /* Nonmatching */
+static s16 daTag_Ba1_XyCheck_cB(void* obj, int idx) {
+    return static_cast<daTag_Ba1_c*>(obj)->XyCheck_cB(idx);
 }
 
 /* 00000164-00000184       .text XyCheck_cB__11daTag_Ba1_cFi */
-void daTag_Ba1_c::XyCheck_cB(int) {
-    /* Nonmatching */
+s16 daTag_Ba1_c::XyCheck_cB(int index) {
+    return g_dComIfG_gameInfo.play.mSelectItem[index] == dItem_FAIRY_BOTTLE_e;
```
LagoLunatic: Missing inline. Check the debug map for this function and it tells you what inlines should be used in what order in the linker tree since this is a REL.

src/d/actor/d_a_tag_ba1.cpp
```diff
@@ -5,57 +5,124 @@
...
-    /* Nonmatching */
+static s16 daTag_Ba1_XyEvent_cB(void* obj, int idx) {
+    return static_cast<daTag_Ba1_c*>(obj)->XyEvent_cB(idx);
 }
 
 /* 000001A4-000001C0       .text XyEvent_cB__11daTag_Ba1_cFi */
-void daTag_Ba1_c::XyEvent_cB(int) {
-    /* Nonmatching */
+s16 daTag_Ba1_c::XyEvent_cB(int idx) {
+    this->eventIdx = 0;
```
LagoLunatic: Typing `this->` is unnecessary. Applies everywhere.

src/d/actor/d_a_tag_ba1.cpp
```diff
@@ -5,57 +5,124 @@
...
 BOOL daTag_Ba1_c::_draw() {
-    /* Nonmatching */
+    return TRUE;
 }
 
 /* 00000290-00000340       .text _execute__11daTag_Ba1_cFv */
 BOOL daTag_Ba1_c::_execute() {
-    /* Nonmatching */
+    int staffId = -1;
+    dComIfG_play_c* play = &g_dComIfG_gameInfo.play;
```
LagoLunatic: g_dComIfG_gameInfo generally shouldn't be accessed directly. I realize you said this function doesn't match if you use the correct inlines, but the debug maps show `dComIfGp_event_runCheck` was used here, and the issues that used to exist with gameinfo that sometimes required these fakematches were resolved last year, so we should avoid going back to this hack again.

The fix will likely be some small variation on how you have to type the code to get it to match exactly. If you're having trouble matching the function with the proper inlines, follow the tutorial's instructions on how to ask for help by creating a scratch and someone can probably help you match it the correct way.

include/d/actor/d_a_tag_ba1.h
```diff
@@ -5,24 +5,17 @@
...
-};
-
-class daTag_Ba1_HIO_c {
-public:
-    daTag_Ba1_HIO_c();
-
-public:
-    /* Place member variables here */
+    /* 0x290 */ s16 eventIds[1];
+    /* 0x292 */ s16 eventIdx;
```
LagoLunatic: For C++-style classes (name ends in `_c`) the original devs prefix member names with `m`
```suggestion
    /* 0x290 */ s16 mEventIds[1];
    /* 0x292 */ s16 mEventIdx;
```
https://github.com/zeldaret/tww/blob/main/docs/coding_guidelines.md#naming-style

If this was a C-style class instead (name ends in `_class`) then no prefix like how you did it would match the original devs style.

LagoLunatic (commented):
Missed one thing

src/d/actor/d_a_tag_ba1.cpp
swgillespie: I moved it to the cpp in the hopes that it would improve the order, but it didn't - moving it to the header doesn't help either.

I'm pretty sure that the two symbols (`std::sqrtf(float)::_three` and `std::sqrtf(float)::_half`) are coming in transitively from `cXyz::abs`, which isn't used here but potentially reachable from from this TU; I'll keep digging.

LagoLunatic: `_three` and `_half` are in every TU. They don't affect linking, you can safely ignore them completely. They're also variables, not functions like the comment you added to configure.py suggests.

Did you try actually linking the TU and see that it failed? If you didn't, then it might succeed.

swgillespie:
@LagoLunatic thanks for the review - I've pushed everything except the `std::sqrtf` thing, which I'll try actually linking.

src/d/actor/d_a_tag_ba1.cpp
swgillespie: will try!

swgillespie: I think the real problem is that my `a_prm_tbl` is wrong. I'll figure it out and come back to this PR

999sian:
I independently finished the remaining D44J01/demo matching work and pushed the whole-object patch here: https://github.com/999sian/tww/commit/a76889519ae3545281a762c08bce7e39f4c76259

The key demo divergence is `createInit`: D44J01 calls `dComIfGp_getPEvtManager()->getEventIdx(...)`, while retail uses `dComIfGp_evmng_getEventIdx(...)`. The patch also keeps the reviewed event-manager inlines and correct member naming.

Fresh local verification on the pushed commit:

- D44J01: 17/17 functions, 1148/1148 code, 166/166 data; REL `28beebf8d9a715b9bfe7fefd0c3a68b340086283`
- GZLP01: 17/17 functions, 1140/1140 code, 166/166 data; REL `177eadcaa4723f8534e4a7c98269edeb417c62b4`

Both hashes equal `config/<version>/build.sha1`. Please feel free to cherry-pick or reuse the demo fix.

PR #960, opened 2026-04-20
JAIBasic and JAISound mostly matching

PJB3005:
Only non-matching functions are some regalloc/ordering issues I couldn't quite figure out.

### DIFF
--- a/src/JSystem/JAudio/JAIBasic.cpp
+++ b/src/JSystem/JAudio/JAIBasic.cpp
@@ -215,7 +215,7 @@ void JAIBasic::processFrameWork() {
 
 /* 802904B4-802904EC       .text startSoundVec__8JAIBasicFUlPP8JAISoundP3VecUlUlUc */
 void JAIBasic::startSoundVec(u32 param_1, JAISound** param_2, Vec* param_3, u32 param_4, u32 param_5, u8 param_6) {
-    JAInter::Actor actor(param_3, param_3, (u32)param_3, (void*) param_5);
+    JAInter::Actor actor(param_3, param_3, TRUE, (void*) param_5);
 
     startSoundActor(param_1, param_2, &actor, param_4, param_6);
 }
--- a/src/JSystem/JAudio/JAIBasic.cpp
+++ b/src/JSystem/JAudio/JAIBasic.cpp
@@ -215,7 +215,7 @@ void JAIBasic::processFrameWork() {
 
 /* 802904B4-802904EC       .text startSoundVec__8JAIBasicFUlPP8JAISoundP3VecUlUlUc */
 void JAIBasic::startSoundVec(u32 param_1, JAISound** param_2, Vec* param_3, u32 param_4, u32 param_5, u8 param_6) {
-    JAInter::Actor actor(param_3, param_3, (u32)param_3, (void*) param_5);
+    JAInter::Actor actor(param_3, param_3, param_5, param_3);
 
     startSoundActor(param_1, param_2, &actor, param_4, param_6);
 }
--- a/src/JSystem/JAudio/JAIBasic.cpp
+++ b/src/JSystem/JAudio/JAIBasic.cpp
@@ -215,7 +215,7 @@ void JAIBasic::processFrameWork() {
 
 /* 802904B4-802904EC       .text startSoundVec__8JAIBasicFUlPP8JAISoundP3VecUlUlUc */
 void JAIBasic::startSoundVec(u32 param_1, JAISound** param_2, Vec* param_3, u32 param_4, u32 param_5, u8 param_6) {
-    JAInter::Actor actor(param_3, param_3, TRUE, (void*) param_5);
+    JAInter::Actor actor(param_3, param_3, (u32)param_3, (void*) param_5);
 
     startSoundActor(param_1, param_2, &actor, param_4, param_6);
 }
--- a/src/JSystem/JAudio/JAIBasic.cpp
+++ b/src/JSystem/JAudio/JAIBasic.cpp
@@ -215,7 +215,7 @@ void JAIBasic::processFrameWork() {
 
 /* 802904B4-802904EC       .text startSoundVec__8JAIBasicFUlPP8JAISoundP3VecUlUlUc */
 void JAIBasic::startSoundVec(u32 param_1, JAISound** param_2, Vec* param_3, u32 param_4, u32 param_5, u8 param_6) {
-    JAInter::Actor actor(param_3, param_3, 0, (void*) param_5);
+    JAInter::Actor actor(param_3, param_3, TRUE, (void*) param_5);
 
     startSoundActor(param_1, param_2, &actor, param_4, param_6);
 }
--- a/src/JSystem/JAudio/JAIBasic.cpp
+++ b/src/JSystem/JAudio/JAIBasic.cpp
@@ -15,7 +15,9 @@
 #include "JSystem/JAudio/JAISeMgr.h"
 #include "JSystem/JAudio/JAISequenceHeap.h"
 #include "JSystem/JAudio/JAISequenceMgr.h"
+#include "JSystem/JAudio/JAISoundTable.h"
 #include "JSystem/JAudio/JAIStreamMgr.h"
+#include "JSystem/JAudio/JAISystemInterface.h"
 #include "JSystem/JAudio/JASAudioThread.h"
 #include "JSystem/JAudio/JASDriverIF.h"
 #include "JSystem/JAudio/JASSystemHeap.h"
@@ -92,9 +94,6 @@ void JAIBasic::initHeap() {
         msCurrentHeap = field_0x8;
     } else {
         msCurrentHeap = JASDram;
-        if (msCurrentHeap) {
-            return;
-        }
     }
     JUT_ASSERT_MSG(186, msCurrentHeap, "JAIBasic::initHeap オーディオヒープが異常（NULL）です。\n");
 }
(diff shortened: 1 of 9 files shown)


src/JSystem/JAudio/JAISound.cpp
```diff
@@ -251,12 +254,84 @@ f32 JAISound::setDistanceVolumeCommon(f32 param_1, u8 param_2) {
...
-    /* Nonmatching */
+    f32 fVar1;
+    f32 fVar2;
+    f32 fVar3;
+
+    f32 realRes;
+
+    if (JAIGlobalParameter::audioCameraMax == 1) {
+        f32 result;
+        PositionInfo_t* pPVar4 = this->mPositionInfo;
```
LagoLunatic: Typing out `this->` is not necessary, it's implicit in C++ classes.

src/JSystem/JAudio/JAISound.cpp
```diff
@@ -435,106 +509,304 @@ void JAISound::setTrackInterruptSwitch(u8 param_1, u8 param_2) {
...
+            fxMix = (s32)((f32)JAIGlobalParameter::seDistanceFxParameter * (fVar1 / JAIGlobalParameter::distanceMax));
+        } else {
+            fxMix = JAIGlobalParameter::seDistanceFxParameter;
+        }
+    }
+    if ((fxMix & 0xffff) > 0x7f) {
+        fxMix = 0x7f;
+    }
+    setSeInterFxmix(4, (float)(fxMix & 0xff) / 127.0f, param_1, 0);
+    return;
```
LagoLunatic: Explicit returns aren't needed at the end of void functions. Only needed for early returns or for non-void functions.

src/JSystem/JAudio/JAISound.cpp
```diff
@@ -628,29 +920,110 @@ JAInter::StreamParameter* JAISound::getStreamParameter() {
...
+void JAISound::initParameter(JAISound** param_1, JAInter::Actor* param_2, u32 param_3, u32 param_4, u8 param_5, void* param_6) {
+    u8 uVar1;
+
+    this->mSoundID = param_3;
+    if (param_2) {
+        this->field_0x24 = param_2->field_0x0;
+        if (param_2->field_0x0) {
+            this->field_0x28 = param_2->field_0x4;
+            this->field_0x2c = (int)param_2->field_0x8;
+            this->field_0x1c = (int)param_2->field_0xc;
```
LagoLunatic: Are you sure these casts are correct? Looks like the right side is a pointer type and the left is an integer type which is a bit unusual. The field types already in the decomp aren't necessarily correct especially since there's a comment on some existing JAI fields saying they're just guesses. If you can swap things around to avoid weird casts while still matching that's more likely to be correct.

src/JSystem/JAudio/JAIBasic.cpp
```diff
@@ -206,12 +205,19 @@ BOOL JAIBasic::initReadFile() {
...
+    JAInter::SequenceMgr::processGFrameSequence();
+    JAInter::StreamMgr::processGFrameStream();
+    field_0x10 += 1;
 }
 
 /* 802904B4-802904EC       .text startSoundVec__8JAIBasicFUlPP8JAISoundP3VecUlUlUc */
-void JAIBasic::startSoundVec(u32, JAISound**, Vec*, u32, u32, u8) {
-    /* Nonmatching */
+void JAIBasic::startSoundVec(u32 param_1, JAISound** param_2, Vec* param_3, u32 param_4, u32 param_5, u8 param_6) {
+    JAInter::Actor actor(param_3, param_3, (u32)param_3, (void*) param_5);
```
LagoLunatic: Same here, the casts look unusual and I'd guess that `JAInter::Actor` is currently wrong and needs to be fixed so these casts aren't needed.

src/JSystem/JAudio/JAIBasic.cpp
```diff
@@ -220,23 +226,90 @@ void JAIBasic::startSoundActor(u32 param_1, JAISound** param_2, JAInter::Actor*
...
+            param_2 = (JAISound**)&JAInter::SequenceMgr::FixSeqBufPointer[((SoundInfo*)param_6)->_05];
+        }
+        JAInter::SequenceMgr::storeSeqBuffer(param_2, param_3, param_1, param_4, param_5, param_6);
+        return;
+    case JAISoundID_Type_Se:
+        if (JAInter::SeMgr::seEntryCancel[param_1 >> 0xc] == 0) {
+            JAInter::SeMgr::storeSeBuffer(param_2, param_3, param_1, param_4, param_5, param_6);
+            return;
+        }
+        if (param_2 == (JAISound**)0x0) {
```
LagoLunatic: These `(Type*)0x0` things you see all over are just weird ghidra-isms, the actual way to write a null pointer is:
```suggestion
        if (param_2 == NULL) {
```

src/JSystem/JAudio/JAIBasic.cpp
```diff
@@ -291,44 +453,131 @@ BOOL JAIBasic::getMapInfoGround(u32 param_1) {
...
+        uVar2 = param_1->field_0x36c & 0xff;
+        pJVar3 = SeMgr::seTrackUpdate;
+        outer = param_1->mOuterParam;
+        outer->setParam(1, SeMgr::seTrackUpdate[uVar2].field_0x4);
+        outer->setParam(8, pJVar3[uVar2].field_0x10);
+        outer->setParam(2, pJVar3[uVar2].field_0x8);
+        outer->setParam(4, pJVar3[uVar2].field_0xc);
+        f32 fVar1;
+        if (msBasic->field_0xd != 2) {
+            fVar1 = 0.0;
```
LagoLunatic: These singles should be written as
```suggestion
            fVar1 = 0.0f;
```
the form without the suffix is for doubles. Ghidra is easily confused and doesn't know the difference between a single and a double so keep that in mind.

src/JSystem/JAudio/JAIBasic.cpp
```diff
@@ -291,44 +453,131 @@ BOOL JAIBasic::getMapInfoGround(u32 param_1) {
...
+        u32 uVar2;
+        JASystem::TTrack::TOuterParam* outer;
+
+        uVar2 = param_1->field_0x36c & 0xff;
+        pJVar3 = SeMgr::seTrackUpdate;
+        outer = param_1->mOuterParam;
+        outer->setParam(1, SeMgr::seTrackUpdate[uVar2].field_0x4);
+        outer->setParam(8, pJVar3[uVar2].field_0x10);
+        outer->setParam(2, pJVar3[uVar2].field_0x8);
+        outer->setParam(4, pJVar3[uVar2].field_0xc);
```
LagoLunatic: The first args here look like the `JASOuterParamFlag` enum. You can use the reference search on a function in vscode to see if anywhere in the existing decomp calls a function using enums.

src/JSystem/JAudio/JAISound.cpp
```diff
@@ -435,106 +509,304 @@ void JAISound::setTrackInterruptSwitch(u8 param_1, u8 param_2) {
...
+            if (fld < JAIGlobalParameter::seDolbyBehindDistanceMax) {
+                value = JAIGlobalParameter::seDolbyCenterValue +
+                        (127.0f - JAIGlobalParameter::seDolbyCenterValue) * (fld / JAIGlobalParameter::seDolbyBehindDistanceMax);
+            } else {
+                value = 127.0f;
+            }
+        }
+    }
+
+    setSeInterDolby(4, (u32)((u8)value & 0xff) / 127.0f, param_1, 0);
```
LagoLunatic: Are the `(u8)` `& 0xFF` and `(u32)` all necessary for this to match, or does it still match with just one of them?

src/JSystem/JAudio/JAISound.cpp
```diff
@@ -435,106 +509,304 @@ void JAISound::setTrackInterruptSwitch(u8 param_1, u8 param_2) {
...
+    }
 }
 
 /* 8029A950-8029AA34       .text setStreamInterPan__8JAISoundFUcfUl */
-void JAISound::setStreamInterPan(u8, f32, u32) {
-    /* Nonmatching */
+void JAISound::setStreamInterPan(u8 param_1, f32 param_2, u32 param_3) {
+    int iVar2;
+
+    if (((this->mSoundID & 0xc0000000) == 0xc0000000) && getStreamParameter()) {
```
LagoLunatic: 0xc0000000 looks like another enum

src/JSystem/JAudio/JAISound.cpp
```diff
@@ -435,106 +509,304 @@ void JAISound::setTrackInterruptSwitch(u8 param_1, u8 param_2) {
...
+        iVar2 = getStreamParameter()->pan[param_1].set(param_2, param_3);
+        if (iVar2 == 1) {
+            JAInter::StreamParameter* param = getStreamParameter();
+            param->mPanFlags = param->mPanFlags | 1 << (param_1 & 0xff);
+        }
+        if (getStreamParameter()->mUpdateData && iVar2 != 2) {
+            JAInter::StreamParameter* param = getStreamParameter();
+            param->mUpdateData->mActiveTrackFlag = param->mUpdateData->mActiveTrackFlag | 0x80000;
```
LagoLunatic: 0x80000 looks like another enum.

Also you can shorten `x = x | y` into just `x |= y` for simplicity. (Ignore if you already tried that and it doesn't match.)

src/JSystem/JAudio/JAIBasic.cpp
PJB3005: If I'm reading the debug map correctly, then the signature for that ctor is indeed `Actor(Vec* p1, Vec* p2, u32 p3, void* p4)`. That said, I suspect it might instead be something like `Actor(Vec* p1, Vec* p2, BOOL p3, void* p4)` and then an internal branch is constant folded away so we can't really tell. I guess I'll fix it like that.

The `void*` I think they're just using as a generic userdata-type field, looking at some of the code I've been able to trace through in Ghidra. So I'm assuming it's correct to keep as-is.

src/JSystem/JAudio/JAISound.cpp
PJB3005: Fixed these casts to the best I've been able to infer based on the existing code Ghidra was able to find touching the fields. I decided to make `field_0x1c` a u32 since that seems more common than the (likely) `void*` in the `Actor` struct. See also my other comment.

src/JSystem/JAudio/JAIBasic.cpp
LagoLunatic: Sorry, I think there was a miscommunication. When I said "JAInter::Actor is currently wrong" I didn't mean the ctor (which is from the symbol maps and thus was definitely correct), I meant the actual fields in that class (which are just a guess) were wrong because they were making you add casts that shouldn't be there:
https://github.com/zeldaret/tww/blob/6aa7ba91f3ae3cba6aa584970dfdb3f26f6def5d/include/JSystem/JAudio/JAIConst.h#L23-L27

If my explanation is unclear then don't worry about fixing it, I'll check the casts out later. For now just revert that last commit with the `BOOL` as that contradicts the symbol map.

PJB3005: I did fix the fields too, yes. Though I just realized that `BOOL` isn't the same as the `u32` when mangled so that can't be it. Still I think the answer is likely that it's some sort of constant fold on the `u32` param.

LagoLunatic: Okay, I pushed a commit to your PR to show you what I meant: https://github.com/zeldaret/tww/pull/960/changes/d02f37fe25aba0a10fff3b1dbaf5698dd8a63c4d

You were casting a `u32` to `void*`, then casting it back to `u32` later to use it. By simply fixing the field types like in this commit, you don't need either of those casts. Does that make sense?

AH64-dll:
Thanks for the JAI work. Two things when you get a chance: (1) the PR currently has a merge conflict — could you rebase onto current main? (2) The green CI is from ~4 months ago; a fresh all-4-build run after the rebase would confirm everything still holds, especially since JAudio headers are shared across many TUs.

PR #962, opened 2026-04-22
d_a_gm - 99% matching

123456789758:


AH64-dll:
Nice work on gm. One completeness ask: the PR body is empty, and the configure.py flip is `MatchingFor("D44J01")` — meaning CI verifies the rel only on the demo build, while the source carries a documented `/* Nonmatching - retail-only regalloc */` marker. That's an honest partial (like the saku single-region flip), but the flip shape + what's verified where deserves a sentence or two in the description: which versions are checksummed, which function remains nonmatching on retail, and why the demo build matches. Also noting the scope additions are fine: the `m31E` → `m031E` member rename updates both d_a_gm.cpp and its one external caller d_a_ks.cpp consistently (codegen-neutral), and the `setEmitterRotation` inline implementation in JPAEmitter.h is contained — no other TU calls it. Just the body explanation, then this is ready.

AH64-dll:
One tooling tip for the remaining holdouts (esp. `useHeapInit`): the constant-pooling pattern there (multiple distinct float constants pooled into a `.rodata.0` base register while the original emits per-constant lis/lfs) is the signature of an mwcc **version** mismatch, not source structure. Some original TWW TUs were built with `GC/1.3.2r` instead of the project-default `GC/1.3.2` — flipping the compiler version for the TU makes those functions match instantly (verified org-side on d_a_mozo: anime_proc 1%→100%, towait_proc_init 3%→100%). Cheap test before fighting the source further: compile the TU with `build/compilers/GC/1.3.2r/mwcceppc.exe` and objdiff-swap the `.o`. The per-actor override is a one-line `ActorRel(..., mw_version="GC/1.3.2r")` in configure.py (passthrough already landed).

PR #968, opened 2026-05-04
D_a_obj_magmarock 100% matching

noskap:
ninja succeeds on both GZLE01 and kiosk


noskap:
so it turns out the vtable does match. my objdiff was failing to build it for some reason.

AH64-dll:
Nice magmarock work. Two asks before this can merge: (1) the PR currently has a merge conflict — could you rebase onto current main? (2) The green CI is from mid-May (~3 months old); a fresh run after the rebase will confirm the matching still holds on today's tree.

AH64-dll:
heads-up: the original TU shows the mwcc GC/1.3.2r constant-pooling signature — 6 per-constant `lis`/`lfs` pairs in `play_anim` (vs project default 1.3.2 which pools constants under one base register). Verified case: d_a_mozo only snapped to 100% after compiling with 1.3.2r. If `play_anim` is stuck at 1.45%, try compiling the TU with `build/compilers/GC/1.3.2r/mwcceppc.exe` and objdiff-swap the .o; if it jumps to 100%, the flip needs the `mw_version` override on the ActorRel (`ActorRel(Matching, "d_a_d_a_obj_magmarock", mw_version="GC/1.3.2r")` — passthrough is on main).

PR #969, opened 2026-05-04
D_a_magma kiosk 100% matching

noskap:
Noticed d_a_magma kiosk was unmatched so I thought I'd take a shot at it.


AH64-dll:
Nice work on the kiosk difference — the VERSION-guarded dtor (inline dRes_control_c::deleteRes on the demo build vs the dComIfG_resDelete call on retail) and the MatchingFor(3-retail) → Matching flip follow exactly how other demo-divergent actors are handled. One ask: the green CI is from mid-May (~3 months old) and the tree has moved a lot since — could you rebase onto current main and let CI re-run so the full-Matching claim (now covering D44J01) is verified fresh?

noskap:
> Nice work on the kiosk difference — the VERSION-guarded dtor (inline dRes_control_c::deleteRes on the demo build vs the dComIfG_resDelete call on retail) and the MatchingFor(3-retail) → Matching flip follow exactly how other demo-divergent actors are handled. One ask: the green CI is from mid-May (~3 months old) and the tree has moved a lot since — could you rebase onto current main and let CI re-run so the full-Matching claim (now covering D44J01) is verified fresh?

Pretty sure you're an ai agent or something but either way this pr is from months ago and I've moved on to other projects. Anyone here is free to use my changes, or not, I don't mind

AH64-dll:
FYI: I adopted your changes into PR #1155 (d_a_magma kiosk 100% matching) — same VERSION-guarded dtor + Matching flip, rebased on current main and re-verified (GZLE01 + D44J01 rels both match). Thanks for releasing them!

PR #970, opened 2026-05-04
d_a_npc_so

### LINKED ISSUE #227: d_a_npc_so
NpcSo

Kelebek1:
Closes #227

### DIFF
--- a/include/m_Do/m_Do_hostIO.h
+++ b/include/m_Do/m_Do_hostIO.h
@@ -109,7 +109,7 @@ public:
 #else
     virtual ~mDoHIO_entry_c() {}
 #endif
-};
+}; // size = 0x8
 
 extern mDoHIO_root_c mDoHIO_root;
 
--- a/src/d/actor/d_a_npc_so.cpp
+++ b/src/d/actor/d_a_npc_so.cpp
@@ -5,397 +5,68 @@
 
 #include "d/dolzel_rel.h" // IWYU pragma: keep
 #include "d/actor/d_a_npc_so.h"
+#include "d/d_npc.h"
 #include "d/d_procname.h"
+#include "d/d_snap.h"
+#include "d/d_s_play.h"
+#include "d/actor/d_a_ship.h"
+#include "d/actor/d_a_player.h"
+#include "d/actor/d_a_tag_so.h"
+#include "d/actor/d_a_esa.h"
 #include "d/d_priority.h"
-
-/* 000000EC-000002A4       .text __ct__14daNpc_So_HIO_cFv */
-daNpc_So_HIO_c::daNpc_So_HIO_c() {
-    /* Nonmatching */
-}
-
-/* 000002A4-000002D0       .text searchEsa_CB__FPvPv */
-void searchEsa_CB(void*, void*) {
-    /* Nonmatching */
-}
-
-/* 000002D0-000002F8       .text _searchEsa__10daNpc_So_cFP10fopAc_ac_c */
-void daNpc_So_c::_searchEsa(fopAc_ac_c*) {
-    /* Nonmatching */
-}
-
-/* 000002F8-00000344       .text nodeControl_CB__FP7J3DNodei */
-static BOOL nodeControl_CB(J3DNode*, int) {
-    /* Nonmatching */
-}
-
-/* 00000344-00000474       .text _nodeControl__10daNpc_So_cFP7J3DNodeP8J3DModel */
-void daNpc_So_c::_nodeControl(J3DNode*, J3DModel*) {
-    /* Nonmatching */
-}
-
-/* 000004B0-000004DC       .text searchTagSo_CB__FPvPv */
-void searchTagSo_CB(void*, void*) {
-    /* Nonmatching */
-}
-
-/* 000004DC-0000054C       .text _searchTagSo__10daNpc_So_cFP10fopAc_ac_c */
-void daNpc_So_c::_searchTagSo(fopAc_ac_c*) {
-    /* Nonmatching */
-}
-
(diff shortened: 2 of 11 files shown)


src/d/actor/d_a_npc_so.cpp
```diff
@@ -5,367 +5,1807 @@
...
+    J3DModelData* modelData = static_cast<J3DModelData*>(dComIfG_getObjectRes(m_arc_name, SO_BDL_SO));
+    JUT_ASSERT(DEMO_SELECT(537, 509), modelData != NULL);
+
+    mpMorf = new mDoExt_McaMorf(modelData, NULL, NULL, NULL, ~J3DFrameCtrl::EMode_NONE, 1.0f, 0, -1, 1, NULL, 0x80000, 0x11020022);
+    if (mpMorf == NULL || mpMorf->getModel() == NULL) {
+        return FALSE;
+    }
+    mpMorf->getModel()->setUserArea(reinterpret_cast<u32>(this));
+
+#if VERSION > EVRSION_DEMO
```
LagoLunatic, line 317: EVRSION

include/m_Do/m_Do_hostIO.h
```diff
@@ -109,7 +109,7 @@ class mDoHIO_entry_c : public JORReflexible {
 #else
     virtual ~mDoHIO_entry_c() {}
 #endif
-};
+}; // size = 0x8
```
LagoLunatic, line 112: These size comments are usually for the USA retail version, but here the comment is only correct for the demo version which is misleading

src/d/actor/d_a_npc_so_cut.inc
```diff
@@ -2,228 +2,1043 @@
...
+    mBDA = 1;
+    mAFC = 0.0f;
+    speedF = 0.0f;
+#endif
 }
 
 /* 00005C6C-000060FC       .text cutEatesaFirstProc__10daNpc_So_cFv */
 void daNpc_So_c::cutEatesaFirstProc() {
-    /* Nonmatching */
+    fopAc_ac_c* partner = dComIfGp_getPlayer(1);
```
LagoLunatic, line 333: The player array is only length 1, so getPlayer(1) is an out of bounds access. I think you meant to call `dComIfGp_getLinkPlayer()` here.

Also, the local variable name `partner` is wrong, `dComIfGp_getLinkPlayer` returns a pointer to Link, not Link's partner.

These two things also apply to several other places in this PR.

src/d/actor/d_a_npc_so_cut.inc
```diff
@@ -2,228 +2,1043 @@
...
+
+        f32 fVar1 = current.pos.y;
+        if (speed.y < 1.0) {
+            speed.y = 1.0f;
+        }
+
+        if (cLib_chasePosXZ(fopAcM_GetPosition_p(this), mBB0, l_HIO.m90) != 0) {
+            current.pos.y = fVar1;
+            if (l_HIO.m94 == 0) {
+                g_dComIfG_gameInfo.play.mFmapOpen = 2;
```
LagoLunatic, line 469: This is probably the unimplemented inline `dComIfGp_fmapOpenFishOn`

src/d/actor/d_a_npc_so.cpp
```diff
@@ -5,367 +5,1807 @@
...
+    s32 jntNo = pJnt->getJntNo();
+
+    mDoMtx_stack_c::copy(pModel->getAnmMtx(jntNo));
+    if (jntNo == m_jnt.getHeadJntNum()) {
+        cXyz local_2C(0.0f, 0.0f, 0.0f);
+        cXyz local_38(24.0f, -16.0f, 0.0f);
+        local_38.y += DEMO_SELECT(REG12_F(1), 0.0f);
+
+        mDoMtx_stack_c::multVec(&local_2C, &mB60);
+        if (DEMO_SELECT(REG12_S(1) != 0, false)) {
```
LagoLunatic, line 196: Using DEMO_SELECT in an if statement with false on one side is kind of confusing to read. If you don't want to use a normal `#if` directive here you could try something like:
```suggestion
        if (VERSION == VERSION_DEMO && REG12_S(1) != 0) {
```

src/d/actor/d_a_npc_so.cpp
```diff
@@ -5,367 +5,1807 @@
...
+            mDoMtx_stack_c::XrotM(REG12_S(2));
+            mDoMtx_stack_c::YrotM(REG12_S(3));
+            mDoMtx_stack_c::ZrotM(REG12_S(4));
+        } else {
+            mDoMtx_stack_c::YrotM((s16)m_jnt.getHead_y());
+            mDoMtx_stack_c::ZrotM((s16)m_jnt.getHead_x());
+        }
+        mDoMtx_stack_c::multVec(&local_38, &mB54);
+    } else if (jntNo == m_jnt.getBackboneJntNum()) {
+        if (DEMO_SELECT(REG12_S(0) != 0, false)) {
```
LagoLunatic, line 206: Same here

AH64-dll:
Solid partial — honest Nonmatching retention, no flip. Two items:

1. HEADLINE: the PR is CONFLICTING and 99 days old — it needs a rebase and fresh CI before anything else.
2. QUESTION: the JGeometry.h change removes the `(f32)` casts in TVec3<f32>::set. Note the template is declared `template<typename f32>` — the template parameter shadows the typedef, so inside this function `f32` is the template param, and for the TVec3<f32> instantiation the casts are no-ops (codegen-neutral). BUT for any other instantiation (e.g. TVec3<int>), removing the cast changes behavior (no truncation). JGeometry.h is a shared DOL header used across many TUs — could you confirm no other instantiations rely on the cast, and consider whether this cleanup belongs in this PR at all?

Neither blocks the partial.

PR #979, opened 2026-05-08
d_demo code matching

Lord-McSweeney:


AH64-dll:
Thanks for the PR! Two currency asks so this can land: (1) the branch is currently conflicting with main (mergeStateStatus DIRTY) — please rebase; (2) the CI run is ~3 months old (May 8); after rebasing, a fresh green run on all four builds is needed since d_demo.h / d_demo.cpp are shared framework files whose changes affect every demo-using actor's codegen. Happy to re-review once green.

PR #981, opened 2026-05-08
Use joint enums in several different places

123456789758:


LagoLunatic:
Please commit the script you used to generate these. I imagine it's based on the one Roeming and I wrote for TP but I'm pretty sure that won't work out of the box so you must have made some changes to it.

Also, I was putting off adding the joint enums to TWW because I wanted to move the res enums into version-specific folders to fix the mistakes I made when generating them in the first place. I'd rather that both of these huge enum changes are done at the same time since they're closely related. Adding the joint enums into the existing non-version-specific folders means that if any of the joint indexes differ between versions (I don't know if there are or not) there will be errors in this PR that will need to be fixed later.

LagoLunatic:
Interesting, so the approach you used was to generate `#if` directives automatically via the script?

But also it looks like your most recent push removed those `#if` directives?
image
So now I'm confused, are version differences accounted for in this PR or not? The PR is too large to look through manually and you haven't provided a description.

123456789758:
That one was only changed because there weren't any meaningful differences. There seem to only be a few files that actually have notable differences (such as bgn and hseki).

LagoLunatic:
In #1037 I fixed all the res enums and also added joint enums at the same time in version-specific folders the way TP decomp does it. We decided to do it that way so that modding support will be easier to add in the future.

But I haven't gone through and actually used the joint enums everywhere yet though, so could you cherrypick that "Use joint enums in several different places" commit into a separate PR? That one is probably mergable and would clean up the code a lot.

PR #992, opened 2026-05-11
d_a_grid: 100% equivalent

banteg:
## Summary

Decompiles `d/actor/d_a_grid.cpp` and adds the actor/header structure for `daGrid_c`, `daHo_packet_c`, and `daHo_HIO_c`.

This is marked `Equivalent` because the TU is function/data complete locally, but the linked DOL still differs by weak destructor ordering.

## Status

Local matching status for `d_a_grid`:

```text
matched_functions: 20 / 20
matched_code:      10572 / 10572
matched_data:      3100 / 3100
```

The remaining linked-DOL difference is weak function order only:

```text
target:
__sinit_d_a_grid_cpp
__dt__10daHo_HIO_cFv
__dt__13daHo_packet_cFv
Eye__9dCamera_cFv

current:
__sinit_d_a_grid_cpp
__dt__13daHo_packet_cFv
__dt__10daHo_HIO_cFv
Eye__9dCamera_cFv
```

The packet and HIO destructors themselves match; only their weak emission/link order differs.

## Notes

I avoided fakematching and inline asm here. Most of the final work was source-shape cleanup around `create` and `ho_move`, with local scripted variant testing to narrow down lifetime/expression-order issues instead of adding artificial code.

I also tried the non-fakematch destructor-order fixes I could think of, including explicit/implicit packet destructor changes, out-of-class inline destructor definitions, moving destructor definitions between header/cpp, and class-order variants. Those did not solve the weak order.

ref #578

### DIFF
--- a/config/D44J01/config.yml
+++ b/config/D44J01/config.yml
@@ -3226,19 +3226,15 @@ extract:
 - symbol: l_pos!.data:0x8037F358
   binary: assets/l_grid_pos.bin
   header: assets/l_grid_pos.h
-  rename: l_grid_pos
 - symbol: l_texCoord!.data:0x8037F754
   binary: assets/l_grid_texCoord.bin
   header: assets/l_grid_texCoord.h
-  rename: l_grid_texCoord
 - symbol: l_DL!.data:0x8037FA00
   binary: assets/l_grid_DL.bin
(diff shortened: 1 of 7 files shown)


banteg:
withdrawing this. decided not to work on the repo.

PR #995, opened 2026-05-12
d_a_himo2: improve match to 57/59 functions

banteg:
## Summary

Improves `d/actor/d_a_himo2.cpp` and tightens the `himo2_class` actor layout.

This gets the TU very close locally: all data is matched, 57 / 59 functions are matched, and the remaining two functions are narrow register-allocation issues.

## Status

Local matching status for `d_a_himo2`:

```text
unit fuzzy:       99.929016
matched_functions: 57 / 59
matched_data:      836 / 836
```

Remaining nonmatching functions:

```text
daHimo2_Draw      99.89209
new_himo2_move    99.86748
```

The remaining diffs are allocator-shape only:

- `new_himo2_move`: early callee-saved GPR ownership around `dComIfGp_getAttention()`
- `daHimo2_Draw`: final 16-point loop volatile register choice

## Notes

I avoided fakematching, inline asm, and artificial no-op code. Most of the final work was source-shape cleanup around rope movement, draw/update loops, spin draw, and camera/target math.

I also ran targeted scripted probes for the remaining allocator issues, including declaration/lifetime variants, `Lockon()` result spelling, bool/full-width no-lockon forms, final-loop type/init/order variants, manual sine lookup spellings, and count-controlled loop forms. Those either stayed neutral or regressed.

ref #579

### DIFF
--- a/include/d/actor/d_a_himo2.h
+++ b/include/d/actor/d_a_himo2.h
@@ -64,8 +64,7 @@ struct himo2_class {
     /* 0x2180 */ fpc_ProcID m2180;
     /* 0x2184 */ f32 m2184;
     /* 0x2188 */ f32 m2188;
-    /* 0x218C */ fopAc_ac_c* m218C;
-    /* 0x2190 */ u8 m2190[0x24AC - 0x2190];
+    /* 0x218C */ fopAc_ac_c* m218C[200];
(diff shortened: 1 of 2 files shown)


banteg:
withdrawing this. decided not to work on the repo.

PR #997, opened 2026-05-12
d_a_obj_jump

marian-m12l:
Not quite there yet. 27 functions match 100%. 4 are 95+%. 1 is 63%. I've only checked GZLP01.
I'll keep looking into it. Also there's some ugly pointer arithmetic that needs to be fixed.

Issue #416


marian-m12l:
I've improved readability substantially, and got the constructor to match by improving class fields definition. Same for attr struct.
No more pointers hack :)
rodata and data also match 100%.
Remaining functions: Create (96%), jnodeCB_lower (98%), Execute and Draw (99%). Remaining mismatch seem to be a mix of stack size, regalloc, and data location, but I don't know how to fix those.

PR #1007, opened 2026-05-18
d_a_obj_YLzou

### LINKED ISSUE #373: d_a_obj_YLzou
YLzou

jakecelentano:
Decompiles `d_a_obj_YLzou` (Hyrule Castle moving Link-statue staircase). Closes #373.

All 49 functions are byte-identical — objdiff reports 100.0% fuzzy match, 49/49 functions at 100%. Full `ninja` is "416 files OK".

Marked `Equivalent` rather than `Matching`: the TU pulls one extra `std::sqrtf` weak COMDAT (`_half`/`_three` local statics) via a header inline that the original's include set avoids; all *emitted* code matches exactly. Happy to dig further if a reviewer knows the canonical include that sidesteps it.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1681,7 +1681,7 @@ config.libs = [
     ActorRel(NonMatching, "d_a_obj_Itnak"),
     ActorRel(NonMatching, "d_a_obj_Vds"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_obj_Vteng"),
-    ActorRel(NonMatching, "d_a_obj_YLzou"),
+    ActorRel(Equivalent, "d_a_obj_YLzou"), # all 49 fns byte-identical; extra std::sqrtf weak COMDAT from a header inline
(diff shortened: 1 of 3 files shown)


Kelebek1:
You have some wrong floats used, which is what's causing the issues with rodata and some functions.

Objdiff should show you that some functions are at 99%, not 100%. If you look at them, they have some coloured symbols, showing you there's some issue with them. It may be the wrong data section (data vs rodata vs bss) or their offset may be wrong, so they're in the wrong place, or it may just be the value.

If you mouse-over the symbols, objdiff will show you the values, like here in `move_ylzou_demo_move_act_proc` you can see you're using -170.0f when it should be -680.0f
Untitled

Same issue in `open_wait_act_init_proc` and `go_up_stairs_demo_move_start_wait_act_init_proc` with the same -170/-680 difference. Fixing those makes the rodata match 100% and you can mark this actor as Matching and it builds OK.

The extra sqrt and data symbols are supposed to be there, they're unused so they get removed at link time and it doesn't affect matching, so you can ignore them.

PR #1013, opened 2026-05-19
d_a_warpls

LegendofLuke03:


Teotia444:
this is about all the inlines i could catch. you should also def try to match demo, most diffs i could catch are pretty simple to adjust/gets resolved from inlines. lmk if you need help with that

Issue #410, opened 2023-09-28
d_a_obj_hlift


magcius:
Hlift, Hliftb

marian-m12l:
Looking into it

PR #1015, opened 2026-05-21
d_a_obj_hlift

marian-m12l:
Almost there.

Issue #410

### DIFF
--- a/src/d/actor/d_a_obj_hlift.cpp
+++ b/src/d/actor/d_a_obj_hlift.cpp
@@ -7,20 +7,23 @@
 #include "d/actor/d_a_obj_hlift.h"
 #include "d/d_procname.h"
 #include "d/d_priority.h"
-
-int L_time_lag_num;
+#include "d/res/res_hlift.h"
 
 namespace daObjHlift {
 
+namespace {
+    int L_time_lag_num;
+    static const Attr_c L_attr = { 30.0f, 50.0f, 10.0f, 2.0f, 0.3f, 1.2f, 250.0f, 10, 16384, 2.5f, 0x00, 0x02, 0x0a, 0x08, 0x07, 0x00, 0x00, 0x00 };
+}
 Mtx Act_c::M_tmp_mtx;
-u8 M_lift_move_flag;
+u8 Act_c::M_lift_move_flag;
 int Act_c::M_control_id = -1;
 const char Act_c::M_arcname[] = "Hlift";
 const char Act_c::M_evname[] = "Hlift_up";
 const s16 Act_c::M_up_dist[] = { 125, 250, 375, 500, 625, 750, 875, 1000 };
 const Act_c::size_data Act_c::M_data_size[2] = {
-    { 4, 9, 5888 },
-    { 5, 10, 5888 }
+    { HLIFT_BDL_HLIFT, HLIFT_DZB_HLIFT, VERSION_SELECT(32768, 5888, 5888, 5888) },
+    { HLIFT_BDL_HLIFTB, HLIFT_DZB_HLIFTB, VERSION_SELECT(32768, 5888, 5888, 5888) }
 };
 
 }
@@ -30,7 +33,7 @@ BOOL daObjHlift::Act_c::CreateHeap() {
     J3DModelData* model_data = (J3DModelData*)dComIfG_getObjectRes(M_arcname, M_data_size[field_0x2DC].field_0x00);
     JUT_ASSERT(288, model_data != NULL);
     mModel1 = mDoExt_J3DModel__create(model_data, 0x80000, 0x11000022);
-    J3DModelData* model_data_c = (J3DModelData*)dComIfG_getObjectRes(M_arcname, 6);
+    J3DModelData* model_data_c = (J3DModelData*)dComIfG_getObjectRes(M_arcname, HLIFT_BDL_HLIFTC);
     JUT_ASSERT(298, model_data_c != NULL);
     mModel2 = mDoExt_J3DModel__create(model_data_c, 0x80000, 0x11000022);
     return mModel1 != NULL && mModel2 != NULL;
--- a/src/d/actor/d_a_obj_hlift.cpp
+++ b/src/d/actor/d_a_obj_hlift.cpp
@@ -264,8 +264,19 @@ void daObjHlift::Act_c::vib_set() {
 /* 00000CA8-00000D70       .text vib_proc__Q210daObjHlift5Act_cFv */
 void daObjHlift::Act_c::vib_proc() {
     if (field_0x2E8 > 0) {
+#if VERSION == VERSION_DEMO
+        float fVar4 = L_attr.field_0x20;
+        float fVar2 = 0.5f;
         float fVar1 = (L_attr.field_0x1C - field_0x2E8);
-        field_0x2EC = (L_attr.field_0x20 * (0.5f + fVar1 * (0.5f / L_attr.field_0x1C))) * cM_ssin(field_0x2EA);
+        float fVar3 = (fVar2 / L_attr.field_0x1C);
+#else
+        float fVar1 = (L_attr.field_0x1C - field_0x2E8);
+        float fVar2 = 0.5f;
+        float fVar3 = (fVar2 / L_attr.field_0x1C);
+        float fVar4 = L_attr.field_0x20;
+#endif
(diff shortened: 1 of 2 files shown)


src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -8,124 +8,316 @@
...
+
+}
+
 /* 00000078-000001BC       .text CreateHeap__Q210daObjHlift5Act_cFv */
 BOOL daObjHlift::Act_c::CreateHeap() {
-    /* Nonmatching */
+    J3DModelData* model_data = (J3DModelData*)dComIfG_getObjectRes(M_arcname, M_data_size[field_0x2DC].field_0x00);
+    JUT_ASSERT(288, model_data != NULL);
+    mModel1 = mDoExt_J3DModel__create(model_data, 0x80000, 0x11000022);
+    J3DModelData* model_data_c = (J3DModelData*)dComIfG_getObjectRes(M_arcname, 6);
```
LagoLunatic: FYI, the repo has enums for every res index, so if you include the relevant header (`res_hlift.h` in this case, since that's the arcname) you can use these instead of magic numbers:
```suggestion
    J3DModelData* model_data_c = (J3DModelData*)dComIfG_getObjectRes(M_arcname, HLIFT_BDL_HLIFTC);
```

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -8,124 +8,316 @@
...
+
+Mtx Act_c::M_tmp_mtx;
+u8 M_lift_move_flag;
+int Act_c::M_control_id = -1;
+const char Act_c::M_arcname[] = "Hlift";
+const char Act_c::M_evname[] = "Hlift_up";
+const s16 Act_c::M_up_dist[] = { 125, 250, 375, 500, 625, 750, 875, 1000 };
+const Act_c::size_data Act_c::M_data_size[2] = {
+    { 4, 9, 5888 },
+    { 5, 10, 5888 }
```
LagoLunatic: And these (4, 9, 5, 10) are also res indexes from the same enum.

marian-m12l:
@LagoLunatic Thanks for the feedback! I've updated with the enums pushed some more functions to 100%. `chk_demo_end` has an instruction swap I cannot figure out, and `vib_proc` still has what looks like regalloc issue in demo version.

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -7,125 +7,338 @@
...
+}
+Mtx Act_c::M_tmp_mtx;
+u8 Act_c::M_lift_move_flag;
+int Act_c::M_control_id = -1;
+const char Act_c::M_arcname[] = "Hlift";
+const char Act_c::M_evname[] = "Hlift_up";
+const s16 Act_c::M_up_dist[] = { 125, 250, 375, 500, 625, 750, 875, 1000 };
+const Act_c::size_data Act_c::M_data_size[2] = {
+    { HLIFT_BDL_HLIFT, HLIFT_DZB_HLIFT, VERSION_SELECT(32768, 5888, 5888, 5888) },
+    { HLIFT_BDL_HLIFTB, HLIFT_DZB_HLIFTB, VERSION_SELECT(32768, 5888, 5888, 5888) }
```
LagoLunatic: If a version difference is only between demo and retail with all 3 retail being the same you can use this macro instead to simplify
```suggestion
    { HLIFT_BDL_HLIFT, HLIFT_DZB_HLIFT, DEMO_SELECT(0x8000, 0x1700) },
    { HLIFT_BDL_HLIFTB, HLIFT_DZB_HLIFTB, DEMO_SELECT(0x8000, 0x1700) }
```

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -7,125 +7,338 @@
...
+            mode_lower_init();
+        }
+    }
 }
 
 /* 000009DC-00000AB0       .text mode_demoreq_init__Q210daObjHlift5Act_cFQ310daObjHlift5Act_c6Mode_e */
-void daObjHlift::Act_c::mode_demoreq_init(daObjHlift::Act_c::Mode_e) {
-    /* Nonmatching */
+void daObjHlift::Act_c::mode_demoreq_init(daObjHlift::Act_c::Mode_e next) {
+    JUT_ASSERT(VERSION_SELECT(573, 575, 575, 575), (next == Mode_U_L) || (next == Mode_L_U));
```
LagoLunatic: Same here

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -7,125 +7,338 @@
...
+        float fVar4 = L_attr.field_0x20;
+        float fVar2 = 0.5f;
+        float fVar1 = (L_attr.field_0x1C - field_0x2E8);
+        float fVar3 = (fVar2 / L_attr.field_0x1C);
+#else
+        float fVar1 = (L_attr.field_0x1C - field_0x2E8);
+        float fVar2 = 0.5f;
+        float fVar3 = (fVar2 / L_attr.field_0x1C);
+        float fVar4 = L_attr.field_0x20;
+#endif
```
LagoLunatic: I doubt this version difference is real since you just changed the order of the lines, something the devs are unlikely to do between versions. The actual cause of the instructions being different order is likely just that demo uses different compiler flags.

Try getting rid of the if directive and using the demo's order for retail. I bet it will match both.

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -7,125 +7,338 @@
...
-    /* Nonmatching */
+#if VERSION == VERSION_DEMO
+    float z = current.pos.z;
+#endif    
+    float y1 = current.pos.y + field_0x2EC;
+    float y2 = y1 + L_attr.field_0x00;
+#if VERSION > VERSION_DEMO
+    float z = current.pos.z;
+#endif
+    mDoMtx_stack_c::transS(current.pos.x, y2, z);
```
LagoLunatic: Same here, try writing it the natural way and try to see if it matches all versions with no `#if`
```suggestion
    mDoMtx_stack_c::transS(current.pos.x, current.pos.y + field_0x2EC + L_attr.field_0x00, current.pos.z);
```

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -7,125 +7,338 @@
...
+
+namespace daObjHlift {
+
+namespace {
+    int L_time_lag_num;
+    static const Attr_c L_attr = { 30.0f, 50.0f, 10.0f, 2.0f, 0.3f, 1.2f, 250.0f, 10, 16384, 2.5f, 0x00, 0x02, 0x0a, 0x08, 0x07, 0x00, 0x00, 0x00 };
+}
+Mtx Act_c::M_tmp_mtx;
+u8 Act_c::M_lift_move_flag;
+int Act_c::M_control_id = -1;
```
LagoLunatic: Enum
```suggestion
fpc_ProcID Act_c::M_control_id = fpcM_ERROR_PROCESS_ID_e;
```

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -7,125 +7,338 @@
...
+    fopAcM_setCullSizeBox(
+        this,
+        -151.0f,
+        -1005.0f,
+        -151.0f,
+        151.0f,
+        1.0f,
+        151.0f
+    );
+    if (M_control_id == 0xffffffff) {
```
LagoLunatic:
```suggestion
    if (M_control_id == fpcM_ERROR_PROCESS_ID_e) {
```

src/d/actor/d_a_obj_hlift.cpp
```diff
@@ -7,125 +7,338 @@
...
+        JUT_ASSERT(366, (phase_state == cPhs_COMPLEATE_e) || (phase_state == cPhs_ERROR_e));
+    }
+    return phase_state;
 }
 
 /* 00000408-00000430       .text Delete__Q210daObjHlift5Act_cFv */
 BOOL daObjHlift::Act_c::Delete() {
-    /* Nonmatching */
+    if (M_control_id == base.base.mBsPcId) {
+        M_control_id = -1;
```
LagoLunatic:
```suggestion
        M_control_id = fpcM_ERROR_PROCESS_ID_e;
```

marian-m12l:
Everything's spot on! :100:

PR #1018, opened 2026-05-24
Matching Retail d_a_waterfall

### LINKED ISSUE #485: d_a_waterfall
Yswtr00

tetrisdev:
closes #485

### DIFF
--- a/src/d/actor/d_a_waterfall.cpp
+++ b/src/d/actor/d_a_waterfall.cpp
@@ -3,10 +3,6 @@
 // Translation Unit: d_a_waterfall.cpp
 //
 
-/*
-* Likely: Entrance waterfall in Tower of the Gods.
-*/
-
 #include "d/dolzel_rel.h" // IWYU pragma: keep
 #include "d/actor/d_a_waterfall.h"
 #include "d/d_bg_s_func.h"
@@ -22,7 +18,6 @@ const s16 daWfall_c::m_heapsize[] = {0x35a0, 0x4870};
 
 /* 00000078-000000F0       .text _delete__9daWfall_cFv */
 bool daWfall_c::_delete() {
-    /* Nonmatching */
     mPCallBack.end();
     mPCallBack2.end();
     dComIfG_resDelete(&mPhs, m_arcname);
@@ -58,7 +53,7 @@ BOOL daWfall_c::CreateHeap() {
         return FALSE;
     }
 
-    if (mSomeFlag == true) {
+    if (mType == true) {
         modelData = (J3DModelData*)dComIfG_getObjectRes(m_arcname, WFALL_BDL_YSMNM00);
         JUT_ASSERT(0x129, modelData != NULL);
         mModelMinamo = mDoExt_J3DModel__create(modelData, 0x80000, 0x11000222);
@@ -86,7 +81,7 @@ BOOL daWfall_c::CreateHeap() {
 /* 0000048C-00000708       .text CreateInit__9daWfall_cFv */
 void daWfall_c::CreateInit() {
     cullMtx = mModelWater->getBaseTRMtx();
-    if (mSomeFlag == true) {
+    if (mType == true) {
         fopAcM_setCullSizeBox(this, -1000.0f, 0.0f, -200.0f, 1200.0f, 1000.0f, 3800.0f);
     } else {
         fopAcM_setCullSizeBox(this, -250.0f, 0.0f, 0.0f, 250.0f, 1000.0f, 800.0f);
@@ -104,14 +99,14 @@ void daWfall_c::CreateInit() {
         mode_wtr_off_init();
         mSePos = current.pos;
         mSomeCounter = 10;
-        if (mSomeFlag == true) {
+        if (mType == true) {
             mBrkAnm.setFrame(mBrkAnm.getEndFrame());
         }
     } else {
         mode_wtr_on_init();
         mSePos = current.pos;
         mSePos.y += 863.0f;
-        if (mSomeFlag == true) {
+        if (mType == true) {
             mBrkAnm.setFrame(0.0f);
         }
     }
@@ -124,8 +119,8 @@ cPhs_State daWfall_c::_create() {
 
     cPhs_State ret = dComIfG_resLoad(&mPhs, m_arcname);
     if (ret == cPhs_COMPLEATE_e) {
-        mSomeFlag = (fopAcM_GetParam(this) >> 8) & 0xf;
-        if (!fopAcM_entrySolidHeap(this, CheckCreateHeap, m_heapsize[mSomeFlag])) {
+        mType = (fopAcM_GetParam(this) >> 8) & 0xf;
+        if (!fopAcM_entrySolidHeap(this, CheckCreateHeap, m_heapsize[mType])) {
             return cPhs_ERROR_e;
         }
         CreateInit();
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_waterfall.cpp
```diff
@@ -3,104 +3,386 @@
...
+        waterHeight = splGrpChk.GetHeight();
+    }
+
+    return waterHeight;
 }
 
 /* 00001370-000013E0       .text set_se__9daWfall_cFv */
 void daWfall_c::set_se() {
-    /* Nonmatching */
+    mDoAud_seStart(0x701f, &mSePos, 0, dComIfGp_getReverb(fopAcM_GetRoomNo(this)));
```
LagoLunatic:
```suggestion
    mDoAud_seStart(JA_SE_ATM_WATER_GATE, &mSePos, 0, dComIfGp_getReverb(fopAcM_GetRoomNo(this)));
```

src/d/actor/d_a_waterfall.cpp
```diff
@@ -3,104 +3,386 @@
...
+    }
+    mode_proc_call();
+    set_mtx();
+    set_gate_mtx();
+    return true;
 }
 
 /* 00000C94-00000D20       .text mode_proc_call__9daWfall_cFv */
 void daWfall_c::mode_proc_call() {
     /* Nonmatching */
```
LagoLunatic: You can remove these nonmatching comments once the function is 100%

src/d/actor/d_a_waterfall.cpp
```diff
@@ -3,104 +3,386 @@
 // Translation Unit: d_a_waterfall.cpp
 //
 
+/*
+* Likely: Entrance waterfall in Tower of the Gods.
+*/
```
LagoLunatic: `d_a_obj_hha` is the TotG entrance waterfall. I think this actor is unused, or used for something currently unknown, so we should leave the description out until we figure it out.

src/d/actor/d_a_waterfall.cpp
```diff
@@ -3,104 +3,386 @@
...
+    mEvtIdx = dComIfGp_evmng_getEventIdx("GATECLOSE", 0xff);
+}
 /* 00000708-00000838       .text _create__9daWfall_cFv */
 cPhs_State daWfall_c::_create() {
-    /* Nonmatching */
+    fopAcM_SetupActor(this, daWfall_c);
+
+    cPhs_State ret = dComIfG_resLoad(&mPhs, m_arcname);
+    if (ret == cPhs_COMPLEATE_e) {
+        mSomeFlag = (fopAcM_GetParam(this) >> 8) & 0xf;
```
LagoLunatic: Missing inline: `daWfall_prm::getType(daWfall_c*)` from the debug map. Which also shows that the name of `mSomeFlag` should be `mType`.

src/d/actor/d_a_waterfall.cpp
```diff
@@ -3,104 +3,386 @@
...
+    }
+    cullSizeFar = 2.5f;
+    set_mtx();
+    mParticlePos1 = current.pos;
+    mParticlePos2 = current.pos;
+    mParticleAngle1 = current.angle;
+    mParticleAngle2 = current.angle;
+    dComIfGp_particle_set(dPa_name::ID_AK_SN_SIRENSUIRYU00, &mParticlePos1, &mParticleAngle1, NULL, 0xff, &mPCallBack, -1, NULL, NULL, NULL);
+    dComIfGp_particle_set(dPa_name::ID_AK_SN_SIRENSUIRYU01, &mParticlePos2, &mParticleAngle2, NULL, 0xff, &mPCallBack2, -1, NULL, NULL, NULL);
+    mWtrSwitch = fpcM_GetParam(this) & 0xff;
```
LagoLunatic: Same here, `daWfall_prm::getSwitchNo(daWfall_c*)`

src/d/actor/d_a_waterfall.cpp
```diff
@@ -3,104 +3,386 @@
...
 /* 00000EE8-00000FF0       .text setEmitter00Pos__9daWfall_cFv */
-void daWfall_c::setEmitter00Pos() {
-    /* Nonmatching */
+BOOL daWfall_c::setEmitter00Pos() {
+    BOOL ret = FALSE;
+    mParticlePos1 = current.pos;
+    float waterHeight = getWaterHeight();
+    if (waterHeight > mSePos.y) {
+        if (mPCallBack.getEmitter() != NULL) {
+            mPCallBack.getEmitter()->setStatus(1);
```
LagoLunatic: stopCreateParticle

src/d/actor/d_a_waterfall.cpp
```diff
@@ -3,104 +3,386 @@
...
+        }
+    } else {
+        if (mPCallBack.getEmitter() != NULL) {
+            mPCallBack.getEmitter()->clearStatus(1);
```
LagoLunatic: `playCreateParticle`

tetrisdev: Done

src/d/actor/d_a_waterfall.cpp
tetrisdev: Done

src/d/actor/d_a_waterfall.cpp
tetrisdev: changed

src/d/actor/d_a_waterfall.cpp
tetrisdev: Didn't quite understand what you meant, changed it to mSwitchNo?

src/d/actor/d_a_waterfall.cpp
LagoLunatic: You changed the field name, but the inline `daWfall_prm::getType(daWfall_c*)` is still missing.

src/d/actor/d_a_waterfall.cpp
LagoLunatic: Changing the field name is fine, but what I meant is that the `daWfall_prm::getSwitchNo(daWfall_c*)` inline should be implemented and used here.

src/d/actor/d_a_waterfall.cpp
LagoLunatic: There are several other places in this file where stopCreateParticle should be used too

src/d/actor/d_a_waterfall.cpp
LagoLunatic: There's another place in this file where playCreateParticle should be used

AH64-dll:
The matching work here is solid (MatchingFor GZLJ01/GZLE01/GZLP01, all Nonmatching markers cleared) — but the PR has aged out of the tree and needs currency work before it can merge:

1. Rebase: the configure.py diff is against an old base, and the four CI builds shown are from 2026-05-25 (79 days ago) — they're stale.
2. Res convention: `#include d/res/res_wfall.h` — that path no longer exists; the current convention is `#include res/Object/Wfall.h` (used by 274 merged actors). The old d/res/ generated-header layout was replaced by the assets/GZLE01/res/Object/ + configure-time headers.
3. Enums: the bare names `WFALL_BDL_YSWTR00`, `WFALL_BTK_YSWTR00`, `WFALL_BDL_HSUI1`, etc. don't exist in the current tree — the enum is `dRes_INDEX_WFALL_BDL_YSWTR00_e` etc. (assets/GZLE01/res/Object/Wfall.h:11-23). These need the dRes_INDEX_ prefix + _e suffix (or dRes_ID_* for the ID variants).

So the ask is: rebase + update to the current res include/enum convention + fresh CI. The matching itself was likely fine in May.

PR #1032, opened 2026-05-27
d_a_oq - 99% matching

123456789758:


AH64-dll:
Nice work on d_a_oq. One note: the `pluss_coming_point1()` inline in d_a_coming2.h (a matched actor) goes from an empty stub to a real body. I checked — nothing in-tree calls it, so the matched coming2 rel is unaffected (and CI green x4 confirms). Good improvement either way; just wanted to flag that shared-header inline changes on matched actors are the kind of thing we watch closely.

PR #1035, opened 2026-05-28
daObj_hsh_HIO_c constructor (100% match)

Armysbro911:
first attempt not entirely confident.

### DIFF
--- a/include/d/actor/d_a_obj_hsehi1.h
+++ b/include/d/actor/d_a_obj_hsehi1.h
@@ -69,12 +69,17 @@ public:
     /* Place member variables here */
 };
 
-class daObj_hsh_HIO_c {
-public:
-    daObj_hsh_HIO_c();
+class daObj_hsh_HIO_c { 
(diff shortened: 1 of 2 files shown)


AH64-dll:
The ctor match is solid — and the added virtual dtor is correct (the original has both __vt__ and the weak __dt__ for the HIO class). Two small cleanups:

1. A stray `/* Nonmatching */` is floating between `daObj_hsh_HIO_c::~daObj_hsh_HIO_c() {}` and the next function's address comment (d_a_obj_hsehi1.cpp) — it's not inside any function, a leftover from the ctor marker removal. Should be deleted (the daObj_hsh_c dtor below has its own proper marker).
2. Nit: the empty dtor is defined in the cpp; the merged-actor convention is inline in the header (`virtual ~daBdk_HIO_c() {}`, `daBeam_HIO_c`, etc.). Codegen-equal either way, just convention.

Neither affects the match.

PR #1036, opened 2026-05-29
d_a_ph - 99% matching

123456789758:


AH64-dll:
One coordination heads-up: this PR and #1032 (d_a_oq) both make the identical d_a_coming2.h change — completing the pluss_coming_point1() stub with the same event-reg increment. Same author, same change in two PRs, so whichever merges second will conflict (or apply trivially if git dedups). Worth deciding merge order between you so the rebase is painless.

AH64-dll:
heads-up: the original TU shows the mwcc GC/1.3.2r constant-pooling signature — 16 per-constant `lis`/`lfs` pairs in `anm_init` (vs project default 1.3.2 which pools constants under one base register). Verified case: d_a_mozo only snapped to 100% after compiling with 1.3.2r. If `anm_init` is stuck at 0.41%, try compiling the TU with `build/compilers/GC/1.3.2r/mwcceppc.exe` and objdiff-swap the .o; if it jumps to 100%, the flip needs the `mw_version` override on the ActorRel (`ActorRel(Matching, "d_a_d_a_ph", mw_version="GC/1.3.2r")` — passthrough is on main).

PR #1038, opened 2026-05-30
d_a_pt - 100% matching

123456789758:


AH64-dll:
Solid d_a_pt — full match, CI green x4. Same question as d_a_bo: `pt_class` moved from inheritance to composition (`fopEn_enemy_c actor` member). Layout-equivalent and CI-proven, but the repo convention is inheritance — does inheritance also match?

PR #1043, opened 2026-05-31
Decompile d_a_obj_msdan 99% match

riota935:
## Summary

Decompiles `d_a_obj_msdan`.

This adds the actor parameter helpers, static archive/event names, member layout, and implementations for:
- `daObjMsdan::Act_c::Mthd_Create`
- `daObjMsdan::Act_c::Mthd_Execute`
- `daObjMsdan::Act_c::Mthd_Delete`

## Match Status

Current objdiff status for `d_a_obj_msdan/d/actor/d_a_obj_msdan`:

- Overall fuzzy match: `99.155846%`
- Data match: `100.0%`
- Matched functions: `7 / 9`

Remaining nonmatching functions:

| Function | Match |
| --- | ---: |
| `daObjMsdan::Act_c::Mthd_Create()` | `98.953%` |
| `daObjMsdan::Act_c::Mthd_Execute()` | `99.187%` |

## Notes

`Mthd_Execute` appears to be off by a very small control-flow/codegen difference: the target has an extra redundant branch in the initial mode dispatch. Several C-only rewrites were tested, but they either preserved the current 99.187% result or made the function worse.

`Mthd_Create` is close but still has register allocation / ordering differences around the creation loops(regswap issue).

I am opening this as-is so someone with more MWCC codegen experience can help finish the last two nonmatches.

Any feedback is welcome :D

### DIFF
--- a/include/d/actor/d_a_obj_msdan.h
+++ b/include/d/actor/d_a_obj_msdan.h
@@ -1,22 +1,43 @@
 #ifndef D_A_OBJ_MSDAN_H
 #define D_A_OBJ_MSDAN_H
 
+#include "d/d_a_obj.h"
 #include "f_op/f_op_actor.h"
 
 namespace daObjMsdan {
     class Act_c : public fopAc_ac_c {
     public:
-        void prm_get_evId() const {}
-        void prm_get_size() const {}
-        void prm_get_sound() const {}
-        void prm_get_swSave() const {}
+        enum Prm_e {
+            PRM_SIZE_W = 0x1,
+            PRM_SIZE_S = 0x10,
+
+            PRM_SOUND_W = 0x1,
+            PRM_SOUND_S = 0x12,
+
+            PRM_SWSAVE_W = 0x8,
+            PRM_SWSAVE_S = 0x0,
+
+            PRM_EVID_W = 0x8,
+            PRM_EVID_S = 0x18,
+        };
+
+        u8 prm_get_evId() const { return daObj::PrmAbstract<Prm_e>(this, PRM_EVID_W, PRM_EVID_S); }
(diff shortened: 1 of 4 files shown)


riota935:
I added msdan2 here because I think it’s more appropriate to handle this together.

999sian:
I finished the two remaining `d_a_obj_msdan` functions independently and pushed the whole-object patch here: https://github.com/999sian/tww/commit/29e53f25be3266769c37a0623c03082fa4ae4435

The patch source-shapes `Mthd_Create`/`Mthd_Execute` for MWCC and uses `dComIfG_resDeleteDemo` so the same source also matches the demo build.

Fresh local verification on the pushed commit:

- D44J01: 9/9 functions, 1556/1556 code, 122/122 data; REL `e6799969705eea3d60dc10d15ccf611a7d2d949b`
- GZLP01: 9/9 functions, 1540/1540 code, 122/122 data; REL `536a3af9cc157fd6b3f43b5041f7d0c131b0eaf8`

Both hashes equal `config/<version>/build.sha1`. Please feel free to cherry-pick or reuse the matching source.

AH64-dll:
Thanks for the thorough writeup — the match-status table and notes are exactly the right way to document a partial. Two informational notes:

1. Heads-up: PR #1154 (d_a_obj_msdan 100% Matching, opened today) now covers all 9 functions including the two remaining here (Mthd_Create and Mthd_Execute, both at 100% there), with CI green on all four builds. Worth coordinating — you may want to close this or align with that PR (it also supersedes #1059 and #926 for the same actor).
2. CI here is also stale (builds ran 2026-06-05), so it'd need a rebase anyway.

Your partial work is solid — just flagging the overlap.

PR #1048, opened 2026-06-01
d_a_bo - 100% matching

123456789758:


AH64-dll:
Great work on d_a_bo — full match with green CI on all four builds. One question: the class was changed from `class bo_class : public fopEn_enemy_c` to a struct with a `fopEn_enemy_c actor` member (composition at offset 0). It's layout-equivalent and CI proves codegen, but every other matched enemy actor uses inheritance — would inheritance also match here? If yes, keeping the surrounding style would be preferred; if composition is required for the match, a brief note explaining why would help.

AH64-dll:
heads-up: the original TU shows the mwcc GC/1.3.2r constant-pooling signature — 8 per-constant `lis`/`lfs` pairs in `anm_init` (vs project default 1.3.2 which pools constants under one base register). Verified case: d_a_mozo only snapped to 100% after compiling with 1.3.2r. If `anm_init` is stuck at 0.78%, try compiling the TU with `build/compilers/GC/1.3.2r/mwcceppc.exe` and objdiff-swap the .o; if it jumps to 100%, the flip needs the `mw_version` override on the ActorRel (`ActorRel(Matching, "d_a_d_a_bo", mw_version="GC/1.3.2r")` — passthrough is on main).

PR #1050, opened 2026-06-02
d_a_bl - 100% matching

123456789758:


AH64-dll:
Nice work — the flip to Matching with all `/* Nonmatching */` markers removed and green CI on all four builds checks out. One heads-up for coordination: this PR's d_snap.h change (`DSNAP_TYPE_UNKB4` → `DSNAP_TYPE_BL`, offset 0xB4) is byte-identical to the same rename in #1145 (WilliamArnett's dSnap naming sweep). The names agree, so no conflict either way — just flagging so whichever merges second can drop the redundant hunk. Also noting the c_sxyz.h non-MWERKS fallback follows the existing c_xyz.h pattern (same guard shape), so no concerns there.

PR #1051, opened 2026-06-02
d_a_npc_ds1 99%

### LINKED ISSUE #329: d_a_npc_ds1
Ds1

Teotia444:
Missing 1 match for retail, 2 for demo
Closes #329

### DIFF
--- a/include/d/actor/d_a_npc_ds1.h
+++ b/include/d/actor/d_a_npc_ds1.h
@@ -1,90 +1,163 @@
 #ifndef D_A_NPC_DS1_H
 #define D_A_NPC_DS1_H
 
+#include "JSystem/JParticle/JPAEmitter.h"
+#include "d/d_bg_s_acch.h"
+#include "d/d_cc_d.h"
+#include "d/d_lib.h"
+#include "d/d_npc.h"
 #include "f_op/f_op_actor.h"
 #include "d/d_shop.h"
 
 class daNpc_Ds1_c : public fopAc_ac_c {
 public:
-    void checkAction(int (daNpc_Ds1_c::*)(void*)) {}
-    void getBackboneJntNum() {}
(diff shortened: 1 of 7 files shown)


AH64-dll:
Nice work on ds1. A few things when you get a chance:
1. The green CI is ~10 weeks old — a rebase onto current main + fresh run would confirm everything still holds (the tree has moved a lot).
2. Heads-up on review scope: this PR touches d_shop.h/d_shop.cpp plus d_a_npc_bs1.cpp and d_a_npc_rsh1.cpp alongside ds1. The d_shop.h changes look safe — the `m54` → `mSelectIdx` member rename is applied consistently everywhere (header, d_shop.cpp, callers) and is codegen-neutral, and the `SetSelectIdx` inline goes from empty stub to real body with no other in-tree callers — but the stub→real inline on a shared DOL header is exactly the change class that needs a full-build regression check before landing, so the fresh CI run matters.
3. The `/* Nonmatching (demo) */` marker with no configure.py flip is an honest partial — when the demo difference is resolved, remember the flip is a separate step.

AH64-dll:
heads-up: the original TU shows the mwcc GC/1.3.2r constant-pooling signature — 12 per-constant `lis`/`lfs` pairs in `setAnmFromMsgTag` (vs project default 1.3.2 which pools constants under one base register). Verified case: d_a_mozo only snapped to 100% after compiling with 1.3.2r. If `setAnmFromMsgTag` is stuck at 0.57%, try compiling the TU with `build/compilers/GC/1.3.2r/mwcceppc.exe` and objdiff-swap the .o; if it jumps to 100%, the flip needs the `mw_version` override on the ActorRel (`ActorRel(Matching, "d_a_d_a_npc_ds1", mw_version="GC/1.3.2r")` — passthrough is on main).

PR #1053, opened 2026-06-06
d_a_bpw - More cleanup + improvements

123456789758:


AH64-dll:
Hi — this PR currently has a merge conflict; could you rebase onto current main? After the rebase a fresh CI run (the current green checks are from ~2 weeks ago) would confirm the cleanup still builds cleanly.

AH64-dll:
Nice — this is exactly the right fix. The `fopAcM_monsSeStart` param_2 threading is something we tried on our side and had to revert because the am/am2 callsites (which pass non-zero param_2) weren't updated; you've done both halves, and the green all-4-build CI is the regression proof. The bpw enum renames and the `m67E/m680/m682 → csXyz m67E` layout fix look codegen-neutral. If you want, splitting the shared-header fix (which is merge-ready on its own) from the bpw cleanup would let the header fix land first.

PR #1054, opened 2026-06-10
d_a_warpls 100% matching

LegendofLuke03:


AH64-dll:
Nice work — full Matching across all four builds, named SE constants and resource enums throughout. A few small non-blocking polish items:

1. `eventInfo.mCondition |= 2` (x2: eventOrder) — the value 2 is dEvtCnd_UNK2_e (f_op_actor.h:146); using the named constant keeps it greppable and consistent with the dEvtCnd_* usage elsewhere in the file.
2. Missing spaces after commas in a few calls (`fopAcM_seStart(this,JA_SE_LK_WAPR_EFF_WARP,0)`, `dComIfG_getObjectRes(m_arcname[mWarpType],m_bdlidx[mWarpType])`).
3. Static members m_arcname/m_bdlidx/m_brkidx/m_bckidx/m_heapsize use the m_ prefix (reserved for instance members) — M_arcname or similar is the static convention.
4. g_profile_WARPLIGHT comment labels were reformatted (`/* LayerID */` vs the standard `/* Layer ID */`) — restoring the original labels avoids churn.
5. A few trailing-whitespace lines.

None of these block the merge.

src/d/actor/d_a_warpls.cpp
```diff
@@ -5,75 +5,405 @@
...
-void daWarpls_c::check_warp_link() {
-    /* Nonmatching */
+int daWarpls_c::check_warp_link() {
+
+    #if VERSION == VERSION_DEMO
+    f32 warp_distance;
+    #else
+    f32 warp_distance = 112.5f;
+    #endif
+    f64 distance_to_player = 0.5f;
```
LagoLunatic: Why does this line exist? You overwrite the value later so I don't see where 0.5 came from, is it a Ghidra/AI hallucination? Should also be f32 instead of f64 I think.

src/d/actor/d_a_warpls.cpp
```diff
@@ -5,75 +5,405 @@
...
+    fopAc_ac_c* player = dComIfGp_getLinkPlayer();
+    if (!(player == dComIfGp_getPlayer(0) && mWarpActive && !mPlayerStartedInWarp)) {
+        return FALSE;
+    }
+
+    distance_to_player = (player->current.pos - current.pos).absXZ();
+
+    #if VERSION == VERSION_DEMO
+    warp_distance = 112.5f;
+    #endif
```
LagoLunatic: Why did you load this 112.5 variable in two different places depending on the version? It's loaded in the same place in the asm in both versions, so this looks like a fake version difference.

Also, this might not even be a local var at all, did you try using the `daWarpls_c::m_warp_distance` constant directly instead and getting rid of `warp_distance`?

src/d/actor/d_a_warpls.cpp
```diff
@@ -5,75 +5,405 @@
...
-    /* Nonmatching */
+bool daWarpls_c::demo() {
+
+    #if VERSION == VERSION_DEMO
+    int switch_number = mSwitchNo;
+    char room_no = home.roomNo;
+    u8 is_switch_on = dComIfGs_isSwitch(switch_number,room_no);
+    #else
+    u8 is_switch_on = dComIfGs_isSwitch(mSwitchNo, home.roomNo);
+    #endif
```
LagoLunatic: Fake version difference, this is the fopAcM_isSwitch inline mentioned in the debug maps.

src/d/actor/d_a_warpls.cpp
```diff
@@ -5,75 +5,405 @@
...
+    fopAcM_ct(this, daWarpls_c);
+    mWarpType = daWarpls_prm::getType(this);
+    rt = dComIfG_resLoad(&mPhs, m_arcname[mWarpType]);
+
+    if (rt == cPhs_COMPLEATE_e) {
+        if (!fopAcM_entrySolidHeap(this, CheckCreateHeap, m_heapsize[mWarpType])) {
+            return cPhs_ERROR_e;
+        }
+    CreateInit();
+    }
```
LagoLunatic, line 165: Wrong indent

src/d/actor/d_a_warpls.cpp
```diff
@@ -5,75 +5,405 @@
...
+
+    if (mpEmitter != NULL) {
+        mpEmitter->becomeInvalidEmitter();
+        mpEmitter = NULL;
+    }
+    #if VERSION == VERSION_DEMO
+    dComIfG_deleteObjectRes(m_arcname[mWarpType]);
+    #else
+    dComIfG_resDelete(&mPhs, m_arcname[mWarpType]);
+    #endif
```
LagoLunatic: Can use this macro to simplify it a bit
```suggestion
    dComIfG_resDeleteDemo(&mPhs, m_arcname[mWarpType]);
```

src/d/actor/d_a_warpls.cpp
```diff
@@ -5,75 +5,405 @@
...
+    return ((daWarpls_c*)actor)->CreateHeap();
 }
 
 /* 00000100-000003D8       .text CreateHeap__10daWarpls_cFv */
-void daWarpls_c::CreateHeap() {
-    /* Nonmatching */
+int daWarpls_c::CreateHeap() {
+
+    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes(m_arcname[mWarpType],m_bdlidx[mWarpType]);
+    JUT_ASSERT(VERSION_SELECT(0xe6, 0xe9, 0xe9, 0xe9), modelData != NULL);
```
LagoLunatic: Can use DEMO_SELECT instead of VERSION_SELECT if only demo is different from the others. Also line numberrs should ideally be decimal
```suggestion
    JUT_ASSERT(DEMO_SELECT(230, 233), modelData != NULL);
```

src/d/actor/d_a_warpls.cpp
```diff
@@ -5,75 +5,405 @@
...
+            mEvtState = EvtState_None;
+            dComIfGp_evmng_setGoal(&current.pos);
+        }
+        if (dComIfGp_evmng_endCheck(mWarpingEvtIdx)) {
+            #if VERSION == VERSION_DEMO
+            s8 room_no = current.roomNo;
+            dLib_setNextStageBySclsNum(mSceneNo, room_no);
+            #else
+            dLib_setNextStageBySclsNum(mSceneNo, current.roomNo);
+            #endif
```
LagoLunatic: Fake version diff, `fopAcM_GetRoomNo` inline

src/d/actor/d_a_warpls.cpp
LegendofLuke03: ah yeah not sure why that was 0.5 must've been from Ghidra when i started on this?  I guess I wasn't sure what that var was before so the value never raised my suspicion before the name change.

src/d/actor/d_a_warpls.cpp
LegendofLuke03: yeah if i remember correctly when going through a map file the variable was local scoped so i had thought i needed to implement it in a certain way. Using daWarpls_c::m_warp_distance works without version diff

PR #1055, opened 2026-06-15
d_a_obj_mmrr: 100% matching decompilation of mirror actor

LostGoron:
Decompilation of `d_a_obj_mmrr` — the mirror actor that reflects the light beam.

## Match status
All functions 100% matching (GZLE01, GZLJ01, GZLP01).

The only non-100% entries in objdiff are `CrossAtTg` and the base-only `.text (654)` section, both caused by weak-symbol destructor ordering — a known artifact, not a real mismatch.

## Notable technical point
The four `dComIfG_getObjectRes` calls in `create_heap` require `static_cast<T*>` instead of a C-style cast. A C-style cast produces a different register allocation (r28/r29/r30 rotation on the 3-value conflict clique). This matches the pattern documented in `docs/regalloc.md`.

## Notes
- I don't yet have access to the Twilight Princess Ghidra database, so no cross-reference was made with it.
- This is my first contribution to the project.

### DIFF
--- a/include/d/actor/d_a_obj_mmrr.h
+++ b/include/d/actor/d_a_obj_mmrr.h
@@ -2,43 +2,67 @@
 #define D_A_OBJ_MMRR_H
 
 #include "f_op/f_op_actor.h"
+#include "d/d_cc_d.h"
+#include "d/d_particle.h"
+#include "m_Do/m_Do_ext.h"
+
+class J3DModel;
 
 namespace daObjMmrr {
-    class Act_c : public fopAc_ac_c {
-    public:
-        void setup(const cXyz* pos) {
-            current.pos = *pos;
-            mBF8 = 1;
-        }
-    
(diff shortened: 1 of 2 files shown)


AH64-dll:
Excellent writeup — the regalloc note matching docs/regalloc.md is exactly the right kind of documentation. Two currency items before merge:

1. No configure.py flip: the PR touches only the header and cpp, so d_a_obj_mmrr stays NonMatching in configure.py despite 'All functions 100% (GZLE01, GZLJ01, GZLP01)' — flipping to MatchingFor("GZLJ01", "GZLE01", "GZLP01") would let CI verify it (the demo build presumably differs per the CrossAtTg/base .text note).
2. No CI: 'no checks reported' on the branch — pushing to CI is what confirms the flip. Right now the claim is unverified.

The matching itself reads clean (named enums, honest artifact notes).

PR #1056, opened 2026-06-16
d_a_salvage_tbox 40%

samkleespies:
Partial decomp of `d_a_salvage_tbox` (29/36 functions matching).


AH64-dll:
Clean partial on salvage_tbox (29/36) — no flip is honest. CI is from 06-17; a rebase + fresh run when you next touch it would confirm it still holds.

PR #1057, opened 2026-06-17
d_a_fallrock_tag - 100% matching for GZLE01

samkleespies:
Matches d_a_fallrock_tag for GZLE01.\n\nVerified:\n- ninja\n- ninja changes_all\n\nOther regional extracted inputs are unavailable in this local workspace, so configure.py only marks GZLE01 matching.


AH64-dll:
Clean fallrock_tag, and the honest MatchingFor(GZLE01) flip is appreciated. Could you rebase and re-run CI? The green checks are from 06-17.

PR #1063, opened 2026-06-23
d_a_obj_light

### LINKED ISSUE #419: d_a_obj_light
Stoudai

MaskeZenY:
This PR decompiles `d_a_obj_light.cpp` (100% matching). 

It is my very first contribution to the decompilation project and my first PR on GitHub, so please let me know if there is anything I should adjust.

Note: I tested this on the USA version (`GZLE01`), so I am not entirely sure if any specific changes are needed for the other regions. 

Closes #419

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1732,7 +1732,7 @@ config.libs = [
     ActorRel(Matching,    "d_a_obj_jump"),
     ActorRel(NonMatching, "d_a_obj_kanoke"),
     ActorRel(Matching,    "d_a_obj_ladder"),
-    ActorRel(NonMatching, "d_a_obj_light"),
+    ActorRel(MatchingFor("GZLE01"), "d_a_obj_light"),
     ActorRel(Matching,    "d_a_obj_mkie"),
     ActorRel(Matching,    "d_a_obj_mkiek"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_obj_mknjd"),
--- a/include/d/actor/d_a_obj_light.h
+++ b/include/d/actor/d_a_obj_light.h
@@ -2,22 +2,26 @@
 #define D_A_OBJ_LIGHT_H
 
 #include "f_op/f_op_actor.h"
+#include "d/d_cc_d.h"
+#include "d/d_bg_w.h"
+#include "d/d_particle.h"
+#include "m_Do/m_Do_ext.h"
 
 namespace daObjLight {
     class Act_c : public fopAc_ac_c {
     public:
-        void solidHeapCB(fopAc_ac_c*);
-        void create_heap();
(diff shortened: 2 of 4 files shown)


Kelebek1:
You need to change this line in configure.py for your decomp to actually take effect: https://github.com/zeldaret/tww/blob/main/configure.py#L1735

Change `NonMatching` into `MatchingFor("GZLE01")` and then build to see if it's actually OK. When the file is marked as NonMatching it doesn't use your decompiled code so you don't know if it really matches until you change this.

MaskeZenY:
thanks, I’ll try that.

btw quick question, what is the .data-0 section showing like 42% in the build? what does it actually represent?

{09C5645E-3D4B-4406-AC93-8B2A6E6CBDB6}
{583620A4-544F-4BFB-828B-EAE47FB73484}

Kelebek1:
Not really sure what to say, it's a measure of the difference between the .data sections in the original and yours. Same sort of comparison as functions, comparing bytes and sizes and however it does it. There's a bunch of data from included files that ends up being removed in compilation, so it won't show 100% there, but still matches when you compile because that stuff doesn't end up in the final binary.

MaskeZenY:
ok thanks for the answer
so basically the bot decomp-dev is already showing 100% for all versions now, but do you think we should also add the other versions into MatchingFor like GZLP01 / GZLJ01 etc, or is it fine to only keep GZLE01 for now since it’s the one I tested?

AH64-dll:
Thanks for this — great first contribution! Since #1130 also decompiles this actor, the maintainers will likely pick one to move forward. If this one is kept, a few cleanups before merge:

1. Use the resource enums instead of raw indices: `dRes_INDEX_SKANRAN_BDL_STOUDAI_e` / `_BDL_SHIKARI_e` / `_DZB_STOUDAI_e` (assets/GZLE01/res/Object/Skanran.h) instead of 7/5/12.
2. Replace raw `seStart` ids with named constants: `0x6A03` = `JA_SE_OBJ_L_HOUSE_FIRE`, `0x6A04` = `JA_SE_OBJ_L_HOUSE_LIGHT`, `0x303B` = `JA_SE_OBJ_L_HOUSE_ROUND`.
3. Remove the hand-written padding fields (`/* 0x465 */ u8 mPadding[3];`, `/* 0x4a2 */ u8 mPadding3[2];`) — we avoid explicit padding.
4. Use the named enums instead of magic numbers: `dKy_TIME_NIGHT_e` for `== 1`, `cBgW::MOVE_BG_e` for `Set(..., 1, ...)`, `dDlst_alphaModel_c::TYPE_TWO_SPHERES` for `dComIfGd_setAlphaModel(1, ...)`.
5. The static `L_attr[3]` table is never referenced — remove it or use it. (Note it also disagrees in byte interpretation with #1130's equivalent table, so whichever matches the original data should be verified.)

PR #1068, opened 2026-06-27
d_a_warpmj 100% matching

### LINKED ISSUE #484: d_a_warpmj
Warpmj

GustaveHerve:
Hello everyone, this is my first contribution to this project :)

This PR 100% matches `d_a_warpmj` in objdiff for every versions
I tried my best to name fields correctly according to similar matched TUs, let me know if anything seems wrong, I am still learning the decomp process ahah

EDIT: Japanese and Demo versions now also match 100%

Closes #484

### DIFF
--- a/src/d/actor/d_a_warpmj.cpp
+++ b/src/d/actor/d_a_warpmj.cpp
@@ -92,7 +92,9 @@ void daWarpmj_c::CreateInit() {
     setEndAnm();
     mEvtToGanonWarpIdx = dComIfGp_evmng_getEventIdx("TO_GANON_WARP");
     mSceneNo = daWarpmj_prm::getSceneNo(this);
+#if VERSION > VERSION_DEMO
     dKy_tevstr_init(&mTevStr, fopAcM_GetRoomNo(this), 0xff);
+#endif
 }
 
 /* 00000630-00000778       .text _create__10daWarpmj_cFv */
@@ -120,9 +122,11 @@ void daWarpmj_c::set_mtx() {
     cXyz local_28(current.pos);
     local_28.y += 2000.0f;
     current.pos.y = getSeaY(local_28);
+#if VERSION > VERSION_DEMO
     if (current.pos.y == -1e9f) {
         current.pos.y = 0;
     }
+#endif
     mpModel->setBaseScale(scale);
     mDoMtx_stack_c::transS(current.pos);
     mpModel->setBaseTRMtx(mDoMtx_stack_c::get());
@@ -147,7 +151,7 @@ bool daWarpmj_c::_execute() {
 void daWarpmj_c::normal_execute() {
     /* Nonmatching */
     animPlay();
-    if (check_warp() != 0) {
+    if (check_warp()) {
         m2B8 = TRUE;
     }
 }
@@ -155,9 +159,9 @@ void daWarpmj_c::normal_execute() {
 /* 000009D4-00000A60       .text demo_execute__10daWarpmj_cFv */
 void daWarpmj_c::demo_execute() {
     /* Nonmatching */
-    dDemo_actor_c* pdVar1 = dComIfGp_demo_getActor(demoActorID);
-    if (pdVar1 != NULL) {
-        m2C0 = pdVar1->getShapeId();
+    dDemo_actor_c* demo_actor = dComIfGp_demo_getActor(demoActorID);
+    if (demo_actor != NULL) {
+        m2C0 = demo_actor->getShapeId();
         if (m2C0 == 0) {
             fopAcM_offDraw(this);
         } else if (m2C0 == 1) {
@@ -177,15 +181,15 @@ void daWarpmj_c::demo_proc() {
     };
     mStaffIdx = dComIfGp_evmng_getMyStaffId("Warpmj");
     if (dComIfGp_event_runCheck() && !eventInfo.checkCommandTalk() && mStaffIdx != -1) {
-        int actIdx = dComIfGp_evmng_getMyActIdx(mStaffIdx, action_table, 4, FALSE, 0);
-        if (actIdx == -1) {
+        int act_idx = dComIfGp_evmng_getMyActIdx(mStaffIdx, action_table, 4, FALSE, 0);
+        if (act_idx == -1) {
             dComIfGp_evmng_cutEnd(mStaffIdx);
         } else {
             if (dComIfGp_evmng_getIsAddvance(mStaffIdx)) {
-                (this->*event_init_tbl[actIdx])(mStaffIdx);
+                (this->*event_init_tbl[act_idx])(mStaffIdx);
             }
 
-            BOOL ret = (this->*event_action_tbl[actIdx])(mStaffIdx);
+            BOOL ret = (this->*event_action_tbl[act_idx])(mStaffIdx);
             if (ret) {
                 dComIfGp_evmng_cutEnd(mStaffIdx);
             }
@@ -218,7 +222,11 @@ int daWarpmj_c::actWarp(int) {
 /* 00000C38-00000C94       .text initWarpArrive__10daWarpmj_cFi */
 void daWarpmj_c::initWarpArrive(int) {
     setEndAnm();
-    mDoAud_seStart(0x2894);
+#if VERSION >= VERSION_USA
+    mDoAud_seStart(JA_SE_LK_GN_WAPR_U_OUT);
+#else
+    mDoAud_seStart(JA_SE_LK_GN_WAPR_D_OUT);
+#endif
 }
 
 /* 00000C94-00000CB8       .text actWarpArrive__10daWarpmj_cFi */
@@ -278,9 +286,11 @@ f32 daWarpmj_c::getSeaY(cXyz i_pos) {
 
 /* 00000EE0-00000FDC       .text check_warp__10daWarpmj_cFv */
 BOOL daWarpmj_c::check_warp() {
-    daShip_c* pShip = dComIfGp_getShipActor();
-    if (dComIfGp_checkPlayerStatus0(0, daPyStts0_SHIP_RIDE_e) != 0 && (pShip != NULL)) {
-        if ((f32)(pShip->current.pos - current.pos).absXZ() < 200.0f) {
+    daShip_c* ship = dComIfGp_getShipActor();
+    f32 max = 200.0f;
+    if (dComIfGp_checkPlayerStatus0(0, daPyStts0_SHIP_RIDE_e) && ship != NULL) {
+        f32 abs = (ship->current.pos - current.pos).absXZ();
+        if (abs < max) {
             return TRUE;
         }
     }
--- a/src/d/actor/d_a_warpmj.cpp
+++ b/src/d/actor/d_a_warpmj.cpp
@@ -5,120 +5,315 @@
 
 #include "d/dolzel_rel.h" // IWYU pragma: keep
 #include "d/actor/d_a_warpmj.h"
+#include "d/actor/d_a_sea.h"
+#include "d/d_bg_s_func.h"
+#include "d/d_lib.h"
+#include "d/actor/d_a_ship.h"
+#include "res/Object/Gmjwp.h"
+
+const char daWarpmj_c::m_arcname[] = "Gmjwp";
+const u32 daWarpmj_c::m_heapsize = 0x3000;
+const f32 daWarpmj_c::m_warp_distance = 200.0f;
+
+typedef void (daWarpmj_c::*EventInitFunc)(int);
+EventInitFunc event_init_tbl[] = {
+    &daWarpmj_c::initWait,
+    &daWarpmj_c::initWarp,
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
+    }
+#endif
+    mpModel->setBaseScale(scale);
+    mDoMtx_stack_c::transS(current.pos);
+    mpModel->setBaseTRMtx(mDoMtx_stack_c::get());
 }
 
 /* 0000084C-00000990       .text _execute__10daWarpmj_cFv */
 bool daWarpmj_c::_execute() {
     /* Nonmatching */
```
WilliamArnett: Make sure to take out all of your `/* Nonmatching */` comments

src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
+int daWarpmj_c::actWarp(int) {
+    animPlay();
+    return 1;
 }
 
 /* 00000C38-00000C94       .text initWarpArrive__10daWarpmj_cFi */
 void daWarpmj_c::initWarpArrive(int) {
-    /* Nonmatching */
+    setEndAnm();
+#if VERSION >= VERSION_USA
```
WilliamArnett: This isn't wrong by any means, but I almost always see it formatted as
`#if VERSION > VERSION_JPN`

src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
 
 /* 00000D14-00000DD0       .text checkOrder__10daWarpmj_cFv */
 void daWarpmj_c::checkOrder() {
-    /* Nonmatching */
+    if (eventInfo.checkCommandDemoAccrpt()) {
+        if (dComIfGp_evmng_startCheck(mEvtToGanonWarpIdx) && m2B8) {
+            m2B8 = FALSE;
+        }
+        if (dComIfGp_evmng_endCheck(mEvtToGanonWarpIdx)) {
+            dLib_setNextStageBySclsNum(mSceneNo & 0xFF, fopAcM_GetRoomNo(this));
```
WilliamArnett: You don't need the `& 0xFF` in this function call, since `mSceneNo` already gets implicitly cast to `u8`.

src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
+        mpBrkAnm->setFrame(mpBrkAnm->getEndFrame());
+    }
 }
 
 /* 00000E8C-00000EE0       .text getSeaY__10daWarpmj_cF4cXyz */
-void daWarpmj_c::getSeaY(cXyz) {
-    /* Nonmatching */
+f32 daWarpmj_c::getSeaY(cXyz i_pos) {
+    if (daSea_ChkArea(i_pos.x, i_pos.z)) {
+        return daSea_calcWave(static_cast<f64>(i_pos.x), static_cast<f64>(i_pos.z));
```
WilliamArnett: You don't need the static casts here. The function signature takes `(f32, f32)`, so I would avoid this here.

src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
+    }
+    return dBgS_ObjGndChk_Wtr_Func(i_pos);
 }
 
 /* 00000EE0-00000FDC       .text check_warp__10daWarpmj_cFv */
-void daWarpmj_c::check_warp() {
-    /* Nonmatching */
+BOOL daWarpmj_c::check_warp() {
+    daShip_c* ship = dComIfGp_getShipActor();
```
WilliamArnett, line 286: The debug map seems to have a weak function call to `daPy_getPlayerActorClass()` in this function. I don't think it ended up getting used though. It may be worth noting in a comment, but I don't think you need to change anything

src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
+        m2B8 = TRUE;
+    }
 }
 
 /* 000009D4-00000A60       .text demo_execute__10daWarpmj_cFv */
 void daWarpmj_c::demo_execute() {
     /* Nonmatching */
+    dDemo_actor_c* demo_actor = dComIfGp_demo_getActor(demoActorID);
+    if (demo_actor != NULL) {
+        m2C0 = demo_actor->getShapeId();
```
WilliamArnett: If you want, you could call `m2C0` something like `mDemoShapeId`

src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
+    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -400.0f, 0.0f, -400.0f, 400.0f, 800.0f, 400.0f);
+    fopAcM_setCullSizeFar(this, 1.0f);
+    set_mtx();
+
+    mpEmitter1 = dComIfGp_particle_set(dPa_name::ID_IT_SN_MJWPA00, &current.pos);
+    mpEmitter2 = dComIfGp_particle_set(dPa_name::ID_IT_SN_MJWPB00, &current.pos);
+    mpEmitter3 = dComIfGp_particle_setProjection(dPa_name::ID_IT_SP_MJWPC00, &current.pos);
+
+    m2B8 = FALSE;
```
WilliamArnett, line 92: I think you could call `m2B8` something like `mbWarpInUse`. I need to test in-game to be 100% certain this describes the behavior properly.

src/d/actor/d_a_warpmj.cpp
```diff
@@ -5,120 +5,325 @@
...
 
 /* 0000049C-00000630       .text CreateInit__10daWarpmj_cFv */
 void daWarpmj_c::CreateInit() {
-    /* Nonmatching */
+    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -400.0f, 0.0f, -400.0f, 400.0f, 800.0f, 400.0f);
+    fopAcM_setCullSizeFar(this, 1.0f);
+    set_mtx();
+
+    mpEmitter1 = dComIfGp_particle_set(dPa_name::ID_IT_SN_MJWPA00, &current.pos);
```
WilliamArnett: If you want to be as descriptive as possible with these emitter variable names, you could call them `mpEmitterMajyuWarpA`,`mpEmitterMajyuWarpB`, and `mpEmitterMajyuWarpC`. This aligns with the enums you have, but might be a little too verbose. It's up to you.

WilliamArnett (commented):
Your code is very very clean, I'm a bit jealous.

LostGoron:
Hi,

I'm quite new in decompilation, but have some experience now in using Claude (combined with Gemini for critical review).
I'm developping an agent to speed up clean decompilation and PR reviews.

Here is the review of your PR : 


## Overview

This is an excellent contribution. The matching work across all four game versions (GZLJ01, GZLE01, D44J01, GZLP01) is confirmed correct, and the overall code structure is solid — the `daWarpmj_prm` namespace is properly placed in the header with its inline accessor, resource archive enums (`dRes_INDEX_GMJWP_*`) are used throughout, and the delegation pattern (`static_cast<daWarpmj_c*>` in `CheckCreateHeap`) is exactly right. The register allocation choices that achieve 100% on all versions are the hard part, and they're done.

There are a few style issues to fix before merge. None of them affect the matching.

---

## Issue 1 — Residual `/* Nonmatching */` in implemented functions

**Severity: Must fix**

Per `CLAUDE.md`, `/* Nonmatching */` is a stub marker that must be removed as soon as a function has a real body. Several functions received an implementation in this PR but the marker was not deleted — it sits as the first line of the body alongside working code:

```cpp
// Current (incorrect):
bool daWarpmj_c::_execute() {
    /* Nonmatching */        // ← remove this
    if (demoActorID == 0) {
        checkOrder();
        demo_proc();
        eventOrder();
    } else {
        demo_execute();
    }
    fopAcM_seStart(this, JA_SE_OBJ_GN_WAPR_EFF, 0);
    set_mtx();
    return TRUE;
}
```

---

## Issue 2 — Unknown member fields should use `field_0x` naming with lowercase hex

**Severity: Must fix**  

The coding guidelines specify that unknown/unidentified struct members use the `field_0x` prefix followed by the offset in **lowercase hexadecimal**. Three members in `d_a_warpmj.h` use an `m` prefix with uppercase hex instead:

```cpp
// Current (incorrect):
/* 0x2B8 */ BOOL m2B8;
/* 0x2BE */ u8 m2BE[0x2C0 - 0x2BE];
/* 0x2C0 */ u32 m2C0;

// Should be:
/* 0x2b8 */ BOOL field_0x2b8;
/* 0x2be */ u8 field_0x2be[0x2c0 - 0x2be];
/* 0x2c0 */ u32 field_0x2c0;
```

The offset comments in the `/* ... */` prefix should also use lowercase hex, consistent with the rest of the file (`/* 0x290 */`, `/* 0x298 */`, etc. are already lowercase — these three break the pattern).

Remember to rename the corresponding usages in `d_a_warpmj.cpp` as well (any place that reads or writes `m2B8`, `m2BE`, or `m2C0`).

---

## Issue 3 — `event_init_tbl` and `event_action_tbl` are missing `static`

**Severity: Should fix**  

The sibling actor `d_a_warpgn.cpp` (already merged, also 100% matching) declares its equivalent tables with `static`:

```cpp
// d_a_warpgn.cpp — reference pattern:
static EventInitFunc event_init_tbl[] = { ... };
static EventActionFunc event_action_tbl[] = { ... };
```

In this PR, `static` is absent:

```cpp
// d_a_warpmj.cpp — current (missing static):
EventInitFunc event_init_tbl[] = {
    &daWarpmj_c::initWait,
    &daWarpmj_c::initWarp,
    &daWarpmj_c::initWarpArrive,
};
EventActionFunc event_action_tbl[] = {
    &daWarpmj_c::actWait,
    &daWarpmj_c::actWarp,
    &daWarpmj_c::actWarpArrive,
};
```

Without `static`, these arrays have external linkage — they're visible as global symbols across the entire program. That's almost certainly not the original intent. Adding `static` gives them translation-unit scope, consistent with `d_a_warpgn` and with the project's `l_` prefix convention for file-scope statics. Since `static` doesn't change the generated code (only the symbol linkage), it should not affect matching.

---

## Issue 4 — Unnecessary `return;` at end of `animPlay`

**Severity: Minor**  

```cpp
void daWarpmj_c::animPlay() {
    if (mpBtkAnm != NULL) {
        mpBtkAnm->setPlaySpeed(1.0f);
        mpBtkAnm->play();
    }
    return;  // redundant for a void function
}
```

An explicit `return;` at the end of a `void` function is noise. MWCC emits the same `blr` either way. Safe to remove.

---

src/d/actor/d_a_warpmj.cpp
GustaveHerve: Good catch, this is me taking Ghidra's output a little too literally
I'll keep that in mind for next times

src/d/actor/d_a_warpmj.cpp
GustaveHerve: Indeed this inline call is missing. I tried to check again this evening but couldn't understand how to use it in this context (but again this is my first TU, so I am probably just missing something).
I'll try to look into it again in the next days, I added a comment explaining this in the meantime

WilliamArnett: Yeah that's exactly what I would put. Looks good.

GustaveHerve: So what's odd is that the assembly of  `daWarpdm20_c::check_warp` and `daWarpmj_c::check_warp` are almost identical, but the debug map file of the former doesn't mention `daPy_getPlayerActorClass()`
To be frank I have no idea what is going on here...

src/d/actor/d_a_warpmj.cpp
GustaveHerve: This name seems plausible to me but to be honest I am not knowledgeable enough to verify it. Perhaps I should let the `m2B8` name for now until further confirmation ?

WilliamArnett: I agree. The code does make it seem like its accurate, but I don't think its important enough to be given a name before you're sure it works that way.

src/d/actor/d_a_warpmj.cpp
WilliamArnett: Neither do I, its definitely beyond me. Putting a comment for it is definitely good enough for now.

AH64-dll:
Nice work on warpmj, especially the event table and demo guards. One cleanup: `check_warp` still carries a scratch note ('// This function matches as is, but lacks a call to the daPy_getPlayerActorClass() inline ... May be something to look into later'). If it's a known difference, it belongs in the PR description rather than the code. Also the green CI is from 06-30 — a rebase + fresh run would confirm it still matches current main.

PR #1070, opened 2026-06-28
d_a_npc_p2 100% Matching

### LINKED ISSUE #353: d_a_npc_p2
P2a, P2b, P2c

WilliamArnett:
100% Matching for demo and retail builds.

Closes #353


AH64-dll:
Impressive scale — 100% on a big NPC across all four builds. Two things before merge:

1. d_a_himo3.h getPlayerMoveLength() is a ~20-line comment wall: `//TODO: Used in daNpc_p2_c and daPy_lk_c. Implementation is something like this:` followed by the entire candidate implementation commented out. That's the exact pattern we avoid in final code (commented-out code + TODO + speculative body). If this stub is required for layout/matching, it needs a real body or a proper explanation; if not, it should revert to the stub. Either way the comment wall has to go before merge.
2. CI is stale — the four builds ran 2026-07-02 (44 days ago). The PR touches cross-actor files (d_com_inf_game.cpp, d_stage.cpp, d_save_event_flag.inc — the UNK_0808->ARRIVE_MAJYU renames are codegen-neutral and welcome) and the base has moved; a rebase + fresh CI push is needed to confirm the flip still holds.

Matching reads clean otherwise.

PR #1071, opened 2026-06-28
d_a_obj_ashut OK

N1ghtD3c0d3r:
/**
 * d_a_obj_ashut.cpp
 * Object - Metal bars (Pirate Ship rope minigame)
 */

Hi everyone, this is my second contribution.

I took an easy object again because I was not feeling ready for the big ones yet. After those two actors I feel I have more context, maybe I gonna try to work on bigger objects for my next PRs.

I got access to the demo version of the game, so in this PR I was able to make it match with it as well.

For this PR, I struggled a bit with weak funcion order. I tried moving multiple things around but it did not work. My only attempt that worked was changing a casting from from "daObjAshut::Act_c*" to "dBgS_MoveBgActor*". I thought it would be ok because that function is virtual, please let me know if it was a bad decision.

All feedback is very welcome.

Thank you very much ^^

closes https://github.com/zeldaret/tww/pull/1071


src/d/actor/d_a_obj_ashut.cpp
```diff
@@ -125,7 +327,9 @@ BOOL Mthd_Execute(void* i_this) {
 
 /* 00000DCC-00000DF8       .text Mthd_Draw__Q210daObjAshut27@unnamed@d_a_obj_ashut_cpp@FPv */
 BOOL Mthd_Draw(void* i_this) {
-    return ((daObjAshut::Act_c*)i_this)->Draw();
+    return
+        static_cast<dBgS_MoveBgActor*> /* casting as dBgS_MoveBgActor* solves weak func order*/
```
N1ghtD3c0d3r, line 331: I struggled a bit with weak funcion order. I tried moving multiple things around but it did not work. My only attempt that worked was changing a casting from from "daObjAshut::Act_c*" to "dBgS_MoveBgActor*". I thought it would be ok because that function is virtual, please let me know if it was a bad decision.

LostGoron:
Hi,

I'm quite new in decompilation, but have some experience now in using Claude (combined with Gemini for critical review).
I'm developping an agent to speed up clean decompilation and PR reviews.

Here is the review of your PR :


# PR Review: d_a_obj_ashut (#1071)

## Issues to fix before merge

### Important — `L_attr` should use `l_` prefix (lowercase L)

File-scope statics (including inside anonymous namespaces) use the `l_` prefix convention.
`L_attr` uses an uppercase `L`.

Current code:
```cpp
const Attr_c L_attr = { ... };
inline static const Attr_c& attr() { return L_attr; }
```

Suggested fix (check `frameworkD.map` first; if not listed there):
```cpp
const Attr_c l_attr = { ... };
inline const Attr_c& attr() { return l_attr; }
```
(`static` is also redundant on a function already inside an anonymous namespace.)

---

### Minor — File-scope constants at global scope

`ANTI_CRUSH_SAFE_AREA_X/Y/Z` and `BGW_RELEASE_HEIGHT` are declared at **global** scope,
outside any namespace, and don't follow the `l_` convention for file-scope data.

Current code:
```cpp
const f32 ANTI_CRUSH_SAFE_AREA_Y = 100.0f;
const f32 ANTI_CRUSH_SAFE_AREA_Z = 55.0f;
const f32 ANTI_CRUSH_SAFE_AREA_X = 105.0f;
const f32 BGW_RELEASE_HEIGHT = 150.0f;
```

Suggested: move them inside the existing anonymous namespace and apply `l_` naming (or fold
them into `Attr_c` alongside the other tuned values). Verify the map for the real symbol names.

---

### Minor — `bool` mixed with `TRUE` in `mode_demoreq()`

`is_demo_ready` is declared `bool` and initialized with `false`, but assigned the `BOOL` macro
`TRUE` in one branch. Use the matching literal.

Current code:
```cpp
bool is_demo_ready = false;
...
} else {
    is_demo_ready = TRUE;
}
```

Suggested fix:
```cpp
bool is_demo_ready = false;
...
} else {
    is_demo_ready = true;
}
```

---

### Trivial — Offset comment casing

The member offset comments mix cases: `/* 0x2c8 */` (lowercase) sits next to `/* 0x2D0 */`,
`/* 0x2D4 */`, etc. (uppercase). Align them all to the codebase standard (uppercase).

---

### Trivial — Redundant comparison in `mode_demoreq_init()`

`if (mIsEventReady != false)` on a `bool` can simply be `if (mIsEventReady)`.

---

N1ghtD3c0d3r:
Hi @LostGoron tu for the review.

I cannot change L_attr to l_attr, according the map file, the correct value is L_attr.

I adjusted the cont values to follow the pattern with prefix "l_".

N1ghtD3c0d3r:
Thank you @Teotia444 I saw your comment in Discord, I've added some missing enums and the missing inline near the mtx stuff

AH64-dll:
Solid ashut — the weak-func-order cast comment is exactly right. Could you rebase and re-run CI? The green checks are from 07-08.

PR #1085, opened 2026-06-30
d_a_npc_bs1

samkleespies:
Marks the retail versions of d_a_npc_bs1 equivalent while keeping D44J01 matching. The remaining retail difference is a harmless regalloc mismatch in next_msgStatus.

Checked with:
- ninja
- ninja changes_all


samkleespies:
Closing this since the report shows no progress change.

PR #1091, opened 2026-07-01
d_a_obj_ikada: match setWave

samkleespies:
Matches `daObj_Ikada_c::setWave`. The wave-collapse anchor vectors for the left callback are now built as copies of the right anchors and negated with a plain `x *= -1.0f`, matching the idiom used in the already-matching `d_a_oship`/`d_a_ship` `setAnchor` paths.

The unit still has other non-matching functions, so it remains `NonMatching` in configure.py.


AH64-dll:
Heads-up: PR #1152 (d_a_obj_ikada 100% Matching) includes this same setWave fix (byte-identical) plus the configure.py flip to Matching for the whole TU, verified against the pinned rel hash. If this PR's setWave change merges first, #1152 will become a flip-only PR; if it stalls, #1152 supersedes it. Happy to coordinate either way.

PR #1092, opened 2026-07-01
d_a_obj_msdan2 - 100% matching for GZLJ01, GZLE01 and GZLP01

samkleespies:
Fully matches `d_a_obj_msdan2`.

- `Mthd_Create`: spawns 16 `MsdanSub2` child actors along the actor's facing with rotating orientation, registers the `Msdan2` event, and initialises the state from the save switch.
- `Mthd_Execute`: switch-save / event state machine.
- `Mthd_Delete`: stub.

REL is byte-identical for GZLE01; CI covers GZLJ01/GZLP01.


AH64-dll:
Clean msdan2 — the MatchingFor(3-retail) flip matches how the sibling msdan family is handled (demo build differs). One ask: the green CI is from 07-01 — could you rebase onto current main and let CI re-run so we know it still holds?

PR #1093, opened 2026-07-01
d_a_obj_balancelift 59%

samkleespies:
Partial decompilation of d_a_obj_balancelift (Tower of the Gods balance scale object).

- 36/41 functions matching (100%)
- 59.48% matched code, 98.56% fuzzy match
- All functions decompiled, 5 remaining with minor codegen differences (instruction scheduling, register allocation)


AH64-dll:
Nice honest partial on balancelift — 36/41 with the fuzzy-match numbers stated and no flip is exactly how we like to see WIP. Three small things:

1. The .gitignore change (shows as a binary diff) doesn't belong in an actor PR — could you drop it?
2. `dComIfGp_particle_set(0xC, ...)` in _execute — the repo convention is named particle ids (`dPa_name::ID_*`); could you switch it?
3. `M_mdl` is an instance member but uses the M_ static prefix — per convention it should be `mMdl`/`mpMdl` (M_ is for class statics like M_arcname).

And the CI is from 07-01 — a rebase + fresh run when you next touch it would confirm it still holds. The STATIC_ASSERT on the class size is a nice touch.

PR #1095, opened 2026-07-01
d_a_obj_Itnak #370 : 100% Match/linking

tommaso-perondi:
First time trying to decomp something. Matched at 100% and links without error.
Please correct anything i did wrong!
#370

### DIFF
--- a/include/d/actor/d_a_obj_Itnak.h
+++ b/include/d/actor/d_a_obj_Itnak.h
@@ -34,12 +34,12 @@ namespace daObjItnak {
           /* 0x2C */ f32 cyl3_pos_z;
         };
         static const char M_arcname[];
-            enum Prm_e {
-        PRM_STATE_W = 8,
-        PRM_STATE_S = 0,
-        PRM_SWSAVE_W = 8,
-        PRM_SWSAVE_S = 8,
-    };
+        enum Prm_e {
+          PRM_STATE_W = 8,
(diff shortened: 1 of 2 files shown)


AH64-dll:
Nice first PR — matching a whole actor to 100% is no small feat. The virtual dtor is correct (the original has a weak __dt__ + vtable), and the collision/sound calls match the merged conventions. A few friendly suggestions:

1. Small slop: the commented-out member `// /* 0x290 */ void* mpVtbl;` in the header — the vtable is implicit from the virtual dtor, so that line can just be deleted.
2. L_attr is declared and defined but never referenced — its values (68.0/230.0/62.0/121.0/…) appear as literals in set_collision. Either wire L_attr into set_collision or drop the struct so there's no dead data.
3. Style: the cpp mostly uses 2-space indent (project standard is 4-space), and there are a few `::`-qualified global calls (`::daObj::PrmAbstract`), C-style casts (`(int)PRM_STATE_W`, `(cXyz&)cXyz::Zero`), and `if(mdl_data != 0){` spacing nits.
4. Question: the PR doesn't flip configure.py (d_a_obj_Itnak stays NonMatching), so CI builds but never checksums the rel. Since you've matched 100%, flipping to Matching would let CI verify the checksum gate — is it intentionally left NonMatching (e.g. weak function ordering) or just not done yet?

Welcome to the project — happy to help with any of these.

PR #1101, opened 2026-07-03
d_a_npc_pm1: Match and link

### LINKED ISSUE #357: d_a_npc_pm1
Pm1

tommaso-perondi:
Closes #357

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1693,7 +1693,7 @@ config.libs = [
     ActorRel(EquivalentFor("GZLJ01", "GZLE01", "GZLP01") or MatchingFor("D44J01"), "d_a_npc_people"), # regalloc 
     ActorRel(NonMatching, "d_a_npc_pf1"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_npc_photo"),
-    ActorRel(NonMatching, "d_a_npc_pm1"),
+    ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_npc_pm1"),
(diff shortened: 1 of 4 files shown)


AH64-dll:
Nice work on npc_pm1 — matching is honest (MatchingFor 3 versions, all four builds green, all Nonmatching markers cleared) and the HIO class + base class conventions are right. A few non-blocking notes:

1. COORDINATION (worth resolving before merge): d_snap.h line 0x5A — this PR names it DSNAP_TYPE_PM1, but PR #1145 (also open) names the same value DSNAP_TYPE_NPC_PM1_KP1 since the type is shared between the PM1 and KP1 actors. Whoever merges second will conflict on that enum line; NPC_PM1_KP1 is the more descriptive name, so aligning with #1145 would avoid the conflict.
2. In eventOrder() the comma expression writes eventInfo.mCondition directly (`mCondition = mCondition | dEvtCnd_CANTALK_e`). There's an onCondition(u16) inline (f_op_actor.h:171, `mCondition |= cond`) — does `eventInfo.onCondition(dEvtCnd_CANTALK_e)` inside the comma expression also match? We prefer the inline when codegen is identical (same rule we enforce on our own PRs).
3. event_action() returns 1 on both paths — the `if ((BOOL)field_0x7CA > 0)` branch looks redundant; if the function is 16 bytes with the branch required, a brief comment would help, else simplify.
4. Minor: anmResID/BtpNum2ResID hand-expand the assert (JUTAssertion::getSDevice/showAssert/OSPanic) — JUT_ASSERT produces the same codegen if it matches.
5. Minor: next_msgStatus leaves the param unnamed in the definition (header names it pMsgNo).

None block the merge — happy to help if any need discussion.

PR #1104, opened 2026-07-05
DaObjVds retail

Azurpourpre:
#371

Currently, TU is matching in objdiff but not after linking. There is also a regalloc issue in demo.


src/d/actor/d_a_obj_Vds.cpp
```diff
@@ -5,125 +5,458 @@
...
+    daObjVds::Act_c* pVdsAct = static_cast<daObjVds::Act_c*>(i_this);
+    fopAc_ac_c* pAct = static_cast<fopAc_ac_c*>(i_actor);
+    return pVdsAct->search_switchCB(pAct);
 }
 
 /* 00000214-000002B0       .text search_switchCB__Q28daObjVds5Act_cFP10fopAc_ac_c */
-void daObjVds::Act_c::search_switchCB(fopAc_ac_c*) {
-    /* Nonmatching */
+void* daObjVds::Act_c::search_switchCB(fopAc_ac_c* i_actor) {
+    if(fopAc_IsActor(i_actor) && fopAcM_GetName(i_actor) == fpcNm_Obj_Swlight_e){
```
WilliamArnett, line 78:
```suggestion
    cXyz unused;
    if(fopAc_IsActor(i_actor) && fopAcM_GetName(i_actor) == fpcNm_Obj_Swlight_e){
```

WilliamArnett (changes requested):
Hi Azur. This should cause the TU to match for retail builds.

Azurpourpre:
Got it matching for retail (GZLE01, GZLJ01, GZLP01). D44J01 has a regalloc issue in daObjVds::Act_c::process_common

AH64-dll:
Solid work on Vds — matching is honest (MatchingFor GZLE01/GZLJ01/GZLP01 with the demo regalloc admitted, all four builds green). Non-blocking suggestions, author's call:

1. Locals named with the m- prefix: `BOOL mBrkAnm0_init` / `mBrkAnm1_init` in create_heap — our convention reserves m_ for members; suggest `brk0_init`/`brk1_init`.
2. Style: `if(`/`for(` without the space (24/7 occurrences) vs the codebase `if (`/`for (`; a few trailing-whitespace lines; the brace-less `if(...) <stmt>` in create_heap.
3. Leftover generated name `lbl_365_data_1C` in the demo branch — rename per the l_ convention or drop if truly unused.
4. In-code uncertainty comments (`so i tried to infer what i could`, `Unused, so i defined it here`, `Unused data/variable`) — the project avoids speculation comments in code; if arrays are required for .data/.rodata layout, the standard data-table comment would be more informative.
5. `static void* ds_search_switchCB` is declared static in the header (namespace-scope static in a header = per-TU duplicate; it's only referenced from the cpp) — consider moving the declaration into the cpp.
6. The d_a_obj_swlight.h change (stub `void get_power() const {}` -> `f32 get_power() const { return this->mF30; }`) is justified and correct — mF30 is the light-power field, and Vds needs the real value. Just flagging it as an intentional cross-actor edit so reviewers notice.

None of these block; happy to help if any need discussion.

PR #1105, opened 2026-07-05
d_flower Code Matching

SuperDude88:
Matches the code, doesn't handle some of the data arrays (`l_pos`, etc). I think those would best be handled with asset extraction like matDLs + textures instead of hardcoding all the vertex data.

Will work on that as a separate PR to update d_flower + similar TUs (assuming that is an approach we want to go with).


LagoLunatic:
> Will work on that as a separate PR to update d_flower + similar TUs (assuming that is an approach we want to go with).

Sounds good

PR #1106, opened 2026-07-05
JaiSeMgr

tommaso-perondi:
Not 100% match, regalloc issues.


AH64-dll:
Thanks — this is a big framework batch (30 files: JAudio + d_menu_save + d_flower + NPCs + MemCard), and the honest 'regalloc issues / not 100%' framing with the flips removed is appreciated. A few things when you're ready to move it forward:

1. The green CI is from 07-05 — could you rebase onto current main and re-run? A fresh all-4-build run is especially important here because the PR changes a shared header inline: `JAISound::setSeqInterVolumeU7` goes from an empty stub to a real body (`setSeqInterVolume(type, value / 127.0f, moveTime)`). No in-tree caller uses it today, but a stub→real inline on a shared JAudio header changes codegen for every TU that includes it, so a full-build regression check is the proof we'd want before it lands.
2. In d_flower.cpp the hand-written arrays `l_pos`/`l_pos2`/`l_pos3`/`l_texCoord` are all 1.0f — that looks like m2c placeholder data. Since d_flower isn't flipped yet, CI won't catch mismatches; could you verify them against the original .data before the flip?
3. Minor: `VERSION_SELECT(165,165,165,165)` is redundant (all four versions equal).

Also a suggestion: at ~13.8K lines across JAudio, menus, flowers, and NPCs, this is very hard to review as one PR — when it approaches a flip, splitting into per-system PRs would help a lot. Happy to re-review either way.

PR #1107, opened 2026-07-06
d_a_npc_bm1 100% Matching Retail

### LINKED ISSUE #320: d_a_npc_bm1
Bm1, Bm2, Bm3, Bm4, Bm5

WilliamArnett:
Closes #320


src/d/actor/d_a_npc_ls1.cpp
```diff
@@ -1609,7 +1609,7 @@ bool daNpc_Ls1_c::telescope_proc() {
             case 1:
                 temp_msg_no = m810;
                 m810 = 0xBC1;
-                if (chkTelescope_sph(bm1_npc_p->attention_info.position, 0.0f, l_HIO.mPrm.m2C)) {
+                if (chkTelescope_sph(bm1_npc_p->getOdoroki(), 0.0f, l_HIO.mPrm.m2C)) {
```
LagoLunatic: This one doesn't add up, `Odoroki` is Japanese for "astonishment" which isn't related to the attention position. And when inlines do return a position, they usually return `cXyz` instead of `cXyz&`, which would not match the inline size in the debug map. Also, attention_info.position is from the actor base class, so the field is public and doesn't need a getter anyway.

By process of elimination I think `getOdoroki` is supposed to return `bm1_npc_p->m880`, as that's the only other actor-specific field used here. You can change bm1's fields to `private` so that the compiler will complain about them, this makes it more obvious which fields are used by other classes yet are missing inlines.

include/d/actor/d_a_npc_bm1.h
```diff
@@ -1,60 +1,111 @@
...
+#include "JSystem/JParticle/JPAEmitter.h"
 #include "d/d_npc.h"
 #include "f_op/f_op_actor.h"
+#include "m_Do/m_Do_hostIO.h"
 
 class J3DNode;
 
+typedef s8 SpecificType;
+enum {
+    Quill_0,
```
LagoLunatic: You should avoid putting typedefs and enums in actor headers like this. Any other actor that needs to include this actor's header will now have these things polluting its namespace and possibly causing conflicts (e.g. `Invalid` is a very generic name).

There are 3 options for how you can fix this:
* Put them all in the .cpp file so they're out of the header (won't work in this case since then the inlines can't use them, but can be done for actors that don't have inlines)
* Prefix everything with the name of the class like `daNpcBm1_TYPE_Invalid_e` instead of `Invalid`
* Simply move the typedefs and enums into the class definition so that they're namespaced appropriately (even if you do this you should still prefix them with something, like `daNpc_Bm1_c::TYPE_Invalid_e` for example)

Also, it doesn't look like you used these enums you created in most places inside the .cpp file where they would be relevant.

include/d/actor/d_a_npc_bm1.h
```diff
@@ -120,118 +171,224 @@ class daNpc_Bm1_c : public fopNpc_npc_c {
...
-    /* 0x6D5 */ u8 m6D5[0x734 - 0x6D5];
+    /* 0x6D0 */ s8 m_hed_jnt_num;
+    /* 0x6D1 */ s8 m_bbone_jnt_num;
+    /* 0x6D2 */ s8 m_nec_jnt_num;
+    /* 0x6D3 */ s8 m_arm_L_jnt_num;
+    /* 0x6D4 */ s8 m_arm_R_jnt_num;
+                J3DModel* mpBinderModel;
+                J3DModel* mpBagModel;
+                J3DModel* mpKnifeModel;
+                J3DModel* mpStickModel;
```
LagoLunatic: A lot of these fields are missing offset comments

include/d/actor/d_a_npc_bm1.h
```diff
@@ -120,118 +171,224 @@ class daNpc_Bm1_c : public fopNpc_npc_c {
...
-    /* 0x80C */ cXyz m80C;
+    /* 0x794 */ dNpc_PathRun_c mPathRun;
+                dNpc_EventCut_c mEventCut;
+                fpc_ProcID mPartnerProcID;
+    /* 0x80C */ cXyz mInitialPos;
     /* 0x818 */ csXyz m818;
     /* 0x81E */ u8 m81E[0x820 - 0x81E];
     /* 0x820 */ cXyz mEyePos;
-    /* 0x82C */ u8 m82C[0x86E - 0x82C];
+                cXyz mAttentionPos;
```
LagoLunatic: Why is this called `mAttentionPos`? It doesn't seem related to `attention_info.position` or the attention system in general to me, unless I missed something

include/d/actor/d_a_npc_bm1.h
```diff
@@ -120,118 +171,224 @@ class daNpc_Bm1_c : public fopNpc_npc_c {
...
-class daNpc_Bm1_HIO_c {
-public:
-    daNpc_Bm1_HIO_c();
+}; 
+STATIC_ASSERT(sizeof(daNpc_Bm1_c) == DEMO_SELECT(0x90C,0x908));
 
-public:
-    /* Place member variables here */
+
+struct hio_prm_c{
```
LagoLunatic: Same as the enums this can't go in the header for namespace pollution reasons. It can go in the .cpp where it's used

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+public:
+    daNpc_Bm1_HIO_c();
+    virtual ~daNpc_Bm1_HIO_c(){};
+    void genMessage(JORMContext* ctx);
+public:
+
+    s8 m4;
+    s32 m8;
+    daNpc_Bm1_childHIO_c children[10];
+    /* Place member variables here */
```
LagoLunatic: Missing offset comments, and you can get rid of the placeholder comment

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+    
+    short sVar1;
+    if (dComIfGp_evmng_endCheck(m862[m86A])) {
+        sVar1 = m86A;
+        switch(m86A){
+            case 0:
+            m8FD = 1;
+            m87C = 1;
+            break;
+            case 1:
```
LagoLunatic: Indentation is different here from all the other switches in this file

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+            dComIfGs_onEventBit(0x1F40);
```
LagoLunatic: dSv_event_flag_c enum

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+        mpMorf->getModel()->setUserArea((u32)this);
+        mAcchCir.SetWall(30.0f,50.0f);
+#if VERSION == VERSION_DEMO
+
+        mObjAcch.Set(fopAcM_GetPosition_p(this),fopAcM_GetOldPosition_p(this),this,1,&mAcchCir,fopAcM_GetSpeed_p(this),NULL,NULL);
+
+#else
+        mObjAcch.Set(&current.pos,&old.pos,this,1,&mAcchCir,&speed);
+
+#endif
```
LagoLunatic: Fake version difference, the demo one with the inlines should match all versions (unless something unusual is happening)

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+
+    }
+    if(mpStickModel){
+        g_env_light.setLightTevColorType(mpStickModel,&tevStr);
+        mDoExt_modelEntryDL(mpStickModel);
+    }
+    shadowDraw();
+    switch(mType){
+        case 1:
+            dSnap_RegistFig(0x8D,(fopAc_ac_c *)this,1.0,1.0,1.0);
```
LagoLunatic: Missing snap enum, unnecessary casts (this already inherits fopAc_ac_c) and these should be float literals not doubles
```suggestion
            dSnap_RegistFig(DSNAP_TYPE_UNK8D,this,1.0f,1.0f,1.0f);
```

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+    if(mpFeather1EmitterL){
+#if VERSION > VERSION_DEMO
+        mpFeather1EmitterL->clearStatus(0x40);
+#endif
+        mpFeather1EmitterL->becomeInvalidEmitter();
+        mpFeather1EmitterL = NULL;
+    }
+    if(mpFeather1EmitterR){
+#if VERSION > VERSION_DEMO
+        mpFeather1EmitterR->clearStatus(0x40);
```
LagoLunatic: quitImmortalEmitter

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
-void daNpc_Bm1_c::decideType(int, int) {
-    /* Nonmatching */
+bool daNpc_Bm1_c::decideType(int i_type_param, int i_spawn_cond_param) {
+    
+    mSpawnCondition = 0;
+    mType = t_Invalid;
+    mSpecificType = Invalid;
+    fpc_ProcID procName = fopAcM_GetName(this);
+    switch(procName){
+    case DEMO_SELECT(0x149,0x148):
```
LagoLunatic: Proc name enum is in `f_pc_name.h`, no need for DEMO_SELECT

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+    if (mpCurrMsg != NULL) {
+        mOldMsgStat = mpCurrMsg->mStatus;
+    } else {
+        mOldMsgStat = 0;
+    }
+    talk(1);
+    if (mpCurrMsg != NULL) {
+        switch (mpCurrMsg->mStatus) {
+        case 0x6:
+        case 2:
```
LagoLunatic: fopMsg_MessageStatus_e enum

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4411 @@
...
+        *i_msg_no = 0x1997;
+        break;
+    case 0x1998:
+        *i_msg_no = 0x1999;
+        break;
+    case 0x199E:
+        *i_msg_no = 0x199F;
+        break;
+    case 0x199F:
+        if(dComIfGs_getBeastNum(2) >= 0x14){
```
LagoLunatic: BeastIdx enum

include/d/actor/d_a_npc_bm1.h
WilliamArnett: Its used as the target for the LookBack function, I believe. I do agree the name isn't great. I may just change it back to m82C, because I don't know what a good naming convention for it would be

src/d/actor/d_a_npc_ls1.cpp
WilliamArnett: Yeah, this makes a lot of sense. I was looking at a memory viewer in game yesterday, and m880 is sent to true when Quill looks up at the Helmaroc King during the telescope section at the start of the game.

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4432 @@
...
 /* 00001B5C-00001CF4       .text setAnm_anm__11daNpc_Bm1_cFPQ211daNpc_Bm1_c9anm_prm_c */
-void daNpc_Bm1_c::setAnm_anm(daNpc_Bm1_c::anm_prm_c*) {
-    /* Nonmatching */
+BOOL daNpc_Bm1_c::setAnm_anm(daNpc_Bm1_c::anm_prm_c* i_anmPrmP) {
+
+    if(mBckResIndex == i_anmPrmP->mBckResIndex){
+        return 1;
+    }
+    mBckResIndex = i_anmPrmP->mBckResIndex;
+    dNpc_setAnmFNDirect(mpMorf,i_anmPrmP->mLoopMode,i_anmPrmP->mMorf,i_anmPrmP->mSpeed,anmNum_toResID(mBckResIndex),NULL,mArcName);
```
LagoLunatic: These field names are using the term "ResIndex" incorrectly. A res index is an index in a resource arc file (second argument passed to `dComIfG_getObjectRes`).

Based on the func `anmNum_toResID`, it seems the thing you're naming `mBckResIndex` should be something like `mAnmNum`, which maps to multiple different resources (because the wings are separate models from the body).

Similarly, `mResIndex` eventually gets passed to `btpNum_toResID`, so it seems it should be named `mBtpNum` instead.

include/d/actor/d_a_npc_bm1.h
```diff
@@ -120,118 +167,190 @@ class daNpc_Bm1_c : public fopNpc_npc_c {
...
+    /* 0x87B */ s8 m87B;
+    /* 0x87C */ s8 m87C;
+    /* 0x87D */ u8 m87D;
+    /* 0x87E */ u8 m87E;
+    /* 0x87F */ bool m87F;
+private:
+    /* 0x880 */ bool m880;
+    /* 0x881 */ bool m881;
+    /* 0x882 */ bool m882;
+public:
```
LagoLunatic: Is there a reason only these 3 specific fields are private? For NPCs usually I'd assume all fields to be private, though it's also totally fine to leave everything public if it's too hard to match something properly with them private.

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -5,810 +5,4432 @@
...
 /* 000058B8-000059E0       .text privateCut__11daNpc_Bm1_cFi */
-void daNpc_Bm1_c::privateCut(int) {
-    /* Nonmatching */
+void daNpc_Bm1_c::privateCut(int arg0) {
+    
+    static char* a_cut_tbl[] = {"ACTION","360_TRN"};
+    
+    if (arg0 != -1) {
+        m8F5 = dComIfGp_evmng_getMyActIdx(arg0,
+                        a_cut_tbl,2,1,0);
```
LagoLunatic: That's an array size so it's good to use the macro in case modders ever change the array length
```suggestion
        m8F5 = dComIfGp_evmng_getMyActIdx(arg0,
                        a_cut_tbl,ARRAY_SIZE(a_cut_tbl),TRUE,0);
```

LagoLunatic (commented):
Couple more things I missed

src/d/actor/d_a_npc_bm1.cpp
WilliamArnett: That makes sense. I don't find the naming scheme for variables like this all that intuitive, so I just try to copy other TUs with stuff like this, and that leads to mistakes on my end. I took this naming scheme from Aryll's TU `d_a_npc_ls1`, since this part of the `anm_prm_c` struct is seemingly being used in the same way. I'll do it the way you're talking about here from now on, since it is a clearer name.

WilliamArnett: Looking at `d_a_npc_ls1` some more, the reason why its `mResIndex` instead of something like `mBtpNum` is because `init_texPttrnAnm` is using that member variable to set both the btp and the btk. I didn't really notice that difference until now.

include/d/actor/d_a_npc_bm1.h
WilliamArnett: I misinterpreted what you said when we were talking about the inline calls used in Aryll's `telescope_proc` function. You said I should set the fields to private so that its more obvious that the inline functions should be used, but I only did that for the ones that are actually used in those inlines. I'll set all the fields as private

src/d/actor/d_a_npc_bm1.cpp
LagoLunatic: Oh thanks for pointing that out, seems npc_ls1 also uses the wrong naming scheme, I'll have to fix that. It's a pretty subtle mistake.

PR #1109, opened 2026-07-07
d_a_obj_Vteng demo 100% matching

ateddy42:


AH64-dll:
Clean demo-enable — full Matching across all four builds, and the DEMO_SELECT constants + version guards are exactly right. One tiny nit: the HIO child name uses Shift-JIS byte escapes (`\x83\x4b…` = ガノン城の天蓋). Merged actors write the literal Japanese string instead (d_a_bdk "戦闘用大怪鳥", d_a_bgn "Ｇ（クグツ）", d_a_bmd "森ボス") — codegen-equal either way, but the literal matches the convention and is readable at a glance.

PR #1114, opened 2026-07-11
d_a_obj_iceisland 100% demo

ateddy42:


AH64-dll:
Nice — demo 100% verified across all four builds, and the VERSION-guarded demo branches (start point 2, REG17_S(0) = 0x80, reg17 color math) are exactly the right shape. One tiny nit: the two includes at the top got a leading space (line starts with space before #) — cosmetic only, worth stripping for consistency.

PR #1118, opened 2026-07-14
d_a_obj_YLzou matching

Dwlee99:
First contribution!

Matching d_a_obj_YLzou for https://github.com/zeldaret/tww/issues/373.

Not doing too much naming since I don't want to misname stuff.

Thanks to @WilliamArnett for getting it to match on demo!


AH64-dll:
Great first contribution. One fix before merge: the demo HIO class declares `void genMessage(JORMContext*);` but never defines it — every HIO class in the codebase defines it empty (e.g. d_a_bgn.cpp), and JORReflexible::genMessage is pure virtual under DEBUG (m_Do_hostIO.h), so a debug build would fail to instantiate l_HIO. Add `void genMessage(JORMContext*) {}` to the class body.

PR #1121, opened 2026-07-17
Match d_a_obj_htetu1

999sian:
Fully decompiles `d_a_obj_htetu1` (Tower of the Gods yellow gate) — **40/40 functions, 100% code + data**, rel SHA byte-exact across all four builds.

- `ActorRel(Matching)` for D44J01, GZLJ01, GZLE01, and GZLP01.
- D44J01 uses the original demo particle layout: a separate `JPABaseEmitter*` before `dPa_followEcallBack`, making each splash object `0x30` bytes instead of retail's `0x2C`.
- Demo-specific resource deletion, heap guards, emitter cleanup, and unlock scheduling are selected with existing version helpers/guards.
- Verified locally against D44J01, GZLJ01, and GZLP01: 40/40, 100% code/data, exact REL SHA. GZLE01 is covered by CI.

Supersedes #1119.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1749,7 +1749,7 @@ config.libs = [
     ActorRel(Matching,    "d_a_obj_homen"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_obj_homensmoke"),
     ActorRel(NonMatching, "d_a_obj_hsehi1"),
-    ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_obj_htetu1"),
+    ActorRel(Matching,    "d_a_obj_htetu1"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_obj_iceisland"),
     ActorRel(Matching,    "d_a_obj_jump"),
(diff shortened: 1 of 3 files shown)


999sian:
Upgraded #1121 to full `Matching` across all four builds.

The demo mismatch was source-level rather than an irreducible float-pool artifact:
- D44J01 keeps a separate `JPABaseEmitter*` before the follow callback (`0x30` splash object vs retail `0x2C`).
- Demo uses `dComIfG_resDeleteDemo`, omits retail heap/null-reset behavior, and loads `mUnlockSpeed` later.
- The final `_execute()` register swap was fixed by the original-style `if (check_sw())` instead of `!= FALSE`; this leaves only benign relocation-base labels in raw disassembly.

Local verification:
- D44J01: 40/40, 100% code/data, REL SHA `c682b31f40aa6ee02acf82665d9a3f782f29e878`
- GZLJ01: 40/40, 100% code/data, REL SHA `734630f4b4cc6b97da1acafe7d77f1944f927934`
- GZLP01: 40/40, 100% code/data, REL SHA `75e808ccf8fe41d3a12548e5de493e94fb3c1566`
- CI: D44J01, GZLJ01, GZLE01, and GZLP01 all pass.

PR #1123, opened 2026-07-17
d_a_obj_vyasi: 100% matching (GZLE01)

LostGoron:
Palm tree. 100% (GZLE01).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1788,7 +1788,7 @@ config.libs = [
     ActorRel(Matching,    "d_a_obj_vmsms"),
     ActorRel(Matching,    "d_a_obj_volcano"),
     ActorRel(NonMatching, "d_a_obj_vtil"),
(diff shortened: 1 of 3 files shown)


AH64-dll:
Solid work. Small cleanups before merge:

1. `M_bck_data` at 0x04F0 is an instance member but is named with the `M_` static prefix — per conventions it should be `mpBckData` (m/mp + UpperCamelCase; `M_` is reserved for class statics like `M_arcname`).
2. The `Attr_c`/`L_attr` struct is never referenced, while its values (79.0f radius, 250.0f height) appear as literals in `set_collision`. Either wire the table in or remove it — right now the data is duplicated as dead code.
3. Question: the repeated `mCyl.SetTgVec(...)` / `mCyl.OnTgNoHitMark()` inside the mSph init loop (i < 8) — intentional (matches original), or leftover from the mCyl setup above?

PR #1124, opened 2026-07-18
Match d_a_fallrock_tag

999sian:
Fully decompiles `d_a_fallrock_tag` (`daFallRockTag_c` / TagRock) — **7/7 functions, 100% code + data**, rel SHA byte-exact.

The actor is a schedule-triggered spawner that drops falling rocks (`fpcNm_FallRock_e`) via the `cDyl` dynamic-link pattern, gated by the day/night schedule bits (`dKy_get_schbit` / `dKy_get_schbit_timer`). Positional SE via `fopAcM_seStart`.

Marked `Matching` (the linked sibling `d_a_fallrock` is also `Matching`). Struct/naming follow the coding guidelines (Dolphin typedefs, `i_`/`m` prefixes, offset comments).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1621,7 +1621,7 @@ config.libs = [
     ActorRel(Matching,    "d_a_deku_item"),
     ActorRel(Matching,    "d_a_dk"),
     ActorRel(Matching,    "d_a_dummy"),
-    ActorRel(NonMatching, "d_a_fallrock_tag"),
+    ActorRel(Matching, "d_a_fallrock_tag"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_fan"),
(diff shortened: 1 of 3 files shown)


999sian:
Updated to `MatchingFor("GZLJ01", "GZLE01", "GZLP01")` — the three retail builds are byte-exact, but the D44J01 demo build genuinely differs (its `d_a_fallrock_tag.rel` diverges), so it is excluded like the other demo-divergent rels. CI is green across all four versions now.

999sian:
Upgraded from `MatchingFor` to full `Matching` — the demo (D44J01) now matches too. The only demo divergence was `create()` ordering `fopAcM_ct` after the `cDyl_LinkASync` phase check (retail runs it first); a `VERSION` guard keeps both forms. Verified byte-exact rel SHA on D44J01 and GZLP01 locally; CI covers GZLE01/GZLJ01.

PR #1060, opened 2026-06-17
d_a_obj_ekskz 88%

samkleespies:
Partial match for d_a_obj_ekskz (Gale Isle wind-blowing stone statue).

52/53 functions at 100%. Execute remains nonmatching at 99.92% due to the constant-pool base anchoring to ...rodata.0 instead of M_arcname.

Checked with:
- ninja
- ninja changes_all

Closes #388

PR #1130, opened 2026-07-19
Match d_a_obj_light

LostGoron:
47/47 functions. 100% (GZLE01).


AH64-dll:
Nice work — 47/47 is solid. A few things before merge:

1. Please replace the raw sound IDs with the named constants: `0x6a03` = `JA_SE_OBJ_L_HOUSE_FIRE`, `0x6a04` = `JA_SE_OBJ_L_HOUSE_LIGHT`, `0x303b` = `JA_SE_OBJ_L_HOUSE_ROUND` (all in include/JAZelAudio/JAZelAudio_SE.h:624/2026/2027). We enforce named `seStart` ids repo-wide — see the discussion on #1060 for the same finding.
2. The static `L_attr` (a `L_attr_t` with `0x00800002, 0x0014`) is defined but never referenced. Remove it, or wire it in, unless it's required for data layout — in which case a comment explaining that would be good.
3. Question: in `set_mtx`, `cMtx_copy(mDoMtx_stack_c::get(), (&mMtx)[i]);` sits inside `if (i == 0)` with a comment saying only i==0 reaches here. Does plain `mMtx` also match, or is the indexed form needed to reproduce the codegen? If the latter, a shorter comment (e.g. referencing the loop induction) would be easier to maintain.

Also note: #1063 covers the same actor — you may want to coordinate so the maintainers only need to review one.

PR #1132, opened 2026-07-20
JAIZelBasic Regswap

WilliamArnett:
Extension of #570


AH64-dll:
Hi — nice regswap work, but the D44J01 build is failing: the shasum gate fails on framework.dol and 413 units (07-31 run). Since the other three builds pass, this looks like the demo build genuinely diverges after the shared-header changes (e.g. the member type/layout updates and the demo `MAX_CONCURRENT_SE_NUM = 26` / `temppadding` changes). Could you check whether D44J01 needs VERSION-guarding for these, or whether the new types break codegen there? Also: the PR title says 'Regswap' but the diff includes signature/layout changes (bool→BOOL, void→f32, m_bgm_mute_state layout) — those go beyond register swaps and affect every TU including the header, so it'd be good to call them out in the description. Once D44J01 is green this looks ready.

PR #1133, opened 2026-07-20
Match d_a_saku 100%

LostGoron:
71/71 functions. 100% (GZLE01)


AH64-dll:
Two convention nits: (a) class static members should use the M_ prefix like the rest of the actor headers (M_arcname, M_smoke_alpha, M_dust_color...) rather than m_/no prefix — m_cyl_src/m_at_cyl_src are pre-existing so they can stay; (b) `g_env_light.settingTevStruct(0, ...)` should use the named TEV_TYPE_ACTOR (== 0). The const-static placement comment is great, keep it.

PR #1134, opened 2026-07-22
d_a_npc_zk1 100% matching

### LINKED ISSUE #367: d_a_npc_zk1
Zk1

slock83:
Hello,
Started working on d_a_npc_Zk1.
100 % matching for retail
100 % matching for demo too thanks to @WilliamArnett 

Closes #367


slock83:
Should be ready to merge, let me know if I need to change something

AH64-dll:
Great work reaching 100% on both retail and demo — but two things need attention before merge:

1. Admitted fake match: `speedF = speedF; //TODO: Fake Match` in the demo path (d_a_npc_zk1.cpp ~947, inside the #if VERSION == VERSION_DEMO switch). A self-acknowledged fake match alongside a full Matching flip and a '100% demo' claim is a problem: if the self-assignment is genuinely required for codegen (e.g. the original had a dead store), document it properly with an explanation instead of 'TODO: Fake Match'; if the demo isn't actually matching, the claim and flip are wrong. Which is it?
2. No CI: the branch has no checks reported ('no checks reported on the d_a_npc_zk1 branch') — pushing to the build CI is what verifies the Matching flip on all four versions. Without it the 100% claims are unverified.

Neither is a blocker on the decompiling itself — both are honesty/verification items.

PR #1137, opened 2026-07-25
d_a_mgameboard close to matching

Dwlee99:
Got this TU close to matching but I need help with some function relocations (cXyz destructor is supposed to come after set_mtx) and .rodata. I also haven't fixed the indices for the `dComIfG_getObjectRes` calls on JPN or PAL since I don't have those ROMs.


AH64-dll:
Nice progress on a big actor. Before this can merge, some cleanup:

1. Please remove the scratch symbol notes left in include/d/d_seafightgame.h (`// checkState__20dSeaFightGame_info_cFUcUc (func,weak) found in d_a_mgameboard.o`, `// ?? checkState(int i) { }`, etc.).
2. Remove the commented-out `// # include` lines in the VERSION ifdef block of d_a_mgameboard.cpp.
3. Replace the `// TODO: JPN and PAL` markers — `VERSION_SELECT(..., 0, ..., 0)` currently passes resource index 0 for JPN/PAL, which resolves to the wrong resource. If JPN/PAL resource names aren't in the tree yet, resolve them before the flip (or note which resources are missing so a follow-up can land with the flip).
4. The unused local `u8 ship_cur_health = mSeaFightGame.getCurHP(i); // unused but inline is in the debug maps` — if the weak inline is layout-required, a comment on the class is the docs-blessed way (see the weak-func-order convention); otherwise drop the dead read.
5. `f32 shadow_alpha = 80.5f; // alpha is a u8 but this is needed to match ???` — the '???' reads as a scratch note; please clarify or restructure.
6. The new `set()` inline added to the DOL class dDlst_2DNumber_c (d_2dnumber.h): the original REL's `_execute` writes `mValue` directly (`stw` to 0x3C), and no `set__16dDlst_2DNumber_cFi` symbol exists in the DOL — so the inline-for-direct-write diverges from the original source form. If direct member assignment also matches, prefer that; if the inline is needed, explain why.

PR #1139, opened 2026-07-31
JPA improvements

Mrkol:


AH64-dll:
Great progress — nice to see these billboard/stripe visitors come in. Before merge:

1. `f32* hack = &mtx.mMtx[0][0];` (and `hack2` in StripeCross) are never referenced — remove them unless they're required for the match, in which case a comment explaining would be better than the name.
2. The new `SMatrix33R`/`TMatrix33`/`TRotation3::setXYZDir` templates in JGeometry.h/TPosition3.h — could you double-check the naming/behavior against how the sibling projects (e.g. TP) modeled the same JPA code? The transposed `at()/ref()` storage comment is helpful; just want to make sure we're matching the original JSystem shape.
3. Minor: `JGeometry::TVec3<f32> zdir(*(Vec*)...)` — a C-style cast; consider the explicit cast form used nearby.

PR #1141, opened 2026-07-31
d_menu work

mattias-blum:
d_menu_collect class members cleaned up, some functions at 100% matching


include/d/d_menu_collect.h
```diff
@@ -107,7 +113,7 @@ class dMenu_Collect_c : public dMenu_base_c {
 
 private:
     /* 0x0000 */ // vtable
-    /* 0x0004 */ J2DScreen* m004;
+    /* 0x0004 */ MyScreen* scrn;
```
LagoLunatic, line 116: I might be missing something, how do we know this is a `MyScreen` as opposed to the parent class?

mattias-blum: The asm from the `_create` function suggests so 
Screenshot 2026-07-31 120833

LagoLunatic: Oh it has a vtable, duh

PR #1144, opened 2026-08-02
D_a_npc_ba1 100% matching

slock83:
#318


src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+    }
+    
+    return state;
 }
 
 /* 000043E8-00004658       .text create_Anm__11daNpc_Ba1_cFv */
-void daNpc_Ba1_c::create_Anm() {
-    /* Nonmatching */
+J3DModelData* daNpc_Ba1_c::create_Anm() {
+    J3DModelData* a_mdl_dat = (J3DModelData*)dComIfG_getObjectIDRes("Ba",10);
```
LagoLunatic: dRes_ID_BA enum

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+    JUT_ASSERT(VERSION_SELECT(0xB0C, 0xB0B, 0xB0C, 0xB0C),a_mdl_dat != NULL);
+    mpMorf = new mDoExt_McaMorf(
+        a_mdl_dat,
+        NULL,
+        NULL,
+        (J3DAnmTransform*)dComIfG_getObjectIDRes(
+            "Ba", 
+            7
```
LagoLunatic: dRes_ID_BA enum

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+
+    return a_mdl_dat;  
 }
 
 /* 00004658-00004714       .text create_itm_Mdl__11daNpc_Ba1_cFv */
-void daNpc_Ba1_c::create_itm_Mdl() {
-    /* Nonmatching */
+bool daNpc_Ba1_c::create_itm_Mdl() {
+    mpClothModel = NULL;
+    J3DModelData* a_mdl_dat = (J3DModelData*)dComIfG_getObjectIDRes("Ba",9);
```
LagoLunatic: dRes_ID_BA

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+    }
+    
+    return ret;
 }
 
 /* 000012E8-0000137C       .text chk_drct__11daNpc_Ba1_cFf */
-void daNpc_Ba1_c::chk_drct(float) {
-    /* Nonmatching */
+bool daNpc_Ba1_c::chk_drct(float i_param_1) {
+#if VERSION == VERSION_DEMO
```
LagoLunatic: Looks like a fake version diff, have you tried using the top version for all versions and checking if it still matches them?

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+        return this == attention.ActionTarget(0);
+    }
 }
 
 /* 00001A2C-00001A94       .text setAttention__11daNpc_Ba1_cFb */
-void daNpc_Ba1_c::setAttention(bool) {
-    /* Nonmatching */
+void daNpc_Ba1_c::setAttention(bool i_setEyePos) {
+
+#if VERSION == VERSION_DEMO
```
LagoLunatic: Looks like a fake version diff, have you tried using the top version for all versions and checking if it still matches them?

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
-void daNpc_Ba1_c::eMove_CHK_FAIRY_MOV_1() {
-    /* Nonmatching */
+bool daNpc_Ba1_c::eMove_CHK_FAIRY_MOV_1() {
+    bool ret = false;
+    if(!cLib_calcTimer(&mEvTimer)) {
+        if(partner_srch_sub(searchActor_Fa)) {
+            daNpc_Fa1_c * actor_fa;
+            actor_fa = (daNpc_Fa1_c*) searchByID(mPartnerProcID);
+            if(actor_fa != NULL) {
+#if VERSION == VERSION_DEMO
```
LagoLunatic: Looks like a fake version diff, have you tried using the top version for all versions and checking if it still matches them?

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+            fopAcM_posMoveF(this, mStts.GetCCMoveP());
+        }
+        if(!m7F5) {
+            m78A = current.angle;
+            shape_angle = m78A;
+        }
+    }
+    eventOrder();
+    setMtx(false);
+#if VERSION == VERSION_DEMO
```
LagoLunatic: Looks like a fake version diff, have you tried using the top version for all versions and checking if it still matches them?

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+        0x272E0
+    };
+
+    fopAcM_ct_Retail(this, daNpc_Ba1_c);
+    cPhs_State state = dComIfG_resLoad(&mPhs,"Ba");
+    if(state != cPhs_COMPLEATE_e) {
+        return state;
+    } 
+#if VERSION == VERSION_DEMO
+    u32 tmp = fpcM_GetParam(this) & 0xFF;
```
LagoLunatic: Looks like a fake version diff, have you tried using the top version for all versions and checking if it still matches them?

Also should be `fopAcM_GetParam` instead of `fpcM_GetParam

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
 /* 000004F8-000005F4       .text init_BA1_1__11daNpc_Ba1_cFv */
-void daNpc_Ba1_c::init_BA1_1() {
-    /* Nonmatching */
+bool daNpc_Ba1_c::init_BA1_1() {
+    // This int + bool thing is weird, but can't get it to match without ...
+    int res;
+    bool ret;
+    res = dComIfGs_isEventBit(dSv_event_flag_c::UNK_0520);
+    ret = res == 0;
+    if(!ret) { // weird, but couldn't get it to match without this ...
```
LagoLunatic: Doesn't seem weird, they're doing an early return on failure, can remove this comment

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+        return ret;
+    } else {
+        res = dComIfGs_isEventBit(dSv_event_flag_c::UNK_0001);
```
LagoLunatic: Don't need else after early return

```suggestion
        return ret;
    }
    res = dComIfGs_isEventBit(dSv_event_flag_c::UNK_0001);
```

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
-    /* Nonmatching */
+bool daNpc_Ba1_c::init_BA1_3() {
+    /* Nonmatching */
+    int res; // Same as init_1
+    bool ret;
+    res = dComIfGs_isEventBit(dSv_event_flag_c::UNK_0520);
+    ret = res;
+    if(!ret) {
+        return ret;
+    } else {
```
LagoLunatic: Same

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
-    /* Nonmatching */
+bool daNpc_Ba1_c::init_BA1_4() {
+    // Same as 1 and 3
+    int res;
+    bool ret;
+    res = dComIfGs_isEventBit(dSv_event_flag_c::UNK_0520);
+    ret = res;
+    if (!ret) {
+        return ret;
+    }else {
```
LagoLunatic: Same

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+    
+    temp.set(503, 0, -91);
+    if(cLib_calcTimer(&mEvTimer)) {
+        return FALSE;
+    } else {
+        if((int)mEyeOffset != 0) {
+            cLib_chaseF(&mEyeOffset, 0.0f, 2.0f);
+            return FALSE;
+            
+        } else {
```
LagoLunatic: Same, both elses

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
-void daNpc_Ba1_c::eInit_CHK_FAIRY_(int*) {
-    /* Nonmatching */
+void daNpc_Ba1_c::eInit_CHK_FAIRY_(int* i_param_1) {
+    eInit_setEvTimer(i_param_1);
 }
 
 /* 000020E8-000020F8       .text eInit_prmFloat__11daNpc_Ba1_cFPff */
-void daNpc_Ba1_c::eInit_prmFloat(float*, float) {
+f32 daNpc_Ba1_c::eInit_prmFloat(float* i_param_1, float i_param_2) {
     /* Nonmatching */
```
LagoLunatic: Comment can be removed

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+        
+    }
+    if(i_param_2 != NULL) {
+        cXyz tmp;
+        switch (*i_param_2) {
+            case 0:
+            m7F7 = 1;
+            break;
+            case 1:
+            m7F7 = 0;
```
LagoLunatic: Several of the `case`s don't have indention which makes them a bit hard to read, also applies to other funcs in this PR

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+            break;
+        case 0x7FB:
+        case 0x7FC:
+        case 0x7FD:
+        case 0x7FA:
+            if(dComIfGs_isGetBottleItem(0x55)) {
+                if(!dComIfGs_checkEmptyBottle()) {
+                    *i_msg_no = 0x803;
+                }
+                else if(dComIfGs_checkBottle(0x55)) {
```
LagoLunatic: 0x55 is `dItemNo_SOUP_BOTTLE_e` here

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+            break;
+        case 0x7F3:
+            *i_msg_no = 0x7F4;
+            break;
+        case 0x7F4:
+        // Moving this case bellow the goto 
+        // (thus removing the need for the label) 
+        // does not match ...
+        // Maybe a missing inline ?
+        noBottle:
```
LagoLunatic: I'm fairly sure the goto has to be real considering how MWCC doesn't like to reverse the order of code, so might as well remove these comments. Though I would rename the `noBottle` label since it seems the goto happens when the player has never owned the soup (i.e. hasn't previously healed grandma), not when they don't own a bottle. Just `case_0x7F4` works if you can't think of a good name.

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+        current.pos.x = -290.0;
+        current.pos.y = 0.0;
+        current.pos.z = 110.0;
+        current.angle.y = -0x8000;
+        mInitialPos.set(current.pos);
+
+        // Maybe missing a .set(CsXyz) inline function
+        mInitialAngle.x = current.angle.x;
+        mInitialAngle.y = current.angle.y;
+        mInitialAngle.z = current.angle.z;
```
LagoLunatic: There's no `set(CsXyz)` in the debug maps so you can remove the comment, but if you want you can try the `set(s16, s16, s16)` inline and see if that still matches.

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+        }
+    }
+    return 1;
 }
 
 /* 00002E30-00002F70       .text wait_1__11daNpc_Ba1_cFv */
-void daNpc_Ba1_c::wait_1() {
-    /* Nonmatching */
+BOOL daNpc_Ba1_c::wait_1() {
+    /* Nonmatching */
```
LagoLunatic: Comment

src/d/actor/d_a_npc_ba1.cpp
```diff
@@ -1,499 +1,2163 @@
...
+    ret = res;
+    if(!ret) {
+        return ret;
+    } else {
+        res = dComIfGs_isEventBit(dSv_event_flag_c::GRANDMA_HEALED);
+        ret = res ==0;
+        if(ret) {
+            attention_info.distances[1] = 0x1B;
+            attention_info.distances[3] = 0x1C;
+            actor_status &= ~0x180; // Fakematch, didn't find a fitting accessor that would match
```
LagoLunatic:
```suggestion
            fopAcM_OffStatus(this, fopAcStts_CULL_e | fopAcStts_NOCULLEXEC_e);
```

src/d/actor/d_a_npc_ba1.cpp
slock83: Agreed. Blunder with the auto formatter, I thought I fixed them, but apparently I missed some (or the formatter messed things up again ...)

src/d/actor/d_a_npc_ba1.cpp
slock83: I'll check them all again, but I first matched on US, then fixed JPN, then demo. So unless I messed up (which is possible), every time there's a version check, it's because it wouldn't match 100% otherwise. I guess they did a formatting pass between demo and release on this actor

LagoLunatic: Demo uses different compiler flags than the 3 retail versions. So if you did demo last then you would see a bunch of mismatches that you have to fix, but fixing them on demo usually does not break the other 3 versions if it was just compiler flag silliness, it only requires an ifdef if the change is not functionally equivalent. That's why I said "try the top one for all versions", and only for the version differences that are equivalent. Because it should be possible for the demo version of the code to match all versions.

slock83: I was indeed mistaken, the top version does match for every version. Is the assembly slightly different in the demo because of compiler changes ?

src/d/actor/d_a_npc_ba1.cpp
slock83: fake version diff too indeed

src/d/actor/d_a_npc_ba1.cpp
slock83: top version matches all too, good catch

src/d/actor/d_a_npc_ba1.cpp
slock83: fakediff too

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: It's more the int + bool pair that seemed weird, but it looks very widespread, so maybe not ? removed the comment

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
LagoLunatic: The compiler didn't change, but the flags did. The devs disabled one single compiler flag for the demo version which had significant effects on the assembly.

src/d/actor/d_a_npc_ba1.cpp
LagoLunatic: You can just click the "Resolve conversation" button instead of commenting "fixed" for each change

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed all of them

src/d/actor/d_a_npc_ba1.cpp
LagoLunatic: The int is probably supposed to be `BOOL` (a typedef for `int`) so you could change that type to make it look neater.

src/d/actor/d_a_npc_ba1.cpp
slock83: done

src/d/actor/d_a_npc_ba1.cpp
slock83: it was much simpler than that : just `mInitialAngle = current.angle;` . Don't know why I didn't think of this when I wrote this function ... Fixed

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed all of them

src/d/actor/d_a_npc_ba1.cpp
slock83: fixed

PR #1145, opened 2026-08-03
Naming dSnap enum for most NPCs

WilliamArnett:


src/d/actor/d_a_obj_ikada.cpp
```diff
@@ -1265,7 +1265,7 @@ bool daObj_Ikada_c::_draw() {
     }
 
     if (isCrane()) {
-        dSnap_RegistFig(DSNAP_TYPE_UNK83, this, 1.0f, 1.0f, 1.0f);
+        dSnap_RegistFig(DSNAP_TYPE_IKADA, this, 1.0f, 1.0f, 1.0f);
```
LagoLunatic, line 1268: There's no figurines for rafts in general, this is specifically salvage corp's ship (coded as a special type of raft)
```suggestion
        dSnap_RegistFig(DSNAP_TYPE_SALVAGE_CORP, this, 1.0f, 1.0f, 1.0f);
```

src/d/actor/d_a_bpw.cpp
```diff
@@ -202,7 +202,7 @@ void body_draw(bpw_class* i_this) {
     if (i_this->m3E0 == 0) {
         dComIfGd_setListMaskOff();
     }
-    dSnap_RegistFig(DSNAP_TYPE_UNKCE, actor, 1.0f, 1.0f, 1.0f);
+    dSnap_RegistFig(DSNAP_TYPE_GND, actor, 1.0f, 1.0f, 1.0f);
```
LagoLunatic, line 205: Maybe add a small `!@bug` comment

include/d/d_snap.h
```diff
@@ -79,26 +79,26 @@ enum {
     /* 0x47 */ DSNAP_TYPE_UNK47,
     /* 0x48 */ DSNAP_TYPE_UNK48,
     /* 0x49 */ DSNAP_TYPE_NPC_LS1,
-    /* 0x4A */ DSNAP_TYPE_UNK4A,
-    /* 0x4B */ DSNAP_TYPE_UNK4B,
-    /* 0x4C */ DSNAP_TYPE_UNK4C,
+    /* 0x4A */ DSNAP_TYPE_NPC_BA1,
+    /* 0x4B */ DSNAP_TYPE_NPC_YW1,
+    /* 0x4C */ DSNAP_TYPE_NPC_AJ1,
```
LagoLunatic, line 84: I'm not sure if we should name all these enums for TUs that aren't decompiled yet, I was leaving them for later so it's easier to check what the code is doing in case there's unusual going on like the raft/salvage corp thing, or jalhalla using puppet ganon's enum, etc.

src/d/actor/d_a_npc_bm1.cpp
```diff
@@ -3906,34 +3906,34 @@ BOOL daNpc_Bm1_c::_draw() {
...
-            dSnap_RegistFig(DSNAP_TYPE_UNK90, this, 1.0f, 1.0f, 1.0f);
+            dSnap_RegistFig(DSNAP_TYPE_NPC_BMC, this, 1.0f, 1.0f, 1.0f);
             break;
         case TYPE_Pashli_e:
-            dSnap_RegistFig(DSNAP_TYPE_UNK97, this, 1.0f, 1.0f, 1.0f);
+            dSnap_RegistFig(DSNAP_TYPE_NPC_BMC_3, this, 1.0f, 1.0f, 1.0f);
             break;
         case TYPE_Namali_e:
-            dSnap_RegistFig(DSNAP_TYPE_UNK92, this, 1.0f, 1.0f, 1.0f);
+            dSnap_RegistFig(DSNAP_TYPE_NPC_BMD, this, 1.0f, 1.0f, 1.0f);
```
LagoLunatic, line 3933: Wonder if maybe all the birdmen should have `BM1_` in their names to make it more clear from their names what's going on? e.g. "BMD" is kalle demos's internal name, so "NPC_BMD" could be confusing (though this is the original dev's fault).

Also, since there's BMD_1, this should definitely have the _0 to distinguish that
```suggestion
            dSnap_RegistFig(DSNAP_TYPE_NPC_BM1_BMD_0, this, 1.0f, 1.0f, 1.0f);
```

include/d/d_snap.h
WilliamArnett: I see what you mean. Still, there are lots of enums that already have stuff like that filled out, like the one for ProcIDs. If you want, I can just fix the names for NPCs that are already done.

src/d/actor/d_a_obj_ikada.cpp
WilliamArnett: Yeah, I did notice that, but I couldn't tell whether it would be more intuitive to name it after its associated figurine or the TU that is in control of it.

src/d/actor/d_a_npc_bm1.cpp
WilliamArnett: Wow, yeah that naming overlap is really confusing. I think BM1 is probably good to add.

include/d/d_snap.h
LagoLunatic: By ProcIDs did you mean process profile names? fpcNm?

The process profile names enum was not named by hand, it was generated automatically from the official profile names for each process so we're 100% sure they're all correct, even down to the inconsistent capitalization the original devs used. e.g. `d_a_lbridge` has `g_profile_LIGHTBRIDGE`, so the enum is called `fpcNm_LIGHTBRIDGE_e`. `d_a_obj_dmgroom` has `g_profile_Obj_Dmgroom`, so the enum is called `fpcNm_Obj_Dmgroom_e`, etc.

Similar deal with the particle names and sound effect names, they're official, except they're taken from TWWHD instead of TWWSD.

But in cases where something like that doesn't exist (e.g. event bits) I name them each by hand, and only do it once the code that uses the enum is decompiled so I can verify for sure how the enum is being used to avoid incorrect assumptions.

If TWWHD has a list of what each photo is called like that then we can just use that and name the whole enum right now, but I don't think anyone has discovered something like that so far.

AH64-dll:
Nice cleanup — verified every renamed constant's call sites are updated within this PR and the enum values/offsets are untouched, so there's no codegen change and all four builds are green. One small request: the PR body is empty — could you add a one-line note on where the names come from (e.g. the registering actor's own dSnap_RegistFig call sites)? It helps reviewers and future name audits.

LagoLunatic:
Fixed in https://github.com/zeldaret/tww/commit/3fe9499e476172ae0b841655b043430049281544

PR #1146, opened 2026-08-04
d_a_npc_kf1 Matching

### LINKED ISSUE #338: d_a_npc_kf1
Kf1

WilliamArnett:
Closes #338


AH64-dll:
Nice work on kf1 — Matching across all four builds with clean stub-completions on tsubo.h. Two things:

1. Two TODO markers left in the code (d_a_npc_kf1.cpp ~1155 `//TODO: m704 is never written to` and ~1367 `//TODO: One/both of these inlines likely return u16…`). Both are sharp observations — the project convention is open questions live in the PR body or an issue, not the final code, so the PR can merge clean. Worth moving them out.
2. Coordination: d_snap.h 0x59 is renamed DSNAP_TYPE_NPC_KF1 here, but #1145 (also open) names the same value DSNAP_TYPE_NPC_GK1_KF1 — the snap type is shared between GK1 and KF1, hence the combined name (same situation as the PM1/KP1 value on #1101). Whichever merges second will conflict on that line; aligning with #1145's name avoids it.

Neither blocks the matching.

PR #1148, opened 2026-08-07
d_menu_fmap Progress

ateddy42:
Progress on zeldaret#1023. Still a lot more to do, but this is most of the smaller functions and overall structure of the file.

Screen Shot 2026-08-07 at 7 35 38 AM


AH64-dll:
Great progress on fmap — honest framing, and the c_API_controller_pad.h bitfix is a real catch. Two notes:

1. QUESTION: the new `dDlst_2DOutFont_c()` ctor in d_2dnumber.h (inits m20/m18/m2C/m74). The original d_2dnumber.s has no `__ct__16dDlst_2DOutFont_cFv` symbol — so either the original initializes those members at the construction sites (d_menu_fmap2), which would make the shared-header ctor an invented abstraction on a DOL class, or the original had an inlined ctor (then it's fine). Also worth noting d_menu_collect.cpp (already on main, NonMatching) does `new dDlst_2DOutFont_c()` — adding the ctor changes its codegen too, and since neither file is flipped in configure.py, CI doesn't verify any of this. Could you confirm which shape the original has?
2. Scope note (non-blocking): the GHOST_SHIP rename touches d_a_ghostship.cpp and d_a_tag_ghostship.cpp — codegen-neutral and good naming, just flagging it's a cross-actor change in a menu PR.

None of these block the progress merge.

include/d/d_menu_fmap.h
```diff
@@ -1,64 +1,191 @@
...
 class dMenu_Fmap_c {
 public:
     virtual ~dMenu_Fmap_c() {}
 
     void draw() { _draw(); }
-    void getCtCurWX() {}
-    void getCtCurWY() {}
-    void getCtCurX() {}
-    void getCtCurY() {}
+    inline u8 getCtActive() {
```
LagoLunatic: `inline` keyword isn't needed for inlines defined in the class body, it's implied

src/d/d_menu_fmap.cpp
```diff
@@ -4,237 +4,1397 @@
...
+        u32 msgNo;
+        if (grid == 32 && dComIfGs_isEventBit(dSv_event_flag_c::UNK_2D80)) {
+            msgNo = 0x31D7;
+        } else {
+            msgNo = grid + 0x31A6;
+        }
+
+        if (
+            grid == 0 || grid == 3 || grid == 10 || grid == 12 || grid == 19 ||
+            grid == 22|| grid == 25 || (u32)(grid - 39) <= 1 ||
```
LagoLunatic: This looks like a compiler optimization, I think this should match too
```suggestion
            grid == 22|| grid == 25 || grid == 39 || grid == 40 ||
```

Also FYI these are sea island indexes and we have an enum for that (dIslandIndex_e) which can be used here, though it can wait for a documentation pass later too

src/d/d_menu_fmap.cpp
```diff
@@ -4,237 +4,1397 @@
...
+    &dMenu_Fmap_c::fmZoomGridLv2Out,
+    &dMenu_Fmap_c::fmZoomGridLv1Out,
+    &dMenu_Fmap_c::fmEndWait,
+};
 
 /* 801AF848-801AFB64       .text _create__12dMenu_Fmap_cFv */
 void dMenu_Fmap_c::_create() {
-    /* Nonmatching */
+    fmapDl.scrn = new J2DScreen();
+    JUT_ASSERT(571, fmapDl.scrn != 0);
```
LagoLunatic: `!= NULL` for the asserts in this func since they're pointers

PR #1149, opened 2026-08-10
Guchhapav/d a obj firewall

### LINKED ISSUE #392: d_a_obj_firewall
Yswdr00

guchhapav:
My first PR :), please let me know if there are any issues! 

PS: I did use an LLM to help me when I got stuck/answer questions but tried to limit my usage as much as possible in case there's any weirdness in here.

Closes #392

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1727,7 +1727,7 @@ config.libs = [
     ActorRel(Matching,    "d_a_obj_eskban"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),    "d_a_obj_ferris"),
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"), "d_a_obj_figure"),
-    ActorRel(Matching, "d_a_obj_firewall"),
+    ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"), "d_a_obj_firewall"),
     ActorRel(NonMatching, "d_a_obj_flame"),
     ActorRel(NonMatching, "d_a_obj_ftree"),
     ActorRel(Matching,    "d_a_obj_ganonbed"),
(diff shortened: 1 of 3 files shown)


guchhapav:
Taking a look at why the build failed, will update the PR with a fix!

AH64-dll:
Hi! Thanks for the PR — matching looks honest (configure.py flip to MatchingFor 3 versions + green CI on all 4 builds, body doesn't overclaim). A few cleanups the maintainers will want before merge:

1. Remove the `//TODO: Convert values to correct flags` above l_cyl_src — TODO markers aren't allowed in the final diff.
2. Use the generated res enums instead of raw indices: `dComIfG_getObjectRes(l_arcname, 6/0xc/9/0xf)` should be `dRes_INDEX_YSWDR00_BDL_YSWDR00_e`, `_BRK_YSWDR00_e`, `_BTK_YSWDR00_e`, `_DZB_FWALL_e` from assets/GZLE01/res/Object/Yswdr00.h (dComIfG_getObjectRes takes the INDEX enum).
3. Use JA_SE_* names for the raw SE ids in set_se: 0x6950 → JA_SE_OBJ_BAR_FRAME_OFF, 0x694e → JA_SE_OBJ_BAR_FRAME_ON, 0x614f → JA_SE_OBJ_BAR_FRAME_BURN (JAZelAudio/JAZelAudio_SE.h).
4. The `daObjFirewall_HIO_c` class in the header is dead scaffolding — it's declared with a ctor but never defined or used, and the original binary has no HIO ctor (check the asm). Remove it.
5. Decompiler-style names: `temp_r3_3`, `temp_r3_2`, `local_1`, `local_2`, and `param_1` (seStart/set_se) should be meaningful lower_snake_case; the seStart param should be i_/o_ prefixed (e.g. i_seId).
6. Style nits: `if(`/`else{`/`for(` missing space before paren in several spots (repo style is `if (`); and the g_profile comment churn (/* Layer ID */ → /* LayerID */ etc.) is a gratuitous restyle of existing comments — please revert those to the original wording.
7. Minor: the `// There is an inline here but not in debug map` and `// I think this is the wrong way...` comments are better resolved in the PR thread than left in code.

None of these affect matching; all are safe cleanups.

PR #1150, opened 2026-08-12
d_a_obj_hcbh 100% matching

K1ngst0m:
Decompile the Tower of the Gods companion-statue pillar actor. Issue #408.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1743,7 +1743,7 @@ config.libs = [
     ActorRel(Matching,    "d_a_obj_hami4"),
     ActorRel(Matching,    "d_a_obj_hat"),
     ActorRel(Matching,    "d_a_obj_hbrf1"),
-    ActorRel(NonMatching, "d_a_obj_hcbh"),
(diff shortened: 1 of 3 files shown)


AH64-dll:
One formatting fix: in _create(), the final else-block body isn't indented and the closing braces are misaligned (`} else {` with the body at the same level). Whitespace-only, safe to fix, but it breaks the 4-space indent convention.

K1ngst0m:
> One formatting fix: in _create(), the final else-block body isn't indented and the closing braces are misaligned (`} else {` with the body at the same level). Whitespace-only, safe to fix, but it breaks the 4-space indent convention.

Thanks for catching!

K1ngst0m:
Hi @LagoLunatic, I hit an intermittent SSL certificate error during checkout. Since external contributors can't rerun failed jobs, the PR stays blocked without confirming whether the build and matching checks pass. The only workaround is to close/reopen the PR or push another commit. Could we add a safe rerun option or automatically retry checkout-only infrastructure failures?

PR #1166, opened 2026-08-14
d_menu_dmap

ateddy42:
100% matching for retail versions. A lot of the functions had overlap with `d_menu_collect `and `d_menu_option`, which was especially helpful for `itemnameSet`/`dnameSet`/`itemnoteSet` methods.

Tried to come up with descriptive names when they made sense (ex: each panel is m[text]Pane using the text it's initialized as (`mNm00Pane`, `mNm01Pane`, etc).

One weird match in `cursorAnime()` with "TColor shenanigans", but the rest matched Ghidra/objdiff pretty closely.

Improves zeldaret#1022


src/d/d_menu_dmap.cpp
```diff
@@ -4,185 +4,2149 @@
...
+        mBossFloorListOffsetY = (mFbPanes[0].mPosCenterOrig.y - mFbPanes[1].mPosCenterOrig.y) * (mBottomFloor - mBossFloor);
+    } else {
+        mBossFloorListOffsetY = 0;
+    }
+#else
+    mBossFloorListOffsetY = (mFbPanes[0].mPosCenterOrig.y - mFbPanes[1].mPosCenterOrig.y) * (mBottomFloor - mBossFloor);
+#endif
+    
+    s16 y = mMpmkPosY + mMapDrawOffsetY;
+#if VERSION > VERSION_DEMO
```
LagoLunatic: This looks like a fake version difference because the code is doing the same thing in both ifdef blocks but is written differently. Demo uses different compiler flags compared to retail, so the real match involves finding how to write the code in a way that matches both. Usually, the demo one will match retail too, so just try removing the retail one and see if it matches both versions.

If you can't find a way to match both versions with the same code, then it's preferable to just match retail and leave demo unmatched rather than creating verbose fake ifdefs like this.

src/d/d_menu_dmap.cpp
```diff
@@ -4,185 +4,2149 @@
...
+
+    y = mCurFloorMapY + mMapDrawOffsetY;
+    s16 mpp1X = mMpp1PosX;
+    dmap = dmap_c;
+    dmap->field_0x364 = mpp1X;
+    dmap->field_0x366 = y;
+#endif
+
+    for (int i = 0; i < (mTopFloor - mBottomFloor + 1); i++) {
+#if VERSION > VERSION_DEMO
```
LagoLunatic: Fake

src/d/d_menu_dmap.cpp
```diff
@@ -4,185 +4,2149 @@
...
+    mLnkTexFlip2 = false;
+    mLnkTexFlip1 = false;
+    mLnkPanes[1].mUserArea = 0;
+    mLnkPanes[2].mUserArea = 0;
+    mBossPane.mUserArea = 100;
+    mBos2Pane.mUserArea = cM_rndF(200.0f) + 300.0f;
+    mBossEyeState = 0;
+    mBossEyeTimer = ((int)(cM_rndF(18.0f) + 40.0f)) << 1;
+
+#if VERSION > VERSION_DEMO
```
LagoLunatic: Fake

src/d/d_menu_dmap.cpp
```diff
@@ -4,185 +4,2149 @@
...
+                    } else if (!mCurFloorOnly) {
+                        if (mCurFloor + 1 == dMap_GetFloorNo(stage, npcPos.y)) {
+                            npc_p.exists = 1;
+                            npc_p.ppane->setWhite(g_mdHIO.mNpcWhite);
+                            npc_p.ppane->setBlack(g_mdHIO.mNpcBlack);
+                            npc_p.width = g_mdHIO.mNpcWidth;
+                            npc_p.height = g_mdHIO.mNpcHeight;
+                            f32 half = 0.5f;
+                            npc_p.posX = mMpp1PosX + (npcPos.x - originX) / scale - npc_p.width * half;
+#if VERSION == VERSION_DEMO
```
LagoLunatic: Fake

src/d/d_menu_dmap.cpp
```diff
@@ -4,185 +4,2149 @@
...
+                        if (mCurFloor + 1 == dMap_GetFloorNo(stage, tentry->base.position.y)) {
+                            if (!dComIfGs_isTbox(tboxNo) && tboxNo != 0xF && tboxNo != 0x10) {
+                                treasure_p[no].exists = 1;
+                                treasure_p[no].ppane->setWhite(g_mdHIO.mTreasureWhite);
+                                treasure_p[no].ppane->setBlack(g_mdHIO.mTreasureBlack);
+                                treasure_p[no].width = g_mdHIO.mTreasureWidth;
+                                treasure_p[no].height = g_mdHIO.mTreasureHeight;
+                                f32 half = 0.5f;
+                                treasure_p[no].posX = mMpp1PosX + (tentry->base.position.x - originX) / scale - treasure_p[no].width * half;
+#if VERSION == VERSION_DEMO
```
LagoLunatic: Fake

I'm gonna stop pointing out the specific ones here but it looks like there's a lot more in this PR

src/d/d_menu_dmap.cpp
ateddy42: fair, I haven't quite figured out the version switching yet. I'll go ahead and remove the demo switches

PR #1177, opened 2026-08-18
Random Minor Cleanups

SuperDude88:
Will probably add other misc stuff to this with time, opening now so I don't forget about them (also fine to merge as-is if desired, half the changes are just a big find-and-replace commit so I'm not sure how much value it currently provides)


LagoLunatic:
Needs conflict resolved

PR #1183, opened 2026-08-26
matches in dMap

Graphicscore:
This is my first try at decomp,
I implemented 3 matches in dMap_c and 1 in dMap_2DPoint_c

Please let me know if I did anything wrong.


src/d/d_map.cpp
```diff
@@ -1965,18 +1994,191 @@ void dMap_c::drawPointShip(f32 param_1, f32 param_2, s16 param_3, f32 param_4, f
...
+    s16 dispX = mNowScaleX * (posX - mNowCenterX);
+    s16 dispY = mNowScaleZ * (posZ - mNowCenterZ);
+
+    u8 size;
+    if (kind == MAP_POINT_AGB_CURSOR) {
+        size = 0x10;
+    } else {
+        size = 4;
+    }
+    u8 halfSize = size >> 1;
```
LagoLunatic: Does it still match if written like this?
```suggestion
    u8 halfSize = size / 2;
```

src/d/d_map.cpp
```diff
@@ -1965,18 +1994,191 @@ void dMap_c::drawPointShip(f32 param_1, f32 param_2, s16 param_3, f32 param_4, f
...
+
+
+    switch (kind) {
+        case MAP_POINT_PLAYER:
+            drawPointPlayer(dispX, dispY, angle);
+            break;
+
+        case MAP_POINT_ENEMY:
+            if (getKindMapType() == MAP_KIND_DUNGEON &&
+                g_dComIfG_gameInfo.save.mMemory.mMembit.isDungeonItemCompass()) {
```
LagoLunatic: dComIfGs_isDungeonItemCompass inline. Same for the other isDungeonItemCompass calls in here. (You basically never access g_dComIfG_gameInfo without using inlines)

src/d/d_map.cpp
```diff
@@ -1965,18 +1994,191 @@ void dMap_c::drawPointShip(f32 param_1, f32 param_2, s16 param_3, f32 param_4, f
...
+    if (kind == MAP_POINT_AGB_CURSOR) {
+        size = 0x10;
+    } else {
+        size = 4;
+    }
+    u8 halfSize = size >> 1;
+
+    if ((s16)dispX + halfSize >= -MAP_VIEW_HALF && dispX - halfSize <= MAP_VIEW_HALF &&
+        dispY + halfSize >= -MAP_VIEW_HALF && dispY - halfSize <= MAP_VIEW_HALF) {
+
```
LagoLunatic: The indentation is messed up from here to the rest of the function, everything after this goes inside this if block but the indentation doesn't show that

src/d/d_map.cpp
```diff
@@ -1965,18 +1994,191 @@ void dMap_c::drawPointShip(f32 param_1, f32 param_2, s16 param_3, f32 param_4, f
...
+        }
+    }
 }
 
 /* 8004CC7C-8004CD68       .text drawPointMain__6dMap_cFUcUcfffScsUcUcUcUc */
-void dMap_c::drawPointMain(u8, u8, f32, f32, f32, s8, s16, u8, u8, u8, u8) {
-    /* Nonmatching */
+void dMap_c::drawPointMain(u8 pointType,u8 param_2, f32 posX,f32 posY, f32 posZ,s8 param_6,s16 param_7,u8 param_8,u8 mGbaName,u8 param_10,u8 param_11)
+{
+    if (g_mDoGaC_gbaCom.mDoGaC_GbaLink() != 0 && g_mDoGaC_gbaCom.mDoGaC_SendStatusCheck(3) != 0) {
```
LagoLunatic: More inlines
```suggestion
    if (mDoGaC_GbaLink() != 0 && mDoGac_SendStatusCheck(3) != 0) {
```

src/d/d_map.cpp
```diff
@@ -1965,18 +1994,191 @@ void dMap_c::drawPointShip(f32 param_1, f32 param_2, s16 param_3, f32 param_4, f
...
+        drawPointAgb(pointType, posX, posY, posZ, param_6, param_7, param_8, mGbaName, param_10, param_11);
+    }
+    drawPointGc(param_2, posX, posY, posZ, param_6, param_7, param_8, mGbaName, param_10, param_11);
+    return;
```
LagoLunatic: Explicit void return; not needed at end of function

src/d/d_map.cpp
Graphicscore: still matches

PR #1186, opened 2026-08-27
d_a_npc_de1 100% Matching

aussa:
Fully decompiles d_a_npc_de1

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1672,7 +1672,7 @@ config.libs = [
     ActorRel(MatchingFor("GZLJ01", "GZLE01", "GZLP01"),  "d_a_npc_btsw"),
     ActorRel(Matching,    "d_a_npc_btsw2"),
     ActorRel(NonMatching, "d_a_npc_co1"),
-    ActorRel(NonMatching, "d_a_npc_de1"),
+    ActorRel(Matching, "d_a_npc_de1"),
     ActorRel(NonMatching, "d_a_npc_ds1"),
     ActorRel(NonMatching, "d_a_npc_gk1"),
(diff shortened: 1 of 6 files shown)


LagoLunatic:
It looks like you added several fake inlines that don't exist in the debug maps.

Did you read the guide on how to decompile? Specifically [the parts about reading the debug maps for inlines](https://github.com/zeldaret/tww/blob/main/docs/decompiling.md#inline-functions-and-how-to-read-the-debug-maps). You can see what the real inlines are in the debug maps so that you know to call those instead of creating fake ones.

aussa:
> It looks like you added several fake inlines that don't exist in the debug maps.
> 
> Did you read the guide on how to decompile? Specifically [the parts about reading the debug maps for inlines](https://github.com/zeldaret/tww/blob/main/docs/decompiling.md#inline-functions-and-how-to-read-the-debug-maps). You can see what the real inlines are in the debug maps so that you know to call those instead of creating fake ones.

Yes, I didn't run the map script ! Sorry for that, I had some help of @WilliamArnett to fix this :)
I think it's good now ? I still see some FAKE and MISSING, but I'm pretty sure they are not relevant for my file

LagoLunatic:
> Yes, I didn't run the map script ! Sorry for that, I had some help of @WilliamArnett to fix this :)
> I think it's good now ? I still see some FAKE and MISSING, but I'm pretty sure they are not relevant for my file

No, the fake inlines you created are still there. I recommend reading the debug maps with your own eyes as well, because the script doesn't give you all the information, it's just for catching things you missed at the end.
image

aussa:
Hello, can you review it now ? It should be good :)

PR #1193, opened 2026-09-01
camera & ev_camera (#3)

Graphicscore:


Graphicscore:
I'll check for demo build differences now, so might add a commit for those before potential merge

LagoLunatic:
Most of those demo ifdefs you added look fake. Demo uses fewer compiler optimizations than retail, meaning that the asm instructions you see in demo are closer to how the devs wrote the code for all versions. If you remove the version difference ifdefs and just write it the demo way for all versions, it should match for all versions while being much more readable. You only need ifdefs when the code is functionally different somehow.

Graphicscore:
> Most of those demo ifdefs you added look fake. Demo uses fewer compiler optimizations than retail, meaning that the asm instructions you see in demo are closer to how the devs wrote the code for all versions. If you remove the version difference ifdefs and just write it the demo way for all versions, it should match for all versions while being much more readable. You only need ifdefs when the code is functionally different somehow.

You're right, I added those because without them demo was not matching, It completely slipped my mind to re-check if retail would work with those changes. I should probably setup 2 workspaces for retail and demo, because changing back and forth has caused me quite the trouble while building as well, and I would have probably noticed this earlier. I'll update the PR

Issue #1023, opened 2026-05-26
d_menu_fmap


LagoLunatic:

ateddy42:
I'll see if I can take a look at this one

PR #1195, opened 2026-09-04
d_menu_fmap

ateddy42:
Matched retail versions for d_menu_fmap. Tried to come up with meaningful names for fields if they made sense, and added a few enums for better readability. Tried to add some demo support, but had a hard time with new padding offsets and fake matches vs real so decided to stop.


ateddy42:
Progress on #1023

PR #1212, opened 2026-09-20
d_a_obj_mmrr 100% matching

### LINKED ISSUE #423: d_a_obj_mmrr

PlatinPython:
Fully decompiled the Mirror Statues from the Earth Temple.

### Notes on variable names
Several of the variables are still unnamed since I'm not sure enough about their function to give them accurate names, but I have some ideas, just nothing concrete.
`Act_c::field_0x2C8` and `Act_c::field_0x958` seem to define the area of the mirror surface, based on the shape of the polygon formed by `Attr_c::field_0x00`, I assume this is used to determine if a light beam is hitting the mirror or not, but I'm not 100% sure.
I think `Attr_c::field_0xB4`, `Attr_c::field_0xB8` and `Attr_c::field_0xBC` define `xz` offset and rotation of said polygon from the actors base position, but again, not 100% sure.
As far as I can tell, `Act_c::field_0xBF8` and `Act_c::field_0xBF9` seem to be flags for determining if the mirror is receiving light, transmitting light or hitting an object, but I'm not sure what exactly.

 Closes #423.


src/d/actor/d_a_obj_mmrr.cpp
```diff
@@ -68,95 +119,264 @@ const dCcD_SrcCps daObjMmrr::Act_c::M_cps_src = {
...
+            }
+        }
+    }
+    return ret;
 }
 
 /* 00001418-00001480       .text eff_start__Q29daObjMmrr5Act_cFv */
 void daObjMmrr::Act_c::eff_start() {
-    /* Nonmatching */
+    dComIfGp_particle_set(0x8294, &current.pos, &shape_angle, NULL, 0xff, &field_0xC04, -1, NULL, NULL, NULL);
```
LagoLunatic: dPa_name enum

PlatinPython: Missed that. Added.

